PR #6, opened 2018-07-06
RST-1118 Adding 3D orientation variable and tests

ayrton04:
Adds a 3D orientation variable that is represented as a quaternion.

### DIFF
--- a/fuse_variables/src/orientation_3d_stamped.cpp
+++ b/fuse_variables/src/orientation_3d_stamped.cpp
@@ -36,6 +36,8 @@
 
 #include <ceres/local_parameterization.h>
 
+#include <cmath>
+
 namespace fuse_variables
 {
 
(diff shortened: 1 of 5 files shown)


fuse_variables/src/orientation_3d_stamped.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+fuse_core::Variable::UniquePtr Orientation3DStamped::clone() const
+{
+  return Orientation3DStamped::make_unique(*this);
+}
+
+ceres::LocalParameterization* Orientation3DStamped::localParameterization() const
+{
+  ceres::LocalParameterization* local_parameterization = new ceres::QuaternionParameterization();
```
svwilliams: Do you need an include for this? Or is it part of `ceres/local_parameterization.h`

Also, you could just `return new ceres::QuaternionParameterization();` ... instead of making the temporary. But I would be willing to bet the compiler gets rid of it anyway.

fuse_variables/test/test_orientation_3d_stamped.cpp
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  double observation_[4];
+};
+
+struct QuaternionCostFunctor
+{
+  explicit QuaternionCostFunctor(double* observation) :
+    function_to_functor_(
+      new ceres::NumericDiffCostFunction<QuaternionCostFunction, ceres::CENTRAL, 3, 4>(
```
svwilliams: I'm dying to know if the auto-diff version gets closer to a normalized quaternion than the numerical-diff version.

svwilliams (approved):

fuse_variables/test/test_orientation_3d_stamped.cpp
ayrton04: I'm guessing switching to it will require more than changing this to the other constructor. I'll give it a go.

fuse_variables/src/orientation_3d_stamped.cpp
ayrton04: It's inside `local_parameterization.h`. I'll simplify the return.

PR #8, opened 2018-07-11
[RST-1121] Added absolute and relative constraints

svwilliams:
Added generic absolute and relative constraints; added 2D pose constraints


fuse_constraints/include/fuse_constraints/absolute_constraint.h
```diff
@@ -0,0 +1,178 @@
...
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_CONSTRAINTS_ABSOLUTE_CONSTRAINT_H
+#define FUSE_CONSTRAINTS_ABSOLUTE_CONSTRAINT_H
+
+#include <fuse_core/constraint.h>
+#include <fuse_core/macros.h>
+#include <fuse_core/uuid.h>
+#include <fuse_variables/acceleration_angular_2d_stamped.h>
```
svwilliams, line 40: These includes are here solely for the typedefs below. As we add new variables, we would also need to update this file. Not sure this is sustainable/desired.

ayrton04 (approved):

PR #14, opened 2018-07-13
[RST-1121] Added a base class for all stamped variables

svwilliams:
Added a base class for all stamped variables


ayrton04 (approved):

ayrton04:
LGTM. The reduction in repeat code is bonus.

PR #15, opened 2018-07-18
[RST-1121] Moved the Graph classes

svwilliams:
Moved the Graph and HashGraph classes into the public repo


ayrton04 (approved):
I'm assuming these were just moved, and that there were no substantive code changes. LGTM.

PR #16, opened 2018-07-19
[RST-1121] move sensor classes

svwilliams:
Moved the sensor model classes to the public repo. I tried to write some more comments in the AsyncSensorModel, so give that a look.


ayrton04 (approved):
New comments look good.

PR #19, opened 2018-07-24
[RST-1121] Moved the pose publishers

svwilliams:
Moved the pose publisher implementations to the public repo

### DIFF
--- a/fuse_publishers/src/path_2d_publisher.cpp
+++ b/fuse_publishers/src/path_2d_publisher.cpp
@@ -55,6 +55,75 @@
 // Register this publisher with ROS as a plugin.
 PLUGINLIB_EXPORT_CLASS(fuse_publishers::Path2DPublisher, fuse_core::Publisher);
 
+
+// Some file-scope functions in an anonymous namespace
+namespace
+{
+
+bool checkVariable(
+  const fuse_core::Variable& variable,
+  const std::string& requested_type,
+  const fuse_core::UUID& requested_device,
+  ros::Time& output_stamp)
+{
+  if (variable.type() != requested_type)
+  {
+    return false;
+  }
+  try
+  {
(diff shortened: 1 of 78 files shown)


ayrton04 (approved):

fuse_publishers/src/path_2d_publisher.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+// Register this publisher with ROS as a plugin.
+PLUGINLIB_EXPORT_CLASS(fuse_publishers::Path2DPublisher, fuse_core::Publisher);
+
+
+// Some file-scope functions in an anonymous namespace
+namespace
+{
+
+bool checkVariable(
```
svwilliams, line 63: Added some local functions to simplify the some tasks.
Unfortunately:
(a) The function signatures are uglier than I like. For example, the 'checkVariable' function also returns a timestamp. And that just seems wrong.
(b) I need slightly different data returned in my two use-cases.
(c) Since I know what variable types I am requesting, then I know certain non-base-class methods exist. This makes it hard to generalize to something that can work with all types.

fuse_variables/src/orientation_2d_stamped.cpp
```diff
@@ -34,12 +34,15 @@
...
 #include <fuse_variables/orientation_2d_stamped.h>
 
+#include <boost/core/demangle.hpp>
 #include <ceres/autodiff_local_parameterization.h>
 
 
 namespace fuse_variables
 {
 
+const std::string Orientation2DStamped::TYPE = boost::core::demangle(typeid(Orientation2DStamped).name());
```
svwilliams, line 46: I added this static constant to the variable type...which I wish was easier to do.

ayrton04 (approved):
Nice!

PR #23, opened 2018-08-02
RST-1228 Fixing quaternion delta computation

ayrton04:
As discussed.

### DIFF
--- a/fuse_constraints/include/fuse_constraints/normal_prior_orientation_3d_cost_functor.h
+++ b/fuse_constraints/include/fuse_constraints/normal_prior_orientation_3d_cost_functor.h
@@ -93,25 +93,25 @@ public:
     using fuse_variables::Orientation3DStamped;
 
     // 1. Compute the delta quaternion
-    T inverse_quaternion[4] =
+    T variable[4] =
     {
       orientation[0],
-      -orientation[1],
-      -orientation[2],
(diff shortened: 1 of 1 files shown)


fuse_constraints/include/fuse_constraints/normal_prior_orientation_3d_cost_functor.h
```diff
@@ -93,25 +93,25 @@ class NormalPriorOrientation3DCostFunctor
...
-      T(b_(3))
+      T(-b_(1)),
+      T(-b_(2)),
+      T(-b_(3))
     };
 
     T output[4];
 
-    ceres::QuaternionProduct(observation, inverse_quaternion, output);
+    ceres::QuaternionProduct(observation, inverse_variable, output);
```
ayrton04: The only outstanding issue is that `ceres::QuaternionProduct` doesn't normalize the output. If we do it ourselves, the final covariances have worse precision (e.g., in our tests, instead of recovering a variance of, say, 0.2, we get 0.20059). Not sure it matters, since the error vector points in the same direction anyway.

svwilliams (approved):

svwilliams (changes requested):
Actually...update comments at the top

svwilliams (approved):

PR #22, opened 2018-08-01
[RST-1226] Convert all Eigen objects to use row-major order

svwilliams:
Converted all Eigen objects to use row-major order. Eigen::Vector objects do not need the RowMajor attribute.

PR #24, opened 2018-08-02
RST-1224 3D pose absolute constraint

ayrton04:
Adds a 3D pose absolute constraint

### DIFF
--- a/fuse_constraints/include/fuse_constraints/absolute_pose_3d_stamped_constraint.h
+++ b/fuse_constraints/include/fuse_constraints/absolute_pose_3d_stamped_constraint.h
@@ -1,0 +1,140 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2018, Locus Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_CONSTRAINTS_ABSOLUTE_POSE_3D_STAMPED_CONSTRAINT_H
+#define FUSE_CONSTRAINTS_ABSOLUTE_POSE_3D_STAMPED_CONSTRAINT_H
+
+#include <fuse_core/constraint.h>
+#include <fuse_core/eigen.h>
+#include <fuse_core/macros.h>
+#include <fuse_core/uuid.h>
+#include <fuse_variables/orientation_3d_stamped.h>
+#include <fuse_variables/position_3d_stamped.h>
+
+#include <ostream>
+#include <vector>
+
+
+namespace fuse_constraints
+{
+
+/**
+ * @brief A constraint that represents either prior information about a 3D pose, or a direct measurement of the 3D pose.
+ *
+ * A 3D pose is the combination of a 3D position and a 3D orientation variable. As a convenience, this class applies
+ * an absolute constraint on both variables at once. This type of constraint arises in many situations. In mapping
+ * it is common to define the very first pose as the origin. Some sensors, such as GPS, provide direct measurements
+ * of the robot's pose in the global frame. And localization systems often match laserscans or pointclouds to a prior
+ * map (scan-to-map measurements). This constraint holds the measured 3D pose and the measurement
+ * uncertainty/covariance. Orientations are represented as quaternions.
+ */
+class AbsolutePose3DStampedConstraint : public fuse_core::Constraint
+{
+public:
+  SMART_PTR_DEFINITIONS(AbsolutePose3DStampedConstraint);
+
+  /**
+   * @brief Create a constraint using a measurement/prior of the 3D pose
+   *
+   * @param[in] position    The variable representing the position components of the pose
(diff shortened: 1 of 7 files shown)


fuse_constraints/include/fuse_constraints/absolute_pose_3d_stamped_constraint.h
```diff
@@ -0,0 +1,144 @@
...
+ * of the robot's pose in the global frame. And localization systems often match laserscans or pointclouds to a prior
+ * map (scan-to-map measurements). This constraint holds the measured 3D pose and the measurement
+ * uncertainty/covariance. Orientations are represented as quaternions.
+ */
+class AbsolutePose3DStampedConstraint : public fuse_core::Constraint
+{
+public:
+  SMART_PTR_DEFINITIONS(AbsolutePose3DStampedConstraint);
+
+  using Vector7d = Eigen::Matrix<double, 7, 1>;
```
ayrton04: Will rebase when this PR is merged:

https://github.com/locusrobotics/fuse/pull/22

fuse_constraints/include/fuse_constraints/normal_prior_pose_3d_cost_functor.h
```diff
@@ -0,0 +1,127 @@
...
+   * @brief Evaluate the cost function. Used by the Ceres optimization engine.
+   */
+  template <typename T>
+  bool operator()(const T* const position, const T* const orientation, T* residual) const;
+
+private:
+  Matrix6d A_;
+  Vector7d b_;
+
+  NormalPriorOrientation3DCostFunctor orientation_functor_;
```
ayrton04: Delegating cost computation for the quaternion to the orientation cost functor.

fuse_constraints/include/fuse_constraints/normal_prior_pose_3d_cost_functor.h
```diff
@@ -0,0 +1,127 @@
...
+};
+
+NormalPriorPose3DCostFunctor::NormalPriorPose3DCostFunctor(const Matrix6d& A, const Vector7d& b) :
+  A_(A),
+  b_(b),
+  orientation_functor_(Eigen::Matrix3d::Identity(), b_.tail<4>())  // Delta will not be scaled
```
ayrton04: So, we have two choices here:

1. Pass in the submatrix of `A_` that contains the covariance for the orientation here. Then, in `operator()`, we can only apply the covariance to the `position` variables.

2. Pass in the identity here so that whatever scaling is done has no effect, and then multiply the full `A_` by the entire residual.

I opted for (2), but could be convinced to do (1).

fuse_constraints/include/fuse_constraints/normal_prior_pose_3d_cost_functor.h
```diff
@@ -0,0 +1,127 @@
...
+  orientation_functor_(orientation, &residual[3]);
+
+  Eigen::Map<Eigen::Matrix<T, 6, 1> > residuals_map(residual);
+  residuals_map(0) = position[0] - T(b_(0));
+  residuals_map(1) = position[1] - T(b_(1));
+  residuals_map(2) = position[2] - T(b_(2));
+
+  // Scale the residuals by the square root information matrix to account for
+  // the measurement uncertainty.
+  residuals_map.applyOnTheLeft(A_.template cast<T>());
```
ayrton04, line 118: Note that the covariance is being applied to the entire residual, not just the position.

fuse_constraints/test/test_absolute_pose_3d_stamped_constraint.cpp
```diff
@@ -0,0 +1,205 @@
...
+  std::vector<double> cov_pos_or(position_variable->size() * orientation_variable->size());
+  covariance.GetCovarianceBlock(position_variable->data(), orientation_variable->data(), cov_pos_or.data());
+
+  // Assemble the full covariance from the covariance blocks
+  Eigen::Map<Eigen::Matrix3d> pos_pos(cov_pos_pos.data());
+  Eigen::Map<Eigen::Matrix<double, 4, 4>> or_or(cov_or_or.data());
+  Eigen::Map<Eigen::Matrix<double, 3, 4, Eigen::RowMajor>> pos_or(cov_pos_or.data());
+
+  Matrix6d actual_covariance;
+  actual_covariance << pos_pos, pos_or.block<3, 3>(0, 1), pos_or.block<3, 3>(0, 1).transpose(), or_or.block<3, 3>(1, 1);
```
ayrton04: This test will fail until this is merged:

https://github.com/locusrobotics/fuse/pull/23

fuse_core/test/test_constraint.cpp
```diff
@@ -45,7 +45,7 @@ TEST(Constraint, Constructor)
   {
     fuse_core::UUID variable_uuid1 = fuse_core::uuid::generate();
     ExampleConstraint constraint{variable_uuid1};
-    ASSERT_EQ(1, constraint.variables().size());
+    ASSERT_EQ(1u, constraint.variables().size());
```
ayrton04, line 48: Stuff that wouldn't build with my compiler version.

fuse_constraints/include/fuse_constraints/absolute_pose_3d_stamped_constraint.h
svwilliams: #22 is now merged

fuse_constraints/include/fuse_constraints/normal_prior_pose_3d_cost_functor.h
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * The Ceres::NormalPrior cost function only supports a single variable. This is a convenience cost function that
+ * applies a prior constraint on both the 3D position and orientation variables at once.
+ *
+ * The cost function is of the form:
+ *
+ *             ||    [  x       -     b(0)  ] ||^2
+ *             ||    [  y       -     b(1)  ] ||
+ *             ||    [  z       -     b(2)  ] ||
+ *             ||A * [  (b(4:7) * q^-1)(1)  ] ||
```
svwilliams: `(b(3:6)^-1 * q)(1)`

fuse_constraints/include/fuse_constraints/normal_prior_pose_3d_cost_functor.h
svwilliams: (2) is definitely the better answer.

If we ever get so concerned about performance that the needless scaling by identity is an issue, I would reimplement the quaternion cost here.

fuse_constraints/test/test_absolute_pose_3d_stamped_constraint.cpp
```diff
@@ -0,0 +1,205 @@
...
+TEST(AbsolutePose3DStampedConstraint, Constructor)
+{
+  // Construct a constraint just to make sure it compiles.
+  Position3DStamped position_variable(ros::Time(1234, 5678), fuse_core::uuid::generate("walle"));
+  Orientation3DStamped orientation_variable(ros::Time(1234, 5678), fuse_core::uuid::generate("walle"));
+
+  Vector7d mean;
+  mean << 1.0, 2.0, 3.0, 1.0, 0.0, 0.0, 0.0;
+
+  Matrix6d cov = Matrix6d::Random();
```
svwilliams: That looks scary. Will that produce a positive-definite, symmetric matrix?
Maybe use the COV from the test below?

fuse_variables/test/test_fixed_size_variable.cpp
```diff
@@ -59,8 +59,8 @@ TEST(FixedSizeVariable, Size)
 {
   // Verify the expected size is returned
   TestVariable variable;
-  EXPECT_EQ(2, variable.size());  // base class interface
-  EXPECT_EQ(2, TestVariable::SIZE);  // static member variable
+  EXPECT_EQ(2u, variable.size());  // base class interface
+  EXPECT_EQ(2ul, TestVariable::SIZE);  // static member variable
```
svwilliams: weird. Why does that one have to be `ul`, but the others just `u`?

svwilliams (approved):
Approved...pending rebase with fuse_core::MatrixXXX types

fuse_variables/test/test_fixed_size_variable.cpp
ayrton04: I just moused over on `TestVariable::SIZE` and saw its type, and went with it. They really only have to be the same sign. I can change it if it bothers you. :)

fuse_constraints/test/test_absolute_pose_3d_stamped_constraint.cpp
ayrton04: I mean, we're just making sure it won't throw an exception during construction, right? Are you suggesting that perhaps we might or should?

svwilliams: We are solving for the inverse here: https://github.com/locusrobotics/fuse/pull/24/files#diff-6e8da99abd4d3d11572c3cd2a4db0058R50

So something bad _should_ happen if the input covariance is not positive-definite, right? I'm guessing/hoping an exception.

ayrton04: Fixed.

PR #27, opened 2018-08-15
RST-1239 Adding 3D relative pose constraint

ayrton04:


fuse_constraints/include/fuse_constraints/normal_prior_orientation_3d_cost_functor.h
```diff
@@ -110,7 +110,7 @@ class NormalPriorOrientation3DCostFunctor
 
     T output[4];
 
-    ceres::QuaternionProduct(inverse_observation, variable, output);
+    ceres::QuaternionProduct(variable, inverse_observation, output);
```
ayrton04: As discussed.

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
```diff
@@ -0,0 +1,189 @@
...
+  ceres::QuaternionProduct(
+    observation_orientation_delta,
+    variable_orientation_delta_inverse,
+    delta_difference_orientation);
+
+  // 4. Compute the position delta, and throw everything into a residual at the same time
+  residual[0] = variable_position_delta_rotated[0] - b_[0];
+  residual[1] = variable_position_delta_rotated[1] - b_[1];
+  residual[2] = variable_position_delta_rotated[2] - b_[2];
+  residual[3] = T(2.0) * delta_difference_orientation[1];
```
ayrton04, line 165: These 2.0 terms interest me.

svwilliams: Interesting, right. My theory is that the 2.0 scalar is to negate the half-angles used to define the quaternion:
`cos(theta/2), omega * sin(theta/2)`

*If* the error is small, then `sin(theta/2) -> theta/2`. By multiplying by 2.0, the residual becomes just `theta`, which is then normalized by the covariance. Thus, the covariance can be in "angle" units rather than "quaternion" units.

At least, that's the best theory I have. As such, might want to consider adding these 2.0 terms to the absolute/relative orientation constraints as well.

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
```diff
@@ -0,0 +1,189 @@
...
+  // and the measurement's orientation delta
+  T observation_orientation_delta[4] =
+  {
+    T(b_(3)),
+    T(b_(4)),
+    T(b_(5)),
+    T(b_(6))
+  };
+
+  T variable_orientation_delta_inverse[4] =
```
svwilliams: And this doesn't cause "block-stupidness"? I thought this was the formulation that you started with (sign of variable delta is positive for the position, but negative-ish for the rotations), and it caused the off-diagonal blocks to be negative.

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
```diff
@@ -0,0 +1,189 @@
...
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_CONSTRAINTS_NORMAL_DELTA_POSE_3D_COST_FUNCTOR_H
+#define FUSE_CONSTRAINTS_NORMAL_DELTA_POSE_3D_COST_FUNCTOR_H
+
+#include <fuse_constraints/util.h>
+#include <fuse_core/eigen.h>
+
+#include <ceres/internal/disable_warnings.h>
```
svwilliams: I have started removing this, since I'm not really sure what they are disabling. And because "we" forgot to re-enable the warnings in a bunch of places.

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
```diff
@@ -0,0 +1,189 @@
...
+#include <ceres/internal/eigen.h>
```
svwilliams: And since we now have `fuse_core/eigen.h`, we aren't really using any of the `ceres/eigen.h` functions/typedefs anymore.

fuse_constraints/test/test_relative_pose_3d_stamped_constraint.cpp
```diff
@@ -0,0 +1,308 @@
...
+    Eigen::Map<fuse_core::Matrix4d> or_or(cov_or_or.data());
+    Eigen::Map<Eigen::Matrix<double, 3, 4, Eigen::RowMajor>> pos_or(cov_pos_or.data());
+
+    fuse_core::Matrix6d actual_covariance;
+    actual_covariance <<
+      pos_pos, pos_or.block<3, 3>(0, 1),
+      pos_or.block<3, 3>(0, 1).transpose(), or_or.block<3, 3>(1, 1);
+
+    fuse_core::Matrix6d expected_covariance;
+    expected_covariance <<
```
svwilliams, line 299: Where did you get the expected covariance from? I would have naively expected the rotation covariance to be near 2.0.

svwilliams (approved):

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
ayrton04: Hrm, strangely, no. This is an odd one, because they are also following the standard of `error = variable - measurement`, and I believe I am doing the same thing here that they were.

https://github.com/ceres-solver/ceres-solver/blob/master/examples/slam/pose_graph_3d/pose_graph_3d_error_term.h#L103
https://github.com/ceres-solver/ceres-solver/blob/master/examples/slam/pose_graph_3d/pose_graph_3d_error_term.h#L96

Bizarrely, putting them the way we did for the orientation variable has zero effect on the outcome; the covariance matrix is utterly identical. Given that one form is the inverse of the other, I find this very strange.

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
ayrton04: OK, can do that as part of this PR.

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
ayrton04: Could they have this wrong in theirs? Check this Octave script out. I can provide the functions, if you want to see those:

```
% Standard is error = variable - measurement

quat_variable = rpy_to_quat(0, 0, 0.2);
quat_measurement = rpy_to_quat(0, 0, 0.1);

quat_delta1 = quat_mult(quat_variable, quat_conj(quat_measurement));  % What we do for the orientation constraint
quat_delta2 = quat_mult(quat_measurement, quat_conj(quat_variable));  % The way this code is now

[r1 p1 y1] = quat_to_rpy(quat_delta1)    % This one should be correct
[r2 p2 y2] = quat_to_rpy(quat_delta2)
```

Sure enough:

```
r1 = 0
p1 = 0
y1 = 0.1

r2 = 0
p2 = 0
y2 = -0.1
```

So I think we need to change this here. Since it apparently doesn't affect the results in my test, I feel both good and bad about that.

fuse_constraints/test/test_relative_pose_3d_stamped_constraint.cpp
ayrton04: So, funny you should ask. These terms are a result of the `2.0` term in the orientation residuals. If I remove them, the covariance is 2, just as you expect.

Furthermore, if I _add_ the 2.0 term to the orientation cost functor, it breaks the prior covariance check here:

https://github.com/locusrobotics/fuse/pull/27/files#diff-49f598a9a090323e4f949817796747c2R258

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
svwilliams: I'll be honest, what bothered me before was that it _did_ change the covariance. I thought the auto-computed Jacobian should contain some negative signs that would fix the (A - B) vs (B - A) issues. But I like the quat_delta1 formulation better, just for symmetry with the translation.

fuse_constraints/test/test_relative_pose_3d_stamped_constraint.cpp
svwilliams: Does that mean these 2.0 terms are a bad idea? I feel like they should exist everywhere, or nowhere. I suppose I'm going to have to read that 95 page documents. :(

Also, have you played with that `GetCovarianceBlockInTangentSpace()` function you showed me at some point? I'm wondering if that is part of the answer as well.

ayrton04: No, I think they're probably fine, but it means that we'll either have to (a) see how the tangent space covariance works out, or (b) understand what the ramifications are for the covariance as-is.

svwilliams:
shipit

PR #33, opened 2018-12-06
[RST-1567] Check the system has started before attempting to optimize

svwilliams:
Running a bagfile-based optimizer test throws an error during a covariance calculation, claiming the system is rank deficient.

```
WARNING: Logging before InitGoogleLogging() is written to STDERR
E1206 06:45:02.572094 18895 covariance_impl.cc:656] Jacobian matrix is rank deficient. Number of columns: 6 rank: 3
terminate called after throwing an instance of 'std::runtime_error'
what(): Could not compute requested covariance blocks.
```

* The error is thrown almost immediately after the bagfile starts playing.
* Currently the covariance computation is only done in the publishers, for displaying the PoseWithCovariance.
* The rank deficiency of 6 columns but rank 3 makes me think the first relative pose is added, but the initial prior/origin constraint is not. This is supposed to be guarded from happening by the "ignition sensor" system.
* The fact that the ignition sensor is not publishing is an unrelated problem with my machine. :)

### DIFF
--- a/fuse_optimizers/src/batch_optimizer.cpp
+++ b/fuse_optimizers/src/batch_optimizer.cpp
@@ -202,6 +202,11 @@ void BatchOptimizer::optimizationLoop()
 
 void BatchOptimizer::optimizerTimerCallback(const ros::TimerEvent& event)
 {
+  // If an "ignition" transaction hasn't been received, then we can't do anything yet.
+  if (!started_)
+  {
+    return;
+  }
   // Attempt to generate motion models for any queued transactions
   applyMotionModelsToQueue();
   // Check if there is any pending information to be applied to the graph.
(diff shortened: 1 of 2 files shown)


davelkan-zz (approved):
shouldn't this have been broken before if it was missing an import? you haven't added anything that depends on vector

fuse_optimizers/src/batch_optimizer.cpp
```diff
@@ -202,6 +202,11 @@ void BatchOptimizer::optimizationLoop()
 
 void BatchOptimizer::optimizerTimerCallback(const ros::TimerEvent& event)
 {
+  // If an "ignition" transaction hasn't been received, then we can't do anything yet.
+  if (!started_)
```
svwilliams, line 206: It was guarded in one place, but not both...
https://github.com/locusrobotics/fuse/blob/2cdbbfba815e585104ba4f3457462853564fba9b/fuse_optimizers/src/batch_optimizer.cpp#L268

svwilliams:
It should have been, yes. Not sure how Tom got that passed the linter.

ayrton04 (approved):

PR #35, opened 2019-01-18
[RST-1477] Simplified the sensor<-->optimizer API

svwilliams:
Simplified the sensor<-->optimizer API. Moved the implementation details of the optimizer transaction callback into the optimizer where it belongs.

Originally the SensorModel API required the Optimizer to provide the Optimizer's callback queue. This was really an implementation detail of the Optimizer callback and didn't belong in the SensorModel. If a callback queue is needed to execute the desired callback function, that callback queue can easily be bound to the callback function itself using std::bind or similar.

This PR:
* removes the callback queue from SensorModel class
* moves the "insert function into the optimizer's callback queue" logic into the Optimizer class
* renames to the "injectCallback()" method of the AsyncSensorModel to simply "sendTransaction()", since we are no longer explicitly injecting a callback into a callback queue.


fuse_core/include/fuse_core/async_sensor_model.h
```diff
@@ -119,14 +118,10 @@ class AsyncSensorModel : public SensorModel
...
-   *                                       queue is performed.
    */
   void initialize(
     const std::string& name,
-    TransactionCallback transaction_callback,
-    ros::CallbackQueue* transaction_callback_queue) final;
+    TransactionCallback transaction_callback) final;
 
   /**
    * @brief Inject the Transaction into the callback queue registered during initialize()
```
ayrton04: If we are no longer explicitly injecting, should this description change?

fuse_core/test/test_async_sensor_model.cpp
```diff
@@ -109,23 +109,15 @@ TEST(AsyncSensorModel, OnGraphUpdate)
...
   std::set<ros::Time> stamps;
   fuse_core::Transaction::SharedPtr transaction;  // nullptr, okay because we don't actually use it for anything
-  sensor.injectCallback(stamps, transaction);
-  EXPECT_FALSE(received_transaction);
-  ros::Time wait_time_elapsed = ros::Time::now() + ros::Duration(10.0);
-  while (!received_transaction && ros::Time::now() < wait_time_elapsed)
-  {
-    ros::Duration(0.1).sleep();
-  }
+  sensor.publish(stamps, transaction);
```
ayrton04: Yeah, I can see it being confused with the ROS sense of publication, but I'm not really sure what to do about that, or whether we need to. It ends up in a callback anyway. Works for me.

ayrton04 (approved):

fuse_core/include/fuse_core/sensor_model.h
```diff
@@ -62,11 +62,12 @@ using TransactionCallback = std::function<void(const std::set<ros::Time>& stamps
...
  *
- * The interface SensorModel class provides the injectCallback() function. This function serves the same basic purpose
- * as publishing a Transaction using a standard ROS publisher, but this method is specific to sensor models and the
- * fuse ecosystem. Under the hood, this inserts a call into the fuse optimizer's callback queue, just as publishing a
- * message would, but the memory copy and serialization are avoided. Derived classes should call this method whenever
- * they are ready to send a fuse_core::Transaction object to the optimizer.
+ * During the initialize() call, the optimizer will provide a transaction callback. This function serves the same
+ * basic purpose as publishing a Transaction using a standard ROS publisher. Derived classes should call this function
+ * whenever they are ready to send a fuse_core::Transaction object to the optimizer. This function will be called in
+ * the SensorModel's thread. The optimizer implmentation should ensure that this callback to safe to execute in this
```
matthewbwhitaker: *implementation

matthewbwhitaker: "_is_ safe to execute..."?

fuse_optimizers/src/optimizer.cpp
```diff
@@ -284,4 +284,16 @@ void Optimizer::notify(
   }
 }
 
+void Optimizer::injectCallback(
+  const std::string& sensor_name,
+  const std::set<ros::Time>& stamps,
+  const fuse_core::Transaction::SharedPtr& transaction)
+{
+  // We are going to insert a call to the derived class's transactionCallback() method into the global callback queue.
```
matthewbwhitaker, line 292: *class'?

matthewbwhitaker (approved):

fuse_optimizers/src/optimizer.cpp
svwilliams: No. I believe `class's` is correct.
https://english.stackexchange.com/questions/1073/what-is-the-correct-possessive-for-nouns-ending-in-s

PR #36, opened 2019-01-18
[RST-1476] Added API docs for the Variable class

svwilliams:
Given the length of time we spent discussing Variables for fuse_rl, I felt like some extensive documentation was needed to describe exactly what a Variable is. Give it a read, let me know what you think, and what "Tom from 3 months ago" would have thought.

Also, it's a lot of text. Let me know anything that might jazz it up a bit..so people will actually read it.

[Markdown version](https://github.com/locusrobotics/fuse/blob/RST-1476-documentation-2/doc/Variables.md)


doc/Variables.md
```diff
@@ -0,0 +1,117 @@
...
+two things. First, it holds the variable's _value_, acting as a one-or-more-dimensional tuple of floating-point data
+that defines the physical property. Second, it defines the variable's _identity_, holding the additional information
+that distinguishes one variable instance from another. Interestingly, and somewhat counter-intuitively, the _value_
+of the variable tuple **does not affect** the identity of a variable instance. The purpose of fuse is to solve for
+the most likely _value_ of all of the variables, and hence the variable instance _values_ will change over time as
+new evidence is accumulated. But the _identity_ of a variable instance is immutable, and independent of its _value_.
+
+A concrete example will help explain these concepts. We have two ground robots, cleverly named R1 and R2, and we want
+to estimate the pose of both robots as they drive around. Thus, the physical property of interest is the 2D pose. A
+2D pose consists of three floating-point quantities: `x`, `y`, and `'yaw`. Thus our Variable type must include a
```
ayrton04: Extra single quote next to `yaw`.

doc/Variables.md
```diff
@@ -0,0 +1,117 @@
...
+different robots, so we need to differentiate between the pose of R1 and the pose of R2. Thus our Variable type must
+include some sort of `robot_id` field. Different instances of our variable will set the `robot_id` field to either 'R1'
+or 'R2'. Additionally, the robots move around over time. So the pose of a robot is not valid for _any_ time, but
+represents the pose of the robot at a _specific_ time. And so our Variable type must also include a `timestamp` field
+to differentiate a pose at time `t1` for a pose at time `t2`.
+
+Just to reiterate the important concepts:
+
+* A variable type defines a _value_ that may be one or more floating-point quantities.
+* A variable instance's _value_ is unknown and/or uncertain. It will change over time as evidence is added.
```
ayrton04: So I get what you're saying here: a given node in the graph, representing a variable, will have its value change as more variables and constraints are added to the graph over time. However, I can see someone confusing the notion of "variable value changing over time" and the notion of "a variable measuring pose at time t1 is a different variable than the measurement of pose at time t2." You call that fact out in the paragraph above, but it gets wrapped into the point on line 26 down here.

In EKF land, I have been referring to each dimension in the state as a variable, and its value changes over time. But we only have one state, so the value of, e.g., variable `x` is 23.1 at time `t1`, and 23.4 at time `t2`. But that's not the same as it is in graph land, where a variable is a measurement of N dimensions from a given sensor at a given time.

doc/Variables.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Variable Design Principles
+
+The most important consideration when creating a new Variable type is deciding what data dimensions should be
+included. Including too many physical properties in a Variable type will be inefficient and cumbersome when used in
+places where most of the Variable's value dimensions are unused. However, including too few physical properties in
+a Variable type also leads to inefficient and cumbersome usage when even the simplest of observation models involve
+many variables. This is one of those "Goldilocks principle" situations.
+
+![Goldilocks principle](http://home.netcom.com/~swansont_2/goldilocks.jpg)
```
ayrton04, line 50: Hahaha

doc/Variables.md
```diff
@@ -0,0 +1,117 @@
...
+
+Understanding how Variable types interact with the rest of the system will help in the design of "good" Variable
+types.
+
+1. The fuse stack is designed to combine observations _of the same variable instance_ from multiple sources. In
+   the robotics example above, one could imagine that the robots measure their current pose using wheel odometry as
+   well as with a laserscan -- two different measurements of the same pose. In order for these two measurements to
+   be combined together in fuse, **they must use the same variable type**. This means that if a variable type defines
+   a 15-dimensional, but a sensor only measures 3 of those dimensions, the sensor must still consume all 15 dimensions
```
ayrton04: * define a 15-dimensional _state_?

doc/Variables.md
```diff
@@ -0,0 +1,117 @@
...
+   and define the measurement error for all 15 dimensions (which will likely mean specifying 12 dimensions worth
```
ayrton04: So I can see this point causing confusion, at least until they get to point (2). But point (2) makes things very clear.

doc/Variables.md
```diff
@@ -0,0 +1,117 @@
...
+* `Derived::uuid() -> fuse_core::UUID`: This is the _identity_ of the variable. Each derived class is required to
+  return a unique ID to act as the identity of the variable. The UUID must be different for each different combination
+  of class and _identity_ property values. In the robotics example, the 2D pose UUID should be built from the 2D pose
+  type string, the `robot_id` field, and the `timestamp` field. Importantly, if a Variable instance has the same
+  _identity_ property values, **the same UUID should be returned**. This is how multiple sensors can produce the same
+  variable identity in a distributed fashion. Again, in the robotics example, the UUID of a 2D pose with a `robot_id`
+  of 'R1' and a `timestamp` of 10.0 produced by the odometry sensor should be the same UUID of a 2D pose with a
+  `robot_id` of 'R1' and a `timestamp` of 10.0 produced by the laserscan sensor. Some functions for generating UUIDs
+  are provided [here](fuse_core/include/fuse_core/uuid.h).
+* `Derived::print(std::ostream& stream)`: It's nice to print variable information during debugging. Each
```
ayrton04: Given that we have the UUID, the data, and the data length, it makes me wonder if having a base class print implementation might be worthwhile.

doc/Variables.md
```diff
@@ -0,0 +1,117 @@
...
+  [Ceres documentation](http://ceres-solver.org/nnls_modeling.html#localparameterization) for an in-depth discussion
+  of "local parameterizations" and their uses. If the derived Variable type requires a local parameterization, this
+  method may be overridden to provide it.
+* It is highly recommended that all derived Variable types include `SMART_PTR_DEFINITIONS(Derived);` in the public
+  interface. This defines some common smart pointer aliases.
+
+Additional member properties and member functions may be added to the derived Variable class. These can only be used
+when an object is created with a known type. Despite this limitation, providing some syntax sugar is encouraged, as
+it can make working with the derived Variable object more satisfying. For example, `fuse_variables` classes provide
+named accessors for the individual dimension values. This allows use of `var.y()` in lieu of `var.data()[1]`.
```
ayrton04, line 145: I think I have the latter in a few places in `fuse_rl`.

ayrton04 (approved):
Looks great!

doc/Variables.md
svwilliams: Probably. Making a separate ticket.

doc/Variables.md
```diff
@@ -0,0 +1,144 @@
+# Variables
+
+In the fuse stack, a Variable does two things:
+
+## Variable Value
+
+First, a Variable has a _value_. A Variable holds one or more dimensions of floating point data that represents some
+physical or logical property. For example, a two dimensional pose consists of `x`, `y`, and `yaw` floating point
+values. A single Variable class could be implemented hold all three pose dimensions. The exact value of each dimension
```
matthewbwhitaker: "implemented *to* hold" or "implemented *that holds* _all_"

doc/Variables.md
```diff
@@ -0,0 +1,144 @@
...
+  for each time instant for which the process value is to be estimated. For example, the pose of the robot will change
+  over time, so we need a unique Variable identity representing the robot pose at time `t1` **and** another unique
+  Variable identity representing the robot pose at time `t2`. Any time-varying process must be discretized within
+  the fuse stack.
+
+The Variable identity takes the form of a UUID or hash value, and is generally derived from a set of additional
+properties that describe what makes each occurrence unique from other occurrences. In the case of a time-varying
+process, this will likely involve the timestamp. In the case of a Variable that can describe multiple robots, this
+might involve the serial number of the robot.
+
```
matthewbwhitaker: Tricky and time-consuming, but some diagrams to illustrate each of the uses of the identity might be an aid to more visual learners. Even just something linking the component parts of the identity to bot positions, a time series, etc.

doc/Variables.md
```diff
@@ -0,0 +1,144 @@
...
+* `SMART_PTR_DEFINITIONS(Derived);`
+
+  It is highly recommended that all derived Variable types include `SMART_PTR_DEFINITIONS(Derived);` in the public
+  interface. This defines some common smart pointer aliases, such as `Derived::SharedPtr` and `Derived::UniquePtr`.
+
+Additional member properties and member functions may be added to the derived Variable class. These can only be used
+when an object is created with a known type. Despite this limitation, providing some syntax sugar is encouraged, as
+it can make working with the derived Variable objects more satisfying. For example, `fuse_variables` classes provide
+named accessors for the individual dimension values. This allows use of `var.y()` in lieu of `var.data()[1]`.
```
matthewbwhitaker, line 145: Some general thoughts:

- Suggest just saying "Variable" instead of "Variable type". They start off interchangeable in the beginning, but then "Variable type" becomes dominant, but I think it's unnecessary and cumbersome.
- When talking about Variable value and Variable identity, suggest _dropping_ Variable completely within the relevant sections, and just referring to _value_ or _identity_ (more relevant for the lengthier _identity_ description).
- Variable capitalization of "Variable". Don't have a strong opinion on yay or nay, or whether to italicize, but it should be consistent so that it's clear when we're talking about a Fuse Variable.

matthewbwhitaker: I think also taking one of the existing Variable types within fuse and walking through the reasoning behind it may help to illustrate the core concepts, especially if there is a chance to weigh pros and cons of the way it was defined. Again, an onerous task, but just leaving it here as a suggestion.

matthewbwhitaker (approved):
Looks good, reads well. Followed it fine when I had the chance to focus on it. Some suggestions - but no blockers.

PR #37, opened 2019-01-31
[RST-1653] transaction stamps

svwilliams:
I'm sorry....
This seemed like a simple change...
Then it rippled to a lot more places than I expected.
Might make sense to walk through the changes on a per-commit basis.

Anyway, whenever a sensor/motion model publishes a Transaction, it also sends a set of "involved timestamps" with it. Since the two always go together, I thought "Why not put them together?" It has the added bonus of moving the timestamp container around via pointer instead of via copy...because the Transaction is moved around via pointer.

There will be additional ripples to `fuse_rl` and `fuse_locus` later.

### DIFF
--- a/fuse_core/include/fuse_core/async_motion_model.h
+++ b/fuse_core/include/fuse_core/async_motion_model.h
@@ -41,9 +41,7 @@
 #include <ros/callback_queue.h>
 #include <ros/node_handle.h>
 #include <ros/spinner.h>
-#include <ros/time.h>
 
-#include <set>
 #include <string>
 
 
@@ -67,13 +65,13 @@ namespace fuse_core
  * interaction of motion model nodes is best compared to a service call -- an external actor will provide a set of
  * timestamps and wait for the motion model to respond with the required set of constraints to link the timestamps
  * together (along with any previously existing timestamps). In lieu of a ROS service callback function, the
- * AsyncMotionModel class requires the queryCallback() function to be implemented. This callback will be executed
+ * AsyncMotionModel class requires the applyCallback() function to be implemented. This callback will be executed
  * from the same callback queue as any other subscriptions or service callbacks.
  * 
  * Derived classes:
  * - _probably_ need to implement the onInit() method. This method is used to configure the motion model for operation.
  *   This includes things like accessing the parameter server and subscribing to sensor topics.
- * - _must_ implement the queryCallback() method. This is the communication mechanism between the parent/optimizer and
+ * - _must_ implement the applyCallback() method. This is the communication mechanism between the parent/optimizer and
  *   the derived motion model. This is how the optimizer tells the motion model what timestamps have been added, and
  *   how the motion model sends motion model constraints to the optimizer.
  * - may _optionally_ implement the onGraphUpdate() method. This should only be done if the derived motion model needs
@@ -126,7 +124,7 @@ public:
   void graphCallback(Graph::ConstSharedPtr graph) final;
 
   /**
-   * @brief Augment a transaction structure such that the provided timestamps are connected by motion model constraints.
+   * @brief Augment a transaction object such that all involved timestamps are connected by motion model constraints.
    *
    * This method will be called by the optimizer, in the optimizer's thread, before each sensor transaction is applied
    * to the Graph. This implementation packages a call to the pure virtual method applyCallback() and inserts it into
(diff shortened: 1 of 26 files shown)


fuse_core/include/fuse_core/async_motion_model.h
```diff
@@ -157,20 +154,19 @@ class AsyncMotionModel : public MotionModel
...
    * callback, where the caller makes a request and blocks until the request is completed.
    *
-   * @param[in]  stamps      The set of timestamps that should be connected by motion model constraints
-   * @param[out] transaction The transaction object that should be augmented with motion model constraints
-   * @return                 True if the motion models were generated successfully, false otherwise
+   * @param[in,out] transaction The transaction object that should be augmented with motion model constraints
+   * @return                    True if the motion models were generated successfully, false otherwise
    */
-  virtual bool applyCallback(const std::set<ros::Time>& stamps, Transaction& transaction) = 0;
+  virtual bool applyCallback(Transaction& transaction) = 0;
```
ayrton04, line 169: So tidy!

fuse_core/include/fuse_core/graph.h
```diff
@@ -58,7 +58,7 @@ namespace fuse_core
  * be used in range-based for loops), an empty() method, and a front() method for directly accessing the first
  * member. When dereferenced, an iterator returns a const Constraint&.
  */
-using const_constraint_range = boost::any_range<const fuse_core::Constraint, boost::forward_traversal_tag>;
+using const_constraint_range = boost::any_range<const Constraint, boost::forward_traversal_tag>;
```
ayrton04: I note that this header includes `transaction.h`, which also has this typedef, and in the same namespace. Note sure if you care. You probably don't want to assume anything about the typedefs in your includes.

fuse_core/src/timestamp_manager.cpp
```diff
@@ -75,26 +75,34 @@ void TimestampManager::query(
   }
   // Create a list of all the required timestamps involved in motion model segments that must be created
   // Add all of the existing timestamps between the first and last input stamp
-  std::set<ros::Time> augmented_stamps(stamps);
+  Transaction motion_model_transaction;
+  auto first_stamp = *stamps.begin();
+  auto last_stamp = *stamps.begin();
+  for (const auto& stamp : stamps)
```
ayrton04: Won't this just end up making `last_stamp` == `stamps.back()`? Or is there something about this iteration that might terminate before we reach `stamps.end()`? Because otherwise you could do

```
auto last_stamp = stamps.back();
std::for_each(stamps.begin(), stamps.end(), motion_model_transaction.addInvolvedStamp);
```

Not sure if a bind is required for the last param. You could also add an overload to the transaction to add a range of involved stamps. But again, this assumes that there isn't some logic that I'm missing with `last_stamp`.

fuse_core/src/timestamp_manager.cpp
```diff
@@ -113,40 +121,40 @@ void TimestampManager::query(
...
+  transaction.merge(motion_model_transaction, update_variables);
 }
 
-TimestampManager::stamp_range TimestampManager::stamps() const
+TimestampManager::const_stamp_range TimestampManager::stamps() const
 {
-  return stamp_range(boost::make_transform_iterator(motion_model_history_.begin(), extractStamp),
-                     boost::make_transform_iterator(motion_model_history_.end(), extractStamp));
-}
+  auto extract_stamp = +[](const MotionModelHistory::value_type& element) -> const ros::Time&
```
ayrton04, line 155: The fancy `+` syntax might be worth a comment.

ayrton04 (approved):
LGTM.

fuse_core/include/fuse_core/graph.h
svwilliams: I noticed that as well. I definitely don't have a strong opinion, other than the current setup is inconsistent. Seems like I need to either:
(a) Move the typedefs in graph.h _inside_ the Graph class. This is what was done in the Transaction object, and it prevents any sort of name collision.
(b) Move all of these range definitions into a common file somewhere and include that whenever I need them.
Thoughts? Preferences?

ayrton04: I like (a) better. Presumably, (b) would require that the `types.h` file (or whatever) includes `constraint.h`, `transaction.h`, etc. If that's true, then might we not get into weird circular dependencies where including the `types.h` file?

fuse_core/src/timestamp_manager.cpp
svwilliams: A couple of things:
(a) For reasons that I don't fully buy into, std::set (and others) don't have a `back()` method. They do have an `rbegin()` method.
(b) For reasons that seemed important at the time, I am obfuscating the underlying container inside the Transaction object. The public API provides a forward-only range. I don't believe the forward-only range has an `rbegin()`...because then it would be a bidirectional range.

That explains why the code does what it does.

Whether it should be that way is a different question. I have the ability to return a bidirectional range. I have the ability to just return the set. Philosophically I like the interface-implementation separation the current setup provides. But I'm never going to actually _use_ that ability. It is far more important for the Graph, where I am just defining the interface, and the implementation is left up to derived classes.

Discuss.

ayrton04: Well, my initial guess was that I was missing something, and indeed I was. So I'd say leave it as-is, because I agree with your reasoning. :)

svwilliams: Yeah. I can't find a solution I like any better. I can do:
```
  std::for_each(
    stamps.begin(),
    stamps.end(),
    std::bind(&Transaction::addInvolvedStamp, &motion_model_transaction, std::placeholders::_1));
```
..but that doesn't look any cleaner. And I'd still need to find the last stamp in a separate pass. Strangely I can't find an existing algorithm that does that.

ayrton04: Yeah, I'd say keep it your way.

ayrton04 (approved):

PR #39, opened 2019-02-28
[RST-1625] stamped variable synchronizer

svwilliams:
Created a helper class that finds the latest common timestamp for a set of stamped variable types.

A bit of background:
In `fuse`, there is a category of plugins called "Publishers". These publishers are provided with (a) a Transaction object consisting of the variable additions and deletions that occurred since the last update, and (b) the full SLAM graph containing all variables. The publisher then uses that information however it wants to generate ROS messages. A publisher could publish just the latest pose, or the full robot path, or even build an occupancy grid.

The variable types in `fuse` represent "pieces" of what one might think of as the full state. We have a Position2D variable, an Orientation2D variable, a LinearVelocity2D variable...you get the idea. Because the state is split across multiple variables, it is often necessary to "reassemble" the full state. The StampedVariableSynchronizer is designed for this "reassembly" task. You provide the object with a set of variable _types_ that should all have the same timestamp, and it processes the Transaction and Graph objects to find the latest common timestamp shared by the requested variable types.

Because we must reason about an arbitrary number of types, this uses all sorts of annoying and incomprehensible variadic template magic. You're welcome.

### DIFF
--- a/fuse_publishers/include/fuse_publishers/pose_2d_publisher.h
+++ b/fuse_publishers/include/fuse_publishers/pose_2d_publisher.h
@@ -34,11 +34,16 @@
 #ifndef FUSE_PUBLISHERS_POSE_2D_PUBLISHER_H
 #define FUSE_PUBLISHERS_POSE_2D_PUBLISHER_H
 
+#include <fuse_publishers/stamped_variable_synchronizer.h>
+
 #include <fuse_core/async_publisher.h>
 #include <fuse_core/graph.h>
 #include <fuse_core/macros.h>
 #include <fuse_core/transaction.h>
 #include <fuse_core/uuid.h>
+#include <fuse_variables/position_2d_stamped.h>
+#include <fuse_variables/orientation_2d_stamped.h>
+
 #include <ros/ros.h>
 #include <tf2_ros/transform_broadcaster.h>
 #include <tf2_ros/transform_listener.h>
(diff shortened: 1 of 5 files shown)


fuse_publishers/include/fuse_publishers/pose_2d_publisher.h
```diff
@@ -34,11 +34,16 @@
...
 #define FUSE_PUBLISHERS_POSE_2D_PUBLISHER_H
 
+#include <fuse_publishers/stamped_variable_synchronizer.h>
+
 #include <fuse_core/async_publisher.h>
 #include <fuse_core/graph.h>
 #include <fuse_core/macros.h>
 #include <fuse_core/transaction.h>
 #include <fuse_core/uuid.h>
+#include <fuse_variables/position_2d_stamped.h>
```
ayrton04: New ones aren't alphabetical, if you care.

fuse_publishers/include/fuse_publishers/stamped_variable_synchronizer.h
```diff
@@ -0,0 +1,321 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2018, Locus Robotics
```
ayrton04: 2019

ayrton04 (approved):
This looks like it was fun.

paulbovbel (approved):

PR #40, opened 2019-04-08
[RST-1926] extend local param definition

svwilliams:
This PR extends the definition of a LocalParameterization to include a `Minus()` function, the conceptual inverse of the `Plus()` function. This is a prerequisite for marginalization support.

The Orientation2D and Orientation3D variable classes will be temporarily broken. An upcoming PR (#41) will fix those classes up separately. That means some of the orientation/pose unit tests will be broken.

Scope Creep:
cc91396 - I was cleaning up the Orientation2D class and ended up breaking the Path2D publisher. It relied on non-standard functionality of the variable classes. So I ended up fixing the Path2D publisher in the PR as well.

### DIFF
--- a/fuse_core/include/fuse_core/autodiff_local_parameterization.h
+++ b/fuse_core/include/fuse_core/autodiff_local_parameterization.h
@@ -75,18 +75,12 @@ public:
   /**
    * @brief Constructs new PlusFunctor and MinusFunctor instances
    */
-  AutoDiffLocalParameterization() :
-    plus_functor_(new PlusFunctor()),
-    minus_functor_(new MinusFunctor())
-  {}
+  AutoDiffLocalParameterization();
 
   /**
    * @brief Takes ownership of the provided PlusFunctor and MinusFunctor instances
    */
-  AutoDiffLocalParameterization(PlusFunctor* plus_functor, MinusFunctor* minus_functor) :
-    plus_functor_(plus_functor),
-    minus_functor_(minus_functor)
-  {}
+  AutoDiffLocalParameterization(PlusFunctor* plus_functor, MinusFunctor* minus_functor);
 
   /**
    * @brief Generalization of the addition operation, implemented by the provided PlusFunctor
@@ -99,10 +93,7 @@ public:
   bool Plus(
     const double* x,
     const double* delta,
-    double* x_plus_delta) const override
-  {
-    return (*plus_functor_)(x, delta, x_plus_delta);
-  }
+    double* x_plus_delta) const override;
 
   /**
    * @brief The Jacobian of Plus(x, delta) w.r.t delta at delta = 0, computed using automatic differentiation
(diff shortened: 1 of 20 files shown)


fuse_core/include/fuse_core/autodiff_local_parameterization.h
```diff
@@ -0,0 +1,222 @@
...
+ * To get an auto differentiated local parameterization, you must define two classes with a templated operator()
+ * (a.k.a. a functor).
+ *
+ * The first functor should compute:
+ *
+ *   Plus(x, delta) -> x_plus_delta
+ *
+ * And the second functor should compute the inverse operation:
+ *
+ *   Minus(x1, x2) -> delta, such that Minus(x, Plus(x, delta)) = delta
```
ayrton04: Would this be clearer to say "such that Plus(x1, delta) = x2"?

fuse_core/include/fuse_core/autodiff_local_parameterization.h
```diff
@@ -0,0 +1,222 @@
...
+ * (e.g. a double-precision floating point number).
+ *
+ * Additionally the GlobalSize and LocalSize must be specified as template parameters.
+ * - GlobalSize is the size of the variables x1 and x2. If this is a quaternion, the GloblaSize would be 4.
+ * - LocalSize is the size of delta, and may be different from GlobalSize. For quaternions, there are only 3 degrees
+ *   of freedom, so the LocalSize is 3.
+ *
+ * For more information on local parameterizations, see fuse_core::LocalParameterization
+ */
+template <typename PlusFunctor, typename MinusFunctor, int kGlobalSize, int kLocalSize>
```
ayrton04, line 75: I assume the variable naming scheme is for consistency with Ceres.

fuse_core/include/fuse_core/autodiff_local_parameterization.h
```diff
@@ -0,0 +1,222 @@
...
+  return (*plus_functor_)(x, delta, x_plus_delta);
+}
+
+template <typename PlusFunctor, typename MinusFunctor, int kGlobalSize, int kLocalSize>
+bool AutoDiffLocalParameterization<PlusFunctor, MinusFunctor, kGlobalSize, kLocalSize>::ComputeJacobian(
+  const double* x,
+  double* jacobian) const
+{
+  double zero_delta[kLocalSize];
+  for (int i = 0; i < kLocalSize; ++i)
```
ayrton04: Doubt it's any faster, but can you use `std::fill` here?

```
std::fill(zero_delta, zero_delta + kLocalSize, 0.0);
```

ayrton04 (approved):

fuse_core/include/fuse_core/autodiff_local_parameterization.h
svwilliams: I was going to drop the leading `k`, but `ceres::LocalParameterization` defines `GlobalSize()` and `LocalSize()` methods. That means the template names cannot also be `GlobalSize` and `LocalSize`. Since I didn't have any better ideas, I went with the `kGlobalSize` and `kLocalSize` from the Ceres implementation. Happy to change it if you have a better suggestion.

ayrton04: Nah, that's fine by me.

fuse_core/include/fuse_core/autodiff_local_parameterization.h
svwilliams: These are actually zero-initialized by the compiler. Removing the initialization entirely.
https://en.cppreference.com/w/cpp/language/zero_initialization
> double f[3]; // zero-initialized to three 0.0's

ayrton04: Oh, sweet.

svwilliams: Actually, maybe I'm lying. That's only for static/global-scope.

svwilliams: But empty brace-initialization does work.
`double zero_delta[kLocalSize] = {};`

ayrton04: Yay C++14.

PR #41, opened 2019-04-08
[RST-1927] update orientation variables

svwilliams:
Updated the LocalParameterization classes for Orientation2D and Orientation3D
* Added the required `Minus()` and `ComputeMinusJacobian()` functions to the local parameterizations
* Switched from auto-diff to analytic derivatives. The derivative implementations are trivial, we might as well implement them directly.
* Updated the Orientation3D local parameterization.
* Added a missing Relative Orientation3D constraint

The third bullet deserves further explanation. Instead of the Ceres QuaternionParameterization, we compute the tangent space parameterization using the Rodrigues/exponential map formulation. Namely, `R = exp([theta * k]_x)` where `[*]_x` is the skew-symmetric matrix, `theta` is the angle of rotation, and `k` is the unit-norm axis of rotation. Thus, the 3-parameter tangent space is formed from by multiplying the axis of rotation by the angle of rotation. As a result, you will now see some conversions to axis-angle form in the local parameterization. (Further note: Ceres's definition of axis-angle already multiplies the axis by the rotation angle...which is weird but convenient.)

Among the benefits of this parameterization are that strange factors of two that existed in various parts of the code base are removed, and the tangent-space covariance (which can be computed by Ceres) directly corresponds with the input covariance.


fuse_constraints/include/fuse_constraints/normal_delta_orientation_3d_cost_functor.h
```diff
@@ -0,0 +1,130 @@
...
+namespace fuse_constraints
+{
+
+/**
+ * @brief Implements a cost function that models a difference between 3D orientation variables (quaternion)
+ *
+ * The cost function is of the form:
+ *
+ *             ||                                 ||^2
+ *   cost(x) = ||A * AngleAxis(b^-1 * q1^-1 * q2) ||
```
ayrton04: Nit: space before `A`

fuse_constraints/src/relative_orientation_3d_stamped_constraint.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+fuse_core::Vector4d RelativeOrientation3DStampedConstraint::toEigen(const geometry_msgs::Quaternion& quaternion)
+{
+  fuse_core::Vector4d eigen_quaternion_vector;
+  eigen_quaternion_vector << quaternion.w, quaternion.x, quaternion.y, quaternion.z;
+  return eigen_quaternion_vector;
+}
+
+fuse_core::Matrix3d RelativeOrientation3DStampedConstraint::toEigen(const std::array<double, 9>& covariance)
```
ayrton04, line 114: Could this return a matrix that is an `Eigen::Map`? Would that be valuable at all?

fuse_variables/src/orientation_2d_stamped.cpp
```diff
@@ -31,14 +31,82 @@
...
+ *
+ * 2D orientations add and subtract in the "usual" way, except for the 2*pi rollover issue. This local parameterization
+ * handles the rollover. Because the Jacobians for this parameterization are always identity, we implement this
+ * parameterization with "analytic" derivatives, instead of using the Ceres's autodiff system.
+ */
+class Orientation2DLocalParameterization : public fuse_core::LocalParameterization
+{
+public:
+  // Define some necessary variations of PI
+  static constexpr double M_TWO_PI = 2 * M_PI;
```
ayrton04: I believe you meant `M_TAU`.

fuse_variables/src/orientation_2d_stamped.cpp
```diff
@@ -31,14 +31,82 @@
...
+
+  bool Minus(
+    const double* x1,
+    const double* x2,
+    double* delta) const override
+  {
+    // Compute the difference from x2 to x1
+    delta[0] = x2[0] - x1[0];
+    // Then handle the 2*Pi roll-over
+    delta[0] -= M_TWO_PI * std::floor((delta[0] + M_PI) / M_TWO_PI);
```
ayrton04: Could we make this logic a templated method somewhere?

```
delta[0] = normalizeRotation(x2[0] - x1[0]);
```

ayrton04 (approved):
LGTM. Not sure how we missed the relative orientation constraint.

fuse_constraints/src/relative_orientation_3d_stamped_constraint.cpp
svwilliams: I don't think so...at least not safely. The input _could_ be a temporary.

Based on the signature, something like this should be legal:
```
auto cov_matrix = toEigen({1, 0, 0,  0, 1, 0,  0, 0, 1});
```

 Using Eigen::Map would result in pointing at data that was deleted once the function returns.

fuse_variables/src/orientation_2d_stamped.cpp
svwilliams: We seem to have a slightly annoying problem; we have several utility functions defined in two different places:
https://github.com/locusrobotics/fuse/blob/devel/fuse_constraints/include/fuse_constraints/util.h
https://github.com/locusrobotics/fuse/blob/devel/fuse_variables/include/fuse_variables/util.h

For the moment, I'm making it worse by copying the missing functions from fuse_constraints/util.h to fuse_variables/util.h.

Maybe more them to fuse_core? Even though fuse_core doesn't use them?

svwilliams: Moved util.h to fuse_core

PR #47, opened 2019-04-15
[RST-1745] constraint list container

svwilliams:
Added a container that holds the list of constraints, organized by variable.

This is designed as a helper structure for the variable marginalization code. As such, the variables and constraints are assigned sequential indexes instead of UUIDs.


ayrton04 (approved):
Tidy!

PR #48, opened 2019-04-19
[RST-1745] marginalize function

svwilliams:
Created a function to marginalize out one or more variables from a graph.

This didn't seem like _that_ much work when I started, but I ended up having to recreate about half of GTSAM's inference library to make it happen.

There are also some copyright questions here:
(1) I use a SuiteSparse library function, but SuiteSparse does not ship with one of the standard methods of finding the library. So I copied the Ceres FindSuiteSparse.cmake file. Possibly adding the license files is all that I need to do here.
(2) The `linearize()` method **must** recreate exactly what Ceres does during linearization. If the Ceres function was publicly accessible, I would have just called it...but it isn't, so I can't. Instead I implemented a function that does the same thing by examining the source code in Ceres. Not sure how to handle that.


ayrton04:
What's the license type for Ceres?

svwilliams:
Not exactly sure. Looks BSD-like:
> Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions are met:
>
> * Redistributions of source code must retain the above copyright notice,
  this list of conditions and the following disclaimer.
> * Redistributions in binary form must reproduce the above copyright notice,
  this list of conditions and the following disclaimer in the documentation
  and/or other materials provided with the distribution.
> * Neither the name of Google Inc. nor the names of its contributors may be
  used to endorse or promote products derived from this software without
  specific prior written permission.

I don't think there is a license violation issue. More that I need to do _something_...add in the Google license to the package, make a disclaimer somewhere, ...

ayrton04 (approved):
Whew, that's an involved class.

svwilliams:
The FindSuiteSparse.cmake file that I incorporated from Google Ceres has the license text at the top of the file. Since this is a source distribution, I feel like that is sufficient to meet the license requirements.

I also added links to the original source files, a written disclaimer, and the license text to the one function where I explicitly referenced Ceres while coding. Maybe that is sufficient for the `linearize()` function.

PR #52, opened 2019-05-06
[RST-1747] fixed lag smoother

svwilliams:
Initial implementation of the fixed-lag smoother. I will do some performance testing in the future, which may result in some code refactoring. But for now, this initial versions seems to function as expected.


fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother.h
```diff
@@ -0,0 +1,267 @@
...
+ *  (1) new variables and constraints are added to the graph
+ *  (2) the augmented graph is optimized and the variable values are updated
+ *  (3) all motion models, sensors, and publishers are notified of the updated graph
+ *  (4) all variables older than "current time - lag duration" are marginalized out.
+ *
+ * Optimization is performed at a fixed frequency, controlled by the \p update_frequency parameter. If an optimization
+ * takes longer than the update period, a warning will be logged and the optimization will not be tried again until
+ * the next update period.
+ *
+ * Optimization is performed at a fixed frequency, controlled by the \p optimization_frequency parameter. Received
```
ayrton04, line 69: Repeat of the sentence on line 69, but with a different parameter name.

fuse_optimizers/include/fuse_optimizers/vector_queue.h
```diff
@@ -0,0 +1,203 @@
...
+    {
+      return key_comparitor_(lhs, rhs.first);
+    }
+
+  private:
+    Compare key_comparitor_;
+  };
+
+  Container container_;  //!< The container used to hold the elements
+  const Comparitor comparitor_;  //!< Compare functor used to compare value_type objects based on their key
```
ayrton04: I think it's "comparator."

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -0,0 +1,320 @@
...
+  private_node_handle_.param("lag_duration", lag_duration, default_lag_duration);
+  if (lag_duration <= 0)
+  {
+    ROS_WARN_STREAM("The requested lag_duration is <= 0. Using the default value (" <<
+                    default_lag_duration << "s) instead.");
+    lag_duration = default_lag_duration;
+  }
+  lag_duration_.fromSec(lag_duration);
+
+  double optimization_frequency;
```
ayrton04: We repeat this identical pattern a bunch of times. Maybe make a `getPositiveParam` method or something?

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -0,0 +1,320 @@
...
+    // If a shutdown is requested, exit now.
+    if (!optimization_running_ || !ros::ok())
+    {
+      break;
+    }
+    // Apply motion models
+    auto new_transactions = fuse_core::Transaction::make_shared();
+    processQueue(*new_transactions);
+    // Prepare for selecting the marginal variables
+    preprocessMarginalization(*new_transactions);
```
ayrton04: Just a bit of confusion about plurals here. The `preprocessMarginalization` takes a single `Transaction` object. A single transaction can contain multiple updates to the graph, but it's still one transaction. The description also makes it seem as though it's taking many transactions, but that would imply something like `std::vector<Transaction>`. Looking over the comments, it seems like in many places, we refer to a `Transaction` object, but then reference many transactions.

I do not feel strongly about this at all, though, so keep it as-is if it's a problem to change it.

ayrton04 (approved):
That's some pretty stuff. LGTM!

fuse_optimizers/src/fixed_lag_smoother.cpp
svwilliams: Yeah....I don't know. `new_transactions` is a single Transaction object that results from merging multiple queued Transaction objects into a single object. Does that make it plural or singular? If singular makes more sense to you, it is fine by me.

ayrton04: I don't really feel strongly about it. It's a "special" kind of object, because many transactions can be represented by a single transaction.

PR #60, opened 2019-06-04
[RST-2088] Fix bug causing bad marginal computations occasionally

svwilliams:
Fixed an indexing error when building the dense A matrix for QR.


ayrton04 (approved):
Yaaay

matthewbwhitaker (approved):

PR #61, opened 2019-06-13
[RST-2128] fixed lag smoother reset

svwilliams:
Added a "reset" service to the fixed-lag smoother. This reverts the smoother back to its original state, waiting for a transaction from the configured ignition sensors.

### DIFF
--- a/fuse_optimizers/src/fixed_lag_smoother.cpp
+++ b/fuse_optimizers/src/fixed_lag_smoother.cpp
@@ -117,6 +117,9 @@ FixedLagSmoother::FixedLagSmoother(
     optimization_period_,
     &FixedLagSmoother::optimizerTimerCallback,
     this);
+
+  // Advertise a service that resets the optimizer to its initial state
+  reset_service_server_ = private_node_handle_.advertiseService("reset", &FixedLagSmoother::resetServiceCallback, this);
 }
 
 FixedLagSmoother::~FixedLagSmoother()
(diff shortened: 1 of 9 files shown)


fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -268,6 +279,28 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction)
   }
 }
 
+bool FixedLagSmoother::resetServiceCallback(std_srvs::Empty::Request& req, std_srvs::Empty::Response& res)
```
ayrton04: Neither `req` nor `res` are used, so it might be good to make that clear here and in the header:

```
bool FixedLagSmoother::resetServiceCallback(std_srvs::Empty::Request&, std_srvs::Empty::Response&)
```

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -179,27 +186,31 @@ void FixedLagSmoother::optimizationLoop()
...
-    postprocessMarginalization(marginal_transaction);
-    // Note: The marginal transaction will not be applied until the next optimization iteration
-    // Log a warning if the optimization took too long
-    auto optimization_complete = ros::Time::now();
-    if (optimization_complete > optimization_deadline)
+    // Optimize
     {
-      ROS_WARN_STREAM("Optimization exceeded the configured duration by " <<
-                      (optimization_complete - optimization_deadline) << "s");
+      std::lock_guard<std::mutex> lock(optimization_mutex_);
```
ayrton04, line 191: I assume this is necessary because the reset request will come from another thread.

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -268,6 +279,28 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction)
...
+{
+  started_ = false;
+  start_time_ = ros::TIME_MAX;
+  optimization_request_ = false;
+  // Clear all pending transactions
+  {
+    std::lock_guard<std::mutex> lock(pending_transactions_mutex_);
+    pending_transactions_.clear();
+  }
+  // Clear the graph and marginal tracking states
```
ayrton04, line 292: And there's no weird state that we can get into from a transaction getting somehow inserted in between these two blocks? Just checking.

ayrton04 (approved):

fuse_optimizers/src/fixed_lag_smoother.cpp
svwilliams: Correct. Previously the `graph_` was only modified by the optimizer thread. Now I am modifying the `graph_` from the ROS callback thread as well.

fuse_optimizers/src/fixed_lag_smoother.cpp
svwilliams: I don't think so.
The ROS spinner is configured for a single thread. Since this is called from the ROS callback thread, no other ROS callbacks can fire while this is running. And thus new pending transactions cannot be added between blocks.

matthewbwhitaker (approved):

PR #62, opened 2019-06-14
Add allocate_shared and allocate_aligned_shared

efernandez:
`allocate_aligned_shared` passes an `Eigen::aligned_allocator<T>()` object
to `allocate_shared`, so we don't have to type `T` twice, e.g.:

``` c++
fuse_rl::unicycle_2d::StateKinematicConstraint::allocate_aligned_shared(...);

fuse_rl::unicycle_2d::StateKinematicConstraint::allocate_shared(
  Eigen::aligned_allocator<fuse_rl::unicycle_2d::StateKinematicConstraint:>(),
  ...);
```

We need this because we can't use `make_shared` for classes with the `operator new` overloaded to keep things aligned for `Eigen`. See http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1049

### DIFF
--- a/fuse_core/include/fuse_core/macros.h
+++ b/fuse_core/include/fuse_core/macros.h
@@ -58,6 +58,9 @@
 #include <memory>
 #include <string>
 
+// Required by make_aligned_shared, that uses Eigen::aligned_allocator<T>().
+#include <Eigen/Core>
+
 
 /**
  * Defines aliases and static functions for using the Class with smart pointers.
@@ -86,7 +89,9 @@
 /// Defines aliases and static functions for using the Class with shared_ptrs.
 #define SHARED_PTR_DEFINITIONS(...) \
   __SHARED_PTR_ALIAS(__VA_ARGS__) \
-  __MAKE_SHARED_DEFINITION(__VA_ARGS__)
+  __MAKE_SHARED_DEFINITION(__VA_ARGS__) \
+  __ALLOCATE_SHARED_DEFINITION(__VA_ARGS__) \
+  __ALLOCATE_ALIGNED_SHARED_DEFINITION(__VA_ARGS__)
 
 #define __SHARED_PTR_ALIAS(...) \
   using SharedPtr = std::shared_ptr<__VA_ARGS__>; \
@@ -98,6 +103,22 @@
   make_shared(Args && ... args) \
   { \
     return std::make_shared<__VA_ARGS__>(std::forward<Args>(args) ...); \
+  }
+
+#define __ALLOCATE_SHARED_DEFINITION(...) \
+  template<class Alloc, typename ... Args> \
+  static std::shared_ptr<__VA_ARGS__> \
+  allocate_shared(const Alloc& alloc, Args && ... args) \
+  { \
+    return std::allocate_shared<__VA_ARGS__>(alloc, std::forward<Args>(args) ...); \
+  }
+
+#define __ALLOCATE_ALIGNED_SHARED_DEFINITION(...) \
+  template<typename ... Args> \
+  static std::shared_ptr<__VA_ARGS__> \
+  allocate_aligned_shared(Args && ... args) \
+  { \
+    return std::allocate_shared<__VA_ARGS__>(Eigen::aligned_allocator<__VA_ARGS__>(), std::forward<Args>(args) ...); \
   }
 
 /// Defines aliases and static functions for using the Class with weak_ptrs.


ayrton04 (approved):

efernandez:
@svwilliams Just rebased this on top of `devel`, so you can test the branch as is.

efernandez:
@svwilliams Close this one in favour of https://github.com/locusrobotics/fuse/pull/65 ?

svwilliams:
Assuming #65 works correctly for you, yes.

efernandez:
#65 with https://github.com/locusrobotics/fuse_rl/pull/6 (after you rebased it on top of https://github.com/locusrobotics/fuse_rl/pull/8) doesn't show any crashes, so it LGTM.

However, after a few measurements, I get NaN in the state and consequently I get this crash:
``` bash
CHOLMOD error: invalid xtype
F0622 00:26:12.449174  1001 covariance_impl.cc:665] Check failed: 'permutation' Must be non NULL 
*** Check failure stack trace: ***
```

I've added some prints and it looks like a constraint is added with an all NaN sqrt information matrix, so it should be something wrong with my input measurements, but I think their covariances are always invertible. I still need to narrow that down a little bit more. What do you think?
``` bash
 - fuse_constraints::AbsolutePose2DStampedConstraint
  uuid: 0148be94-9414-4bd2-957f-636e78466ee4
  position variable: aa7fda58-55d2-5318-98e0-8acd7049b90e
  orientation variable: ec7ac097-4034-5bd6-84bd-033aa9eec0d3
  mean:         1.15         0.65 -4.03887e-09
  sqrt_info: -nan -nan -nan
   0 -nan -nan
   0    0 -nan
```

svwilliams:
Yeah, if the constraint is created with a NaN sqrt_info, it sounds like a problem with the input. Or with the computation of the sqrt info from the provided covariance and indices. Since this is an absolute constraint on a pose, did you add an additional sensor to your configuration? The ignition sensor creates absolute constraints on position and orientation separately. And the odometry sensor was configured to  create constraints on velocity only.

Also, I clearly need to add checks for invalid measurement input and complain when that happens.

efernandez:
@svwilliams @ayrton04 The input I have can't be wrong, because there are already checks for that in the code. It's the logic that converts from `full` to `partial` covariance that's wrong. I just found the issue, but I think the API must change to fix it.

The code in https://github.com/locusrobotics/fuse_rl/blob/devel/include/fuse_rl/common/sensor_proc.h#L143-L165 doesn't set off-diagonal elements when the position and orientation are set separately from https://github.com/locusrobotics/fuse_rl/blob/devel/include/fuse_rl/common/sensor_proc.h#L286-L302

It's easy to see (and I've already confirmed it with some prints in the code) that for a `3x3` matrix, when we use all measurements `x`, `y` and `yaw`, we only set the entries indicated with `p` (for the position part) and `o` (for the orientation part) below:
```bash
p p *
p p *
* * o
```

The `*` entries don't get set, so they're just garbage.

I think this is a design flaw, because just looking at the two loops, they clearly can't set the `*` entries if we call `appendPartialMeasurement` for position and orientation separately. They perform a "square" iteration pattern. I like that separating position and orientation makes things cleaner and code is re-used. But it seems to me that we can't do that. What do you think? I can help with a re-implementation if we agree on the changes required.

There's also a couple of indexing bugs in the code already:
* https://github.com/locusrobotics/fuse_rl/blob/devel/include/fuse_rl/common/sensor_proc.h#L156 should be `mean_partial(r + offset) = mean_full(source_row);`
* https://github.com/locusrobotics/fuse_rl/blob/devel/include/fuse_rl/common/sensor_proc.h#L162 should be `covariance_partial(r + offset, c + offset) = covariance_full(source_row, source_col);`

Indeed, for the `_partial` mean and covariance variables we shouldn't use `offset`, but something like `offset_partial`.

All this was making the absolute pose 2D constraint covariance wrong, so the sqrt info became all NaN. Also the mean and covariance `x` entry was being overwritten by the `yaw` one. In my config I was using an odometry_2d sensor model from which I take the position `x`, `y` and orientation `yaw` only.

ayrton04:
We should probably move this discussion to that repo, but I think one alternative would be to check if all the indices are specified (x, y, and yaw) and just use `pose_mean` and `pose_covariance` [here](https://github.com/locusrobotics/fuse_rl/blob/devel/include/fuse_rl/common/sensor_proc.h#L305) instead of the partial versions (and skip building the partial measurement altogether). But I haven't revisited the rest of the code to make sure that makes sense, and can't until Monday.

I also note that the method call right now appears to be using the full mean anyway. Need to revisit that code and check if that's a bug.

efernandez:
@ayrton04 Sure, I'll create a PR in `fuse_rl`.

It looks like I can use the existing API and simply merge all indices before calling `appendPartialMeasurement`. I'll also a a unit test.

Note that if we only use `x` and `yaw`, we still need the correlation between them set, which I don't think it can be done separately, because will get the items marked with `*` unset:
```bash
p *
* o
```

ayrton04:
Yeah, that method likely just needs to be refactored. r_l already solves this problem, so it shouldn’t be too hard here.

efernandez:
Closing in favour of #65, which solves the Eigen alignment issues and it's a lot more correct and elegant that my quick fix here. :clap: 

I've opened https://github.com/locusrobotics/fuse_rl/pull/9 for the `NaN` issues I was having when I was trying to use the pose + orientation part of the `odometry_2d` sensor model.

PR #63, opened 2019-06-19
[RST-2158] Fix underconstained error

svwilliams:
Lock the pending transaction queue before adding things to it. :man_facepalming:

Not sure how I skipped that. I do know better.


ayrton04 (approved):
D'oh!

PR #65, opened 2019-06-21
[RST-2144] Support proper Eigen memory alignment

svwilliams:
This is my proposal to fix the Eigen memory alignment errors indicated in PR #62. The basic problem is Eigen objects need to be aligned on specific memory barriers to support SSE and similar CPU optimizations. Eigen already handles most of this internally...except for classes and structs that contain fixed-sized Eigen objects. Such classes have two problems.
1. These objects need to implement a custom `new()` operator that properly aligns the Eigen objects in memory. To make this easy, Eigen provides a macro to do this (`EIGEN_MAKE_ALIGNED_OPERATOR_NEW`), it just needs to be called in the public region of the class definition.
2. The `std::make_shared` call does not use the class-defined `new()` operator for whatever reason. Instead, you need to use `std::allocate_shared`.

For more details on Eigen memory alignment, see [here](https://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html).

This PR:
* Creates a new macro `SMART_PTR_DEFINITIONS_WITH_EIGEN()` that (a) calls the `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` macro to create the required `new()` operator that allocates memory correctly, and (b) implements an alternative definition of `make_shared` that uses `std::allocate_shared` and some Eigen calls. By naming the alternative version identically to the normal version, client code can call `Derived::make_shared` regardless of whether a class contains Eigen objects or not.
* Creates a `FUSE_VARIABLE_DEFINITIONS_WITH_EIGEN` that is the parallel of the existing `FUSE_VARIABLE_DEFINITIONS` but uses `SMART_PTR_DEFINITIONS_WITH_EIGEN` instead of `SMART_PTR_DEFINITIONS`.
* Creates `FUSE_CONSTRAINT_DEFINITIONS` and `FUSE_CONSTRAINT_DEFINITIONS_WITH_EIGEN` that are the parallel of the `FUSE_VARIABLE_DEFINITIONS` macros, but for Constraints. These macros implement the `std::string type() const` and `Constraint::SharedPtr clone() const` methods of the Constraint class. This is something I've been meaning to do anyway. And since I have to touch most of the constraints anyway, this seemed like a good time to sneak it in.
* Starting with C++17, the standard `new()` operator accepts an alignment size. With this feature, all of the above gymnastics are not required. The defined macros check for "aligned new" support and disable all of the additional code when appropriate. To allow this to happen in all cases, a `FUSE_MAKE_ALIGNED_OPERATOR_NEW` is provided, which calls the Eigen version or not, depending on the availability of "aligned new" support.
* Updates all of the fuse classes with the correct macro commands.

In summary:
* If you have a class that contains fixed-sized Eigen objects, then:
  * Add `FUSE_MAKE_ALIGNED_OPERATOR_NEW()` if you do not need any other smart pointer definitions.
  * Add `SMART_PTR_DEFINITIONS_WITH_EIGEN` if you do want all of the smart pointer definitions, including `Derived::make_shared()`. (Or `FUSE_CONSTRAINT_DEFINTIONS_WITH_EIGEN` if this is derived Constraint class. Or `FUSE_VARIABLE_DEFINTIONS_WITH_EIGEN` if this is derived Variable class)
* If you are creating smart pointers to fuse objects, then:
  * Always use the `Derived::make_shared()` implementation instead of the `std::make_shared<Derived>` formulation. The static member function will ensure the memory is properly aligned.

I have tested with Enrique's bagfile data using gcc-5.5, gcc-7.4, and gcc-7.4 with the `-faligned-new` flag. The bag data runs without error in all cases. All tests used the `-mavx` flag to trigger the alignment errors.

### DIFF
--- a/fuse_constraints/include/fuse_constraints/absolute_pose_2d_stamped_constraint.h
+++ b/fuse_constraints/include/fuse_constraints/absolute_pose_2d_stamped_constraint.h
@@ -63,7 +63,7 @@ namespace fuse_constraints
 class AbsolutePose2DStampedConstraint : public fuse_core::Constraint
 {
 public:
-  FUSE_CONSTRAINT_DEFINITIONS_WITH_EIGEN(AbsolutePose2DStampedConstraint);
+  FUSE_CONSTRAINT_DEFINITIONS(AbsolutePose2DStampedConstraint);
 
   /**
    * @brief Create a constraint using a measurement/prior of the 2D pose
--- a/fuse_constraints/include/fuse_constraints/absolute_pose_2d_stamped_constraint.h
+++ b/fuse_constraints/include/fuse_constraints/absolute_pose_2d_stamped_constraint.h
@@ -63,7 +63,7 @@ namespace fuse_constraints
 class AbsolutePose2DStampedConstraint : public fuse_core::Constraint
 {
 public:
-  SMART_PTR_DEFINITIONS(AbsolutePose2DStampedConstraint);
+  FUSE_CONSTRAINT_DEFINITIONS_WITH_EIGEN(AbsolutePose2DStampedConstraint);
 
   /**
    * @brief Create a constraint using a measurement/prior of the 2D pose
@@ -129,15 +129,6 @@ public:
    * @param[out] stream The stream to write to. Defaults to stdout.
    */
   void print(std::ostream& stream = std::cout) const override;
-
-  /**
-   * @brief Perform a deep copy of the constraint and return a unique pointer to the copy
-   *
-   * Unique pointers can be implicitly upgraded to shared pointers if needed.
-   *
-   * @return A unique pointer to a new instance of the most-derived constraint
-   */
-  fuse_core::Constraint::UniquePtr clone() const override;
 
   /**
    * @brief Construct an instance of this constraint's cost function
--- a/fuse_constraints/include/fuse_constraints/normal_delta_pose_2d_cost_functor.h
+++ b/fuse_constraints/include/fuse_constraints/normal_delta_pose_2d_cost_functor.h
@@ -73,8 +73,6 @@ namespace fuse_constraints
 class NormalDeltaPose2DCostFunctor
 {
 public:
-  FUSE_MAKE_ALIGNED_OPERATOR_NEW();
-
   /**
    * @brief Constructor
    *
--- a/fuse_constraints/include/fuse_constraints/normal_delta_pose_2d_cost_functor.h
+++ b/fuse_constraints/include/fuse_constraints/normal_delta_pose_2d_cost_functor.h
@@ -35,6 +35,7 @@
 #define FUSE_CONSTRAINTS_NORMAL_DELTA_POSE_2D_COST_FUNCTOR_H
 
 #include <fuse_core/eigen.h>
+#include <fuse_core/macros.h>
 #include <fuse_core/util.h>
 
 #include <Eigen/Core>
@@ -72,6 +73,8 @@ namespace fuse_constraints
 class NormalDeltaPose2DCostFunctor
 {
 public:
+  FUSE_MAKE_ALIGNED_OPERATOR_NEW();
+
   /**
    * @brief Constructor
    *
--- a/fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
+++ b/fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
@@ -35,8 +35,8 @@
 #define FUSE_CONSTRAINTS_NORMAL_DELTA_POSE_3D_COST_FUNCTOR_H
 
 #include <fuse_constraints/normal_delta_orientation_3d_cost_functor.h>
-
 #include <fuse_core/eigen.h>
+#include <fuse_core/macros.h>
 #include <fuse_core/util.h>
 
 #include <ceres/rotation.h>
@@ -68,6 +68,8 @@ namespace fuse_constraints
 class NormalDeltaPose3DCostFunctor
 {
 public:
+  FUSE_MAKE_ALIGNED_OPERATOR_NEW();
+
   /**
    * @brief Constructor
    *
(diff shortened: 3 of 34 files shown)


fuse_core/include/fuse_core/macros.h
```diff
@@ -58,19 +58,49 @@
...
+  __MAKE_SHARED_DEFINITION(__VA_ARGS__) \
+  __WEAK_PTR_ALIAS(__VA_ARGS__) \
+  __UNIQUE_PTR_ALIAS(__VA_ARGS__) \
+  __MAKE_UNIQUE_DEFINITION(__VA_ARGS__)
 
 /**
- * Defines aliases only for using the Class with smart pointers.
+ * Defines smart pointer aliases and static functions for a class that contains fixed-sized Eigen member variables.
+ *
+ * Same as SMART_PTR_DEFINITIONS except it ensure that shared ptr memory is allocated with proper byte alignment.
```
efernandez: typo: `it ensure` -> `it ensures`

fuse_constraints/include/fuse_constraints/absolute_pose_2d_stamped_constraint.h
```diff
@@ -63,7 +63,7 @@ namespace fuse_constraints
 class AbsolutePose2DStampedConstraint : public fuse_core::Constraint
 {
 public:
-  SMART_PTR_DEFINITIONS(AbsolutePose2DStampedConstraint);
+  FUSE_CONSTRAINT_DEFINITIONS_WITH_EIGEN(AbsolutePose2DStampedConstraint);
```
efernandez: I think this one doesn't need to be aligned, so you should do `FUSE_CONSTRAINT_DEFINITIONS(AbsolutePose2DStampedConstraint);`

The `Eigen::Vector3d` attribute in https://github.com/locusrobotics/fuse/pull/65/files#diff-bd3a017d88c898ee2cdac33c00c10c0dR145 is not a fixed-size vectorizable Eigen object: https://eigen.tuxfamily.org/dox/group__TopicFixedSizeVectorizable.html

fuse_constraints/include/fuse_constraints/normal_delta_pose_2d_cost_functor.h
```diff
@@ -72,6 +73,8 @@ namespace fuse_constraints
 class NormalDeltaPose2DCostFunctor
 {
 public:
+  FUSE_MAKE_ALIGNED_OPERATOR_NEW();
```
efernandez: Remove, you don't need it for https://github.com/locusrobotics/fuse/pull/65/files#diff-82ac2fce42490aae65c9dfd69e4fd5eeR104 (see https://eigen.tuxfamily.org/dox/group__TopicFixedSizeVectorizable.html)

efernandez: It looks like the `Eigen::MatrixXd A_` attribute could have been of `Eigen::Matrix3d` type, in which case you do need this. But from the comment in the constructor the number of columns is variable.

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
```diff
@@ -68,6 +68,8 @@ namespace fuse_constraints
 class NormalDeltaPose3DCostFunctor
 {
 public:
+  FUSE_MAKE_ALIGNED_OPERATOR_NEW();
```
efernandez, line 71: Here you do need this, but it's curious that the residual weighting matrix is fixed and not variable here https://github.com/locusrobotics/fuse/pull/65/files#diff-0f64d467f76db777867c720f4bae56f7R94, as opposed to the `pose_2d` case above.

fuse_constraints/include/fuse_constraints/normal_prior_pose_2d_cost_functor.h
```diff
@@ -66,6 +67,8 @@ namespace fuse_constraints
 class NormalPriorPose2DCostFunctor
 {
 public:
+  FUSE_MAKE_ALIGNED_OPERATOR_NEW();
```
efernandez: As before, for the `normal_delta` version of `pose_2d`, since https://github.com/locusrobotics/fuse/pull/65/files#diff-e91722dbdc7432f8dac96adfee2b8dbbR92 is dynamic, you don't need this.

fuse_constraints/include/fuse_constraints/relative_pose_2d_stamped_constraint.h
```diff
@@ -61,7 +61,7 @@ namespace fuse_constraints
 class RelativePose2DStampedConstraint : public fuse_core::Constraint
 {
 public:
-  SMART_PTR_DEFINITIONS(RelativePose2DStampedConstraint);
+  FUSE_CONSTRAINT_DEFINITIONS_WITH_EIGEN(RelativePose2DStampedConstraint);
```
efernandez: This could just be `FUSE_CONSTRAINT_DEFINITIONS(RelativePose2DStampedConstraint);`. Not alignment required for https://github.com/locusrobotics/fuse/pull/65/files#diff-f8a725e9aedd9304b2f3a77a8d873757R147

efernandez (approved):

fuse_constraints/include/fuse_constraints/absolute_pose_2d_stamped_constraint.h
svwilliams: Didn't pick up on the importance of the "vectorable" keyword before. Thanks.

fuse_constraints/include/fuse_constraints/normal_delta_pose_2d_cost_functor.h
svwilliams: The cost functor supports partial measurements, so the number of output residuals may not be 3. And thus `A_` will be a Nx3 matrix.
That said, there may be a less confusing way to implement this. I'm making a note to review these partial measurement constraints in the future.

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
svwilliams: For whatever reason we do not currently support partial measurements in the 3D case. Made a note to look into this as well.

efernandez (approved):
Just reviewed your changes after my feedback. Thanks. :+1:

ayrton04 (approved):

PR #70, opened 2019-06-25
Add print method to Graph and HashGraph

efernandez:
This allows to print the `HashGraph`, which is useful for debugging purposes.

### DIFF
--- a/fuse_graphs/src/hash_graph.cpp
+++ b/fuse_graphs/src/hash_graph.cpp
@@ -409,6 +409,22 @@ ceres::Solver::Summary HashGraph::optimize(const ceres::Solver::Options& options
   return summary;
 }
 
+void HashGraph::print(std::ostream& stream) const
+{
+  stream << "HashGraph\n"
+         << "  constraints:\n";
+  for (const auto& constraint : constraints_)
+  {
+    stream << "   - " << *constraint.second << "\n";
+  }
+  stream << "  variables:\n";
+  for (const auto& variable : variables_)
+  {
+    const auto is_on_hold = variables_on_hold_.find(variable.first) != variables_on_hold_.end();
+    stream << "   - " << (is_on_hold ? "[H] " : "") << *variable.second << "\n";
+  }
+}
+
 void HashGraph::createProblem(ceres::Problem& problem) const
 {
   // Add all the variables to the problem
--- a/fuse_graphs/src/hash_graph.cpp
+++ b/fuse_graphs/src/hash_graph.cpp
@@ -421,7 +421,9 @@ void HashGraph::print(std::ostream& stream) const
   for (const auto& variable : variables_)
   {
     const auto is_on_hold = variables_on_hold_.find(variable.first) != variables_on_hold_.end();
-    stream << "   - " << (is_on_hold ? "[H] " : "") << *variable.second << "\n";
+
+    stream << "   - " << *variable.second << "\n"
+           << "     on_hold: " << std::boolalpha << is_on_hold << "\n";
   }
 }
 
--- a/fuse_core/include/fuse_core/graph.h
+++ b/fuse_core/include/fuse_core/graph.h
@@ -44,6 +44,7 @@
 #include <ceres/covariance.h>
 #include <ceres/solver.h>
 
+#include <ostream>
 #include <utility>
 #include <vector>
 
@@ -262,7 +263,19 @@ public:
    * @return            A Ceres Solver Summary structure containing information about the optimization process
    */
   virtual ceres::Solver::Summary optimize(const ceres::Solver::Options& options = ceres::Solver::Options()) = 0;
+
+  /**
+   * @brief Print a human-readable description of the graph to the provided stream.
+   *
+   * @param[out] stream The stream to write to. Defaults to stdout.
+   */
+  virtual void print(std::ostream& stream = std::cout) const = 0;
 };
+
+/**
+ * Stream operator for printing Graph objects.
+ */
+std::ostream& operator <<(std::ostream& stream, const Graph& graph);
 
 }  // namespace fuse_core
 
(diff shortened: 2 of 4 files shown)


fuse_graphs/src/hash_graph.cpp
```diff
@@ -409,6 +409,22 @@ ceres::Solver::Summary HashGraph::optimize(const ceres::Solver::Options& options
   return summary;
 }
 
+void HashGraph::print(std::ostream& stream) const
+{
+  stream << "HashGraph\n"
+         << "  constraints:\n";
+  for (const auto& constraint : constraints_)
+  {
+    stream << "   - " << *constraint.second << "\n";
```
efernandez, line 418: Is there any way to pass some extra indentation to `*constraint.second`? (same for `*variable.second` below)

The main motivation to have proper indentation is that it looks like we could get YAML compliant outputs that could be use to visualize the graph with graphviz if we create a YAML to dot translator. I wonder if that was your plan?

fuse_graphs/src/hash_graph.cpp
```diff
@@ -409,6 +409,22 @@ ceres::Solver::Summary HashGraph::optimize(const ceres::Solver::Options& options
...
+  }
+  stream << "  variables:\n";
+  for (const auto& variable : variables_)
+  {
+    const auto is_on_hold = variables_on_hold_.find(variable.first) != variables_on_hold_.end();
+    stream << "   - " << (is_on_hold ? "[H] " : "") << *variable.second << "\n";
```
efernandez: For variables on hold I'm writting `[H] `, but I wonder if I could instead add another "field" like `on_hold: true` (or `false`). That way it'd be complaint with YAML, which seems to be the format you're trying to follow in all your outputs.

The downside though, is that it'd be harder to see if a variables it's on hold because the `on_hold` field would be at the bottom of the variable output. But for automated tools it'd be perfectly fine.

fuse_graphs/src/hash_graph.cpp
```diff
@@ -409,6 +409,22 @@ ceres::Solver::Summary HashGraph::optimize(const ceres::Solver::Options& options
   return summary;
 }
 
+void HashGraph::print(std::ostream& stream) const
+{
+  stream << "HashGraph\n"
+         << "  constraints:\n";
+  for (const auto& constraint : constraints_)
```
efernandez, line 416: For now I simply print the constraints and variables, but I'm open to suggestions regarding what we print here. There's another map that has variables and constraints pairs that could be printed as well, either replacing what it's printed now, or maybe printing all (constraints list + variables list + variables to constraints list mapping).

fuse_graphs/src/hash_graph.cpp
svwilliams: >Is there any way to pass some extra indentation to *constraint.second? (same for *variable.second below)

Not currently. We could modify the `print()` function to accept an indentation string. The problem with that being that there is no way to set that parameter when using the stream operator. We could drop the stream operator in favor of a `toString()` method. Then you could:
```
std::cout << constraint.toString("    ") << std::endl;
```

> we could get YAML compliant outputs

That was not something I had thought of. If the output looks like YAML, it is pure coincidence. Returning valid YAML is interesting. The one downside I see is formatting the matrices. I was planning on updating the covariance output to use Eigen's [IOFormatter](https://eigen.tuxfamily.org/dox/structEigen_1_1IOFormat.html), to get pretty looking matrices. But it may be possible to create pretty valid YAML matrices and we both win.

fuse_graphs/src/hash_graph.cpp
svwilliams: I added the "on hold" feature because Ceres supported it. However, it did not do what I thought it would do. It acts as though you placed a prior with complete certainty on the "on hold" variables. This results in the variable values being computed as you might expect, but the uncertainty on the "on hold" variables is zero. As such, I'm not sure what the utility of holding a variable is. So, I wouldn't worry overly about the "on hold" being hard to see. I'm not sure how often it will ever get used.

fuse_graphs/src/hash_graph.cpp
svwilliams: The variables and constraints are what I typically print as well. If anything, I'm wondering if this implementation should get moved to the `Graph` base class. We can use `getConstraints()` and `getVariables()` to access the collections in a generic way.

svwilliams (approved):
This works for me as-is. If we want to move to a formatted `toString()` in the future, we can do that in some future PR.

fuse_graphs/src/hash_graph.cpp
efernandez: Regarding `toString`, although it sounds like the only option, I'm afraid that returning an `std::string` would be slower than using streams, which IIRC would manage things efficiently in terms of memory and IO.

Re. Eigen matrix, I think there should be an IOFormatter that should be YAML compliant. I could give that idea a try on another PR.

fuse_graphs/src/hash_graph.cpp
efernandez: Thanks for the explanation. I've just changed this to the YAML compliant way: `on_hold: true/false`.

fuse_graphs/src/hash_graph.cpp
efernandez: I think that could be the generic implementation in the base class, but I think we should still allow the user to override the default/base implementation, which I think even allows extending it (by calling the base class implementation explicitly from the child class implementation).

efernandez:
@svwilliams Are you waiting for any action from my side to merge this? :smiley:

ayrton04:
He's waiting on my review. All of our PRs require two reviews. I'll try to get to it today.

ayrton04 (approved):

matthewbwhitaker (approved):

PR #71, opened 2019-06-27
Depend on libceres-dev instead of ceres-solver

efernandez:
There's no `ceres-solver` rosdep entry in the official rosdistro files.
Indeed there's one entry for Ceres solver, but it's called `libceres-dev`:

https://github.com/ros/rosdistro/blob/9cd7f875f62ee774789893da025f0322783c0fc6/rosdep/base.yaml#L1633

### DIFF
--- a/fuse_constraints/package.xml
+++ b/fuse_constraints/package.xml
@@ -12,7 +12,7 @@
   <license>BSD</license>
 
   <buildtool_depend>catkin</buildtool_depend>
-  <depend>ceres-solver</depend>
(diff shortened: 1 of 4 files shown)


svwilliams (approved):
Specifying the wrong thing in package.xml is obviously useless.

But I do want to mention the tire fire that is the Ubuntu ceres-solver packaging:
https://discourse.ros.org/t/ceres-package/872

paulbovbel (approved):

PR #72, opened 2019-06-28
Default private ~reset service name

efernandez:
This makes the service name be `/<NODE_NAME>/reset` by default, instead of just `/reset`.
If a namespace is use it'll prefix any of the cases above, so with `~reset` we'll get `/<NAMESPACE>/<NODE_NAME>/reset`.


svwilliams (approved):
Works for me. This is how I actually set the param in my launch files anyway.

matthewbwhitaker (approved):

PR #75, opened 2019-07-11
[RST-2148] reset all the things

svwilliams:
The fixed-lag smoother has a reset() service call. There is a desire to propagate the reset signal to all of the fuse plugins (motion models, sensor models, and publishers). This is being done by adding `start()` and `stop()` methods to the API of all the fuse plugins. Separate `start()` and `stop()` are used, instead of a single `reset()` method to make threading issues easier. All of the plugins are stopped, the internal optimizer state is modified, then all of the plugins are started. This ensures we never get a mix of pre-reset and post-reset information at the same time.

This PR:
* Adds `start()` and `stop()` methods to the MotionModel, SensorModel, and Publisher base classes. A default empty implementation is provided for cases when nothing special needs to be done.
* Adds `onStart()` and `onStop()` methods to the AsyncMotionModel, AsyncSensorModel, and AsyncPublisher. These methods are called from the plugin's spinner thread instead of in the optimizer's thread. The Async classes provide an implementation of `start()` and `stop()` that insert calls to `onStart()` and `onStop()`, respectively, into the spinner's callback queue. They then block until those callbacks have been executed.
* Updates the Optimizer base class to call `start()` on all plugins after initialization, and to call `stop()` on all plugins before destruction.
* Updates the FixedLagSmoother reset service callback to use the `start()` and `stop()` methods.
* Fixes a race condition in the FixedLagSmoother reset service callback. Previously the pending queue was locked, work moved to a local variable, then unlocked. And then the graph was locked, the local work was applied to the graph, and the graph was unlocked. However, it is possible (though unlikely) for the work to be moved to the local variable, then have the reset callback execute between the time the pending queue is unlocked and the graph is locked. This would result in the local work variable being applied to the newly reset graph, which is wrong. All of that work was from a time that pre-dates the reset. I believe I have solved this issue by changing the locking strategy a bit.
* Adds an ID number when inserting function calls into the ROS callback queues. This allows the function calls to be removed later, if needed.

Misc changes:
* Alphabetizes some function names in the header, if I happened to be thinking about it at the time.
* Adds a clear() method to the  TimestampManager
* Changes the AsyncXXX function implementations from `final` to merely `override`. No real reason to prevent people from tweaking things when needed.

This PR description is now almost as long as the actual changes included in the PR. :)

### DIFF
--- a/fuse_core/include/fuse_core/async_motion_model.h
+++ b/fuse_core/include/fuse_core/async_motion_model.h
@@ -93,21 +93,18 @@ public:
   virtual ~AsyncMotionModel() = default;
 
   /**
-   * @brief Get the unique name of this motion model
-   */
-  const std::string& name() const final { return name_; }
-
-  /**
-   * @brief Perform any required post-construction initialization, such as subscribing to topics or reading from the
-   * parameter server.
+   * @brief Augment a transaction object such that all involved timestamps are connected by motion model constraints.
    *
-   * This will be called for each plugin after construction and after the ros node has been initialized. The provided
-   * private node handle will be in a namespace based on the plugin's name. This should prevent conflicts and allow
-   * the same plugin to be used multiple times with different settings and topics.
+   * This method will be called by the optimizer, in the optimizer's thread, before each sensor transaction is applied
+   * to the Graph. This implementation packages a call to the pure virtual method applyCallback() and inserts it into
+   * this motion model's local callback queue. This allows the applyCallback() function to be executed from the same
+   * thread as any other configured callbacks. Despite the fact that the queryCallback() function call runs in a
+   * different thread than this function, this function blocks until the query callback returns.
    *
-   * @param[in] name A unique name to give this plugin instance
+   * @param[in,out] transaction The transaction object that should be augmented with motion model constraints
+   * @return                    True if the motion models were generated successfully, false otherwise
    */
-  void initialize(const std::string& name) final;
+  bool apply(Transaction& transaction) final;
 
   /**
    * @brief Function to be executed whenever the optimizer has completed a Graph update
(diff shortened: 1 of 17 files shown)


fuse_core/include/fuse_core/async_motion_model.h
```diff
@@ -93,9 +93,32 @@ class AsyncMotionModel : public MotionModel
...
+   * This method will be called by the optimizer, in the optimizer's thread, before each sensor transaction is applied
+   * to the Graph. This implementation packages a call to the pure virtual method applyCallback() and inserts it into
+   * this motion model's local callback queue. This allows the applyCallback() function to be executed from the same
+   * thread as any other configured callbacks. Despite the fact that the queryCallback() function call runs in a
+   * different thread than this function, this function blocks until the query callback returns.
+   *
+   * @param[in,out] transaction The transaction object that should be augmented with motion model constraints
+   * @return                    True if the motion models were generated successfully, false otherwise
+   */
+  bool apply(Transaction& transaction) override;
```
svwilliams, line 107: Just alphabetizing things. `start()` and `stop()` have been added. Everything else has just moved.

fuse_core/include/fuse_core/async_motion_model.h
```diff
@@ -107,35 +130,44 @@ class AsyncMotionModel : public MotionModel
    *
    * @param[in] name A unique name to give this plugin instance
    */
-  void initialize(const std::string& name) final;
+  void initialize(const std::string& name) override;
```
svwilliams, line 133: Trying to be a bit less heavy-handed.

fuse_core/src/async_motion_model.cpp
```diff
@@ -64,15 +64,16 @@ bool AsyncMotionModel::apply(Transaction& transaction)
   auto callback = boost::make_shared<CallbackWrapper<bool>>(
     std::bind(&AsyncMotionModel::applyCallback, this, std::ref(transaction)));
   auto result = callback->getFuture();
-  callback_queue_.addCallback(callback);
+  callback_queue_.addCallback(callback, reinterpret_cast<uint64_t>(this));
```
svwilliams, line 67: Adding an ID to the callbacks. This allows the callbacks to be deleted, if needed.

efernandez: Just curious: the `reinterpret_cast<uint64_t>(this)` ID would be the same for all the callbacks added by `this` object. Is this intended. From the `callback_queue.cpp` it looks like it'll add all the callbacks and `removeByID` will remove all the callbacks with the given ID.

efernandez (approved):

matthewbwhitaker (approved):

efernandez:
See https://github.com/locusrobotics/fuse_rl/pull/19#issuecomment-510892659

@svwilliams  I wonder if you're getting a similar crash. Are you compiling with `-mavx`?

efernandez:
Sorry for the false alarm @svwilliams , it looks good. See https://github.com/locusrobotics/fuse_rl/pull/19#issuecomment-510897129

ayrton04 (approved):

fuse_core/src/async_motion_model.cpp
svwilliams: Any callbacks added to the callback queue by ROS (e.g. message subscription callbacks, service callbacks, timer callbacks) are managed internally by ROS and each have their own ID number. All of the callbacks added to the callback queue "manually" are now assigned a known ID number, but it is the same for all of the callbacks. For the AsyncMotionModel, this will be calls to `applyCallback()`, `onGraphUpdate()`, `onStart()`, and `onStop()`. At the moment, I have no particular need to remove the callbacks from, say `onStart()`, separately from any of the others. If you have a use-case in mind where clearing individual callbacks types would be valuable, I can definitely assign different IDs to them. Though `onStart()` and `onStop()` should probably be treated as the same "unit".

efernandez: I don't have a use case for removing individual callbacks atm. It's fine as is. I just wanted to confirm that potential (but unlikely) limitation. Thanks. :smiley:

PR #76, opened 2019-07-11
Comment unused parameters

efernandez:
Otherwise the compilation fails with: `-Werror=unused-parameter`
This happens with these flags: `-Wall -Wextra`


svwilliams (approved):
@ayrton04 Do you have any particular syntax for writing the would-be variable name? Or are you good with `Transaction::ConstSharedPtr /*transaction*/`?

ayrton04:
I don't feel strongly about it. I tend to just leave them as

```
void method(const Whatever&)
```

...but I'm indifferent.

ayrton04 (approved):

efernandez:
The reason I commented the unused parameters instead of removing them was because the doxygen block references them, which I think it's a good reason against removing them, I guess. Either way, I also don't have a strong opinion.

PR #77, opened 2019-07-11
Fix -Wall -Wextra warnings

efernandez:
Continuation of https://github.com/locusrobotics/fuse/pull/76 because I forgot to `catkin clean -y` and not everything was compiled with those flags. This time I cleaned everything and re-built.

I only had to fixed two types of warnings/errors:
* `unused-parameter`, as in https://github.com/locusrobotics/fuse/pull/76 . I commented the parameters, but if you want I can remove them for all instances, or for specific ones.
* `ignored-qualifiers`, for functions/methods returning by value that had `const`. I removed the `const` for all of them because they were returning POD types. If the intend was to return a const reference instead, please let me know.

### DIFF
--- a/fuse_constraints/include/fuse_constraints/uuid_ordering.h
+++ b/fuse_constraints/include/fuse_constraints/uuid_ordering.h
@@ -133,7 +133,7 @@ public:
    *
    * Accessing a UUID that does not exist results in the provided UUID being added to the ordering
    */
-  const unsigned int operator[](const fuse_core::UUID& uuid);
+  unsigned int operator[](const fuse_core::UUID& uuid);
 
   /**
    * @brief Access the UUID stored at the provided index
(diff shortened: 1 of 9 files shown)


svwilliams (approved):
Man the compiler is picky. I need to revisit my own workspace settings. I thought I had things configured for `-Wall -Wextra`.

efernandez:
Yep, @svwilliams . I indeed thought I had those flags in my `catkin config`, but I didn't. They're very picky, I agree, but they detect things like not returning from a non-`void` function, which happened to me in some other unrelated code I have in my workspace. This might be also related with the old `gcc` version I'm using (`5.5.0`), but a newer compiler will likely complain of all this issues, and in my experience if you try `clang` it'd be even pickier. :smiley:

ayrton04 (approved):

PR #78, opened 2019-07-16
Expose Ceres Solver, Problem and Covariance Options as ROS parameters

efernandez:
This allows to set all `ceres::Solver::Options`, `ceres::Problem::Options` and `ceres::Covariance::Options` using ROS parameters.

The motivation is to tune the solver and problem for better performance or accuracy, after looking at the tuning results shown in the plots at https://github.com/SteveMacenski/slam_toolbox , which was shared with me by @svwilliams . Unfortunately, it looks like the best configuration is almost the same as the default one :laughing: , except for the `preconditioner_type` option, which defaults to `JACOBI` but it's set to `SCHUR_JACOBI`, which indeed make things a little faster for me (from 55 to 49% CPU with my `fuse_rl` configuration).

Either way, I think this is useful to tune things with other goals in mind, like accuracy, debugging, and still performance. TBH I haven't played a lot with the options/parameters, other than confirming things are being set and used internally.

I also suspect the `enable_fast_removal` option doesn't make a big different for the `ceres::Problem` because I think it's built from scratch every time.

I took the list of options from the latest version of Ceres, which will be `2.0.0`, but also supported `1.13.x`:
* https://github.com/ceres-solver/ceres-solver/blob/master/include/ceres/solver.h#L59
* https://github.com/ceres-solver/ceres-solver/blob/master/include/ceres/problem.h#L124

### DIFF
--- a/fuse_core/include/fuse_core/ceres_options.h
+++ b/fuse_core/include/fuse_core/ceres_options.h
@@ -1,8 +1,36 @@
-/***************************************************************************
- * Copyright (C) 2019 Clearpath Robotics. All rights reserved.
- * Unauthorized copying of this file, via any medium, is strictly prohibited
- * Proprietary and confidential
- ***************************************************************************/
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2019 Clearpath Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
 #ifndef FUSE_CORE_CERES_OPTIONS_H
 #define FUSE_CORE_CERES_OPTIONS_H
 
--- a/fuse_core/include/fuse_core/ceres_options.h
+++ b/fuse_core/include/fuse_core/ceres_options.h
@@ -1,0 +1,188 @@
+/***************************************************************************
+ * Copyright (C) 2019 Clearpath Robotics. All rights reserved.
+ * Unauthorized copying of this file, via any medium, is strictly prohibited
+ * Proprietary and confidential
+ ***************************************************************************/
+#ifndef FUSE_CORE_CERES_OPTIONS_H
+#define FUSE_CORE_CERES_OPTIONS_H
+
+#include <ros/console.h>
+#include <ros/node_handle.h>
+
+#include <ceres/types.h>
+
+#include <string>
+
+/**
+ * Defines ToString overloaded function to Ceres Options.
+ *
+ * For a given Ceres Solver Option <T>, the function ToString calls ceres::<T>ToString
+ */
+#define CERES_OPTION_TO_STRING_DEFINITION(Option) \
+  static inline const char* ToString(ceres::Option value) \
+  { \
+    return ceres::Option##ToString(value); \
+  }
+
+/**
+ * Defines FromString overloaded function to Ceres Options.
+ *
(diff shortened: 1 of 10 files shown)


fuse_core/include/fuse_core/macros.h
```diff
@@ -58,7 +58,7 @@
 #include <memory>
 #include <string>
 
-// Required by make_aligned_shared, that uses Eigen::aligned_allocator<T>().
+// Required by __MAKE_SHARED_ALIGNED_DEFINITION, that uses Eigen::aligned_allocator<T>().
```
efernandez, line 61: Simply updating the comment, because we finally don't use `make_aligned_shared`, which doesn't exist.

fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother_params.h
```diff
@@ -59,16 +60,23 @@ struct FixedLagSmootherParams
...
+
+  /**
+   * @brief Helper function that loads strictly positive integral or floating point values from the parameter server
    *
    * @param[in] node_handle - The node handle used to load the parameter
    * @param[in] parameter_name - The parameter name to load
    * @param[in] default_value - A default value to use if the provided parameter name does not exist
    * @return The loaded (or default) value
    */
-  double getPositiveParam(const ros::NodeHandle& node_handle, const std::string& parameter_name, double default_value)
```
efernandez, line 69: At the end I don't really use `getPositiveParam` for integral types, but I think this change is for good, although it'd be great to have this function in a more generic place. Either way, if you don't like this, it's in its own commit, so it's easy drop it from this PR.

svwilliams:
Sorry it is taking me so long to go through this one. I will eventually get some version of this merged, hopefully next week. It was on my ToDo list to support setting Ceres parameters, so thanks for beating me to it.

efernandez:
I'd check the continuous integration results to fix whatever is failing (likely `roslint` issues), but I'm afraid this is requesting me too much: :thinking: 

Screenshot from 2019-07-25 19-53-33

I'm not 100% sure if that will grant it access to the organizations I'm in, but I wonder why is that needed or what it actually does.

efernandez:
There shouldn't be any `roslint` issues now.

efernandez:
I've just rebased this PR and added the changes from https://github.com/locusrobotics/fuse_rl/pull/22 here.

FYI @ayrton04 @svwilliams

SteveMacenski:
@efernandez on your slam toolbox link, trust me the fact that nearly the defaults were the best is not lost on me. I probably spent 2 weeks thinking how much smarter I was than the Ceres guys trying out combinations to come crawling back. I learned alot so *shrug*. 

Anyhow reason I comment- there are some parameters I did not include as a "don't give them enough rope to hang themselves" I see that you've exposed parameters for both line search and trust region strategies, which for most applications (at least I'm imagining this being used for) line search seems to be a bad choice, and many of the parameters only do things in certain modes. Removing those would simplify the overwhelming number of choices to the things that people will actually want to focus in on.

efernandez:
Thanks for you comment @SteveMacenski . I really appreciate your  effort to analyse the different options. It's a shame though that defaults are almost the best combination, except fo the pre-conditioner, which is actually slightly faster with `SCHUR_JACOBI`. :+1: 

I decided to expose all options because I don't know if someone wants to play with any of them, but you're right that some options are only used for certain types of solvers. I'm happy to drop some options if that's always the case, or document which options are the most relevant ones. That way it'd be clear to the user what options are those by just looking at this package, although ceres already does a good job at documenting all options. :smiley:

SteveMacenski:
@efernandez fair enough -- just thought I'd mention it. I'm always a fan of giving every single little dial they can turn, and especially since this is supposed to be general, it makes sense. I looked at the structure and think "huh this really seems designed towards localization/slam/filtering" which fit trust region over line search categorically so there's an opportunity to simplify since about 50% of the parameters don't do anything in trust region. 

If nothing else Ceres docs exist and this thread :+1:

efernandez:
Not sure why the checks failed. I've rebased this on top of `devel` and run `make roslint` on all pkgs changed, and everything LGTM locally. Could you tell me what checks ares failing, please?

svwilliams:
```

**Comment by svwilliams** [comment]
This is for the Xenial build. There were no errors in the Bionic build.
```
/home/ec2-user/workspace/test_rosdistro_fuse_PR-78/package/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h:135:26: error: ‘struct ceres::Covariance::Options’ has no member named ‘sparse_linear_algebra_library_type’

[2019-08-14T16:56:49.923Z]        covariance_options.sparse_linear_algebra_library_type = fuse_core::getParam(
```
I'm guessing the default version of Ceres in Xenial doesn't have that parameter?

**Comment by efernandez** [comment]
Thanks @svwilliams . I just fixed it. The `sparse_linear_algebra_library_type` options was added to `ceres::Covariance::Options` in version `1.13.0`, so Ubuntu Xenial must be using and older version.

This commit introduced that changed: https://github.com/ceres-solver/ceres-solver/commit/14d8297cf968e421c5db4e3fb0543b3b111155d7 (also mentioned in the code now)

File: fuse_optimizers/src/fixed_lag_smoother_node.cpp [hunk 3]
```diff
@@ -32,14 +32,18 @@
...
 
 
 int main(int argc, char **argv)
 {
   ros::init(argc, argv, "fixed_lag_smoother_node");
-  fuse_optimizers::FixedLagSmoother optimizer(fuse_graphs::HashGraph::make_unique());
+  ros::NodeHandle private_node_handle("~");
+  fuse_graphs::HashGraphParams hash_graph_params;
+  hash_graph_params.loadFromROS(private_node_handle);
+  fuse_optimizers::FixedLagSmoother optimizer(fuse_graphs::HashGraph::make_unique(hash_graph_params.problem_options));
```
svwilliams: Since you made a HashGraphParams object, it would probably make sense to pass the whole struct to the HashGraph constructor, rather than just the `problem_options` field.

svwilliams (approved):
Sorry this took so long to approve. And what I wouldn't give for a Ceres-maintained INI file reader.

fuse_core/include/fuse_core/ceres_options.h
```diff
@@ -0,0 +1,188 @@
+/***************************************************************************
+ * Copyright (C) 2019 Clearpath Robotics. All rights reserved.
```
ayrton04: This license is probably going to have to change.

ayrton04 (approved):

fuse_core/include/fuse_core/ceres_options.h
efernandez: No problem. Would you guys change it before merging?

fuse_core/include/fuse_core/ceres_options.h
```diff
@@ -0,0 +1,188 @@
...
+
+}  // namespace ceres
+#else
+/**
+ * Patch Ceres version 2.0.0 that uses lower case for the LoggingType and DumpFormatType StringTo function.
+ */
+namespace ceres
+{
+
+bool StringtoLoggingType(std::string value, LoggingType* type)
```
efernandez: Oops, it looks like I have this the other way around. Fixing it.

efernandez: Fixed.

fuse_optimizers/src/fixed_lag_smoother_node.cpp
efernandez: Makes sense. Done.

efernandez:
Thanks, I just realized I removed the serialization headers by mistake while solving some conflicts after rebasing. :smiley:

PR #79, opened 2019-07-25
Stamp merged transaction

efernandez:
This sets the transaction's timestamp to the maximum stamp of the merged transactions.

Otherwise, the merged transaction keeps the stamp from `this` transaction. This means that if we merge `other` transactions into `this`, which is originally empty, the timestamp would be zero: `ros::Time(0)`. This actually happens in https://github.com/locusrobotics/fuse/blob/devel/fuse_optimizers/src/fixed_lag_smoother.cpp#L167, where the merged transaction `new_transaction` always have a zero stamp.

Probably not very important, but looks a bit unexpected.

I also wonder if instead of having a `stamp_` attribute we should just use `*involved_stamps_.crbegin()`, as long as `!involved_stamps_.empty()`.

I've created this PR because the changes are trivial, but feel free to consider this a discussion and close with no action or in favour of a different change.

### DIFF
--- a/fuse_core/src/transaction.cpp
+++ b/fuse_core/src/transaction.cpp
@@ -182,6 +182,7 @@ void Transaction::removeVariable(const UUID& variable_uuid)
 
 void Transaction::merge(const Transaction& other, bool overwrite)
 {
+  stamp_ = std::max(stamp_, other.stamp_);
   involved_stamps_.insert(other.involved_stamps_.begin(), other.involved_stamps_.end());
   for (const auto& added_constraint : other.added_constraints_)
   {
--- a/fuse_core/test/test_transaction.cpp
+++ b/fuse_core/test/test_transaction.cpp
@@ -666,6 +666,7 @@ TEST(Transaction, Merge)
   UUID removed_variable3 = fuse_core::uuid::generate();
 
   Transaction transaction1;
+  transaction1.stamp(involved_stamp2);
   transaction1.addInvolvedStamp(involved_stamp1);
   transaction1.addInvolvedStamp(involved_stamp2);
   transaction1.addConstraint(added_constraint1);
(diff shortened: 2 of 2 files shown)


svwilliams:
I like the idea of using the involved_stamps as the source for the Transaction stamp() method.
Currently the stamp() method is used to:
* Sort transactions in the queue so they are applied in order.
Basing this on the most recent involved stamp actually seems like the correct thing to do.
* Check if a transaction is too old and should be purged.
This one is less clear to me. For normal sensor measurements, the difference between when the sensor measurement occurred and when the transaction was created should be small. In such a case, using the involved stamp time is fine. For something like a loop closure sensor, there may be a noticeable lag between the current time and the most recent involved stamp. Such a sensor _might_ want to put a recent stamp on the transaction even though it involves older states. I'm honestly not sure.
* Check if a transaction is before the start time
This one is interesting. What I may want to do is check the _oldest_ stamp in the involved_stamps set, rather than the newest stamp or the separate transaction stamp.
* Also, conceivably a sensor may produce constraints that are not stamped at all. For example, a database of previously observed visual features. When a feature is observed by a camera, it may trigger a lookup into the database to add a prior position on the visual landmark. Since the visual features are assumed to be stationary for all time, they don't really have an involved timestamp. Having the extra field to store an "uninvolved" timestamp could be helpful. Again, I'm not exactly sure.

I am writing myself a ticket to look into how transaction timestamps are used. In the mean time, I'll get this PR applied so that merged transactions are correct.

svwilliams (approved):

matthewbwhitaker (approved):

matthewbwhitaker:
`fuse_core/test/test_transaction.cpp` needs `<algorithm>` including to keep the linter happy.

efernandez:
Thanks for the detailed comment @svwilliams . Certainly, I know see a good reason to have a `stamp_` field on its own.

I guess something like this could be implemented if we want to save the user from setting the stamp explicitly: `return stamp_.isZero() && !involved_stamps_.empty() ? *involved_stamps_.crbegin() : stamp_;`
This would be in the `stamp()` getter, so it'll return the `stamp_` field only if it's been explicitly set by the user. Otherwise it'll return the last / max involved timestamp.

But this is probably over-engineered. :smile:

PR #80, opened 2019-07-25
Fix -Wall -Wextra warnings in tests

efernandez:
I know this is quite pedantic, but since I already did the same for the rest of the code, I think it's better to get the tests done as well.

FYI All warnings were `-Werror=unused-parameter`.


svwilliams (approved):

matthewbwhitaker (approved):
Need to shorten `test_async_publisher.cpp:57` to below 120 characters to keep the linter happy.

PR #84, opened 2019-08-13
RST-2390 Moving fuse_rl to fuse_models

ayrton04:
FYI @efernandez


fuse_models/CHANGELOG.rst
```diff
@@ -0,0 +1,81 @@
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Changelog for package fuse_models
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+0.2.0 (2019-07-12)
+------------------
+* Get predict_to_current_time ROS param (`#17 <https://github.com/locusrobotics/fuse_rl/issues/17>`_)
```
ayrton04, line 13: I left the changelog links pointing to the `fuse_rl` repo so that those PRs can still be viewed.

svwilliams (approved):
I reviewed every single line. It looks perfect.

ayrton04:
You actually _did_ review every single line, at least at some point. I'll start moving these classes...

fuse_models/package.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>fuse_models</name>
+  <version>0.2.0</version>
```
efernandez, line 4: I think this version must match the one in the other packages in the repo.

efernandez:
BTW @ayrton04 If you want to keep `fuse_rl` history, you could try something like this: https://stackoverflow.com/questions/41811986/git-move-directory-to-another-repository-while-keeping-the-history

Let me know if you decide to do so and need help.

ayrton04:
Done.

ayrton04:
BTW, before I pushed this, I did a `git diff origin/RST-2390-move-fuse-rl` and it came up empty (if you ignore the changelog fix). Should be in the same state.

PR #83, opened 2019-08-12
[RST 2340] Serialization

efernandez:
This PR replaces https://github.com/locusrobotics/fuse/pull/81 by extending https://github.com/locusrobotics/fuse/tree/RST-2340-serialization-prototype (see [this comment](https://github.com/locusrobotics/fuse/pull/81#issuecomment-515706574) for a detailed description).

This is still WIP. Next steps are:

- [ ] Fix the crash described in https://github.com/locusrobotics/fuse/pull/81#issuecomment-520477374
- [ ] Support serialization in relative constraints and 3D variables
- [ ] Support serialization in `fuse_rl` kinematic constraint

I'll migrate the graph message to Graphviz DOT converter from https://github.com/locusrobotics/fuse/pull/81 and start looking at an rviz plugin to at least show the pose variables, and relative and absolute constraints with metric information.

PR #91, opened 2019-08-14
Add serialization to all Variable classes

svwilliams:
The PR adds Boost Serialization methods to all Variable classes

### DIFF
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -8,22 +8,21 @@ set(build_depends
 find_package(catkin REQUIRED COMPONENTS
   ${build_depends}
 )
-find_package(Boost REQUIRED)
+find_package(Boost REQUIRED COMPONENTS serialization)
 find_package(Ceres REQUIRED)
 find_package(Eigen3 REQUIRED)
 
 catkin_package(
   INCLUDE_DIRS
     include
-    ${CERES_INCLUDE_DIRS}
-    ${EIGEN3_INCLUDE_DIRS}
   LIBRARIES
     ${PROJECT_NAME}
-    ${CERES_LIBRARIES}
   CATKIN_DEPENDS
     ${build_depends}
   DEPENDS
     Boost
+    CERES
+    EIGEN3
 )
 
 ###########
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -54,6 +54,7 @@ target_include_directories(${PROJECT_NAME}
     ${EIGEN3_INCLUDE_DIRS}
 )
 target_link_libraries(${PROJECT_NAME}
+  ${Boost_LIBRARIES}
   ${catkin_LIBRARIES}
   ${CERES_LIBRARIES}
 )
--- a/fuse_core/include/fuse_core/serialization.h
+++ b/fuse_core/include/fuse_core/serialization.h
@@ -1,0 +1,143 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2019, Locus Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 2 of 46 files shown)


fuse_core/include/fuse_core/serialization.h
```diff
@@ -0,0 +1,143 @@
...
+void serialize(Archive& archive, ros::Time& stamp, const unsigned int /* version */)
+{
+  archive & stamp.sec;
+  archive & stamp.nsec;
+}
+
+/**
+ * @brief Serialize an Eigen Matrix using Boost Serialization
+ */
+template<class Archive,
```
efernandez, line 80: I think the correct way to take an Eigen matrix as a function argument is described here: https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html

It'd also allow to pass Eigen expressions, not just a matrix, which I think it should be fine, and maybe actually useful in some cases.

There are two options:
1. Template function
2. `Eigen::Ref<>`

I think in this case we need the template version, which would look like this:
``` c++
template <class Archive, typename Derived>
void save(Archive& archive, const EIen::MatrixBase<Derived>& matrix, const unsigned int /* version */)
```

The same applies below for the `load` function.

However, if you're planning to have specializations for the 4 cases of `MaxRows_` and `MaxCols_` being fixed size or `Eigen::Dynamic`, then it's probably better to leave this as is now. What you have know works for all cases, but for a fixed size matrix you don't need to serialize the `rows` and `cols`. Calling `resize` on a fixed-size matrix, on the other size, would have no effect or throw an assert, which sounds good to me: https://eigen.tuxfamily.org/dox/group__TutorialMatrixClass.html#title6

fuse_core/CMakeLists.txt
```diff
@@ -8,22 +8,21 @@ set(build_depends
...
-    ${EIGEN3_INCLUDE_DIRS}
   LIBRARIES
     ${PROJECT_NAME}
-    ${CERES_LIBRARIES}
   CATKIN_DEPENDS
     ${build_depends}
   DEPENDS
     Boost
+    CERES
+    EIGEN3
```
efernandez, line 25: Eigen is a header only library. I don't think there's any `EIGEN3_LIBRARIES` variable defined: https://bitbucket.org/eigen/eigen/src/default/cmake/Eigen3Config.cmake.in

fuse_variables/include/fuse_variables/acceleration_angular_3d_stamped.h
```diff
@@ -114,8 +124,27 @@ class AccelerationAngular3DStamped : public FixedSizeVariable<3>, public Stamped
...
+  /**
+   * @brief The Boost Serialize method that serializes all of the data members in to/out of the archive
+   *
+   * @param[in/out] archive - The archive object that holds the serialized class members
+   * @param[in] version - The version of the archive being read/written. Generally unused.
+   */
+  template<class Archive>
+  void serialize(Archive& archive, const unsigned int /* version */)
+  {
+    archive & boost::serialization::base_object<fuse_variables::FixedSizeVariable<3>>(*this);
```
efernandez: Maybe use `SIZE`, there's no chance to make this wrong, although I guess it'd fail at compile time either way: `archive & boost::serialization::base_object<fuse_variables::FixedSizeVariable<SIZE>>(*this);`

efernandez: Another option would be to explicitly call the base class `serialize` method:
``` c++
FixedSizeVariable<SIZE>::serialize(archive, version);
Stamped::serialize(archive, version);
```
which is shorter.

BTW I don't think you need `fuse_variables::` here.

efernandez (approved):

fuse_core/include/fuse_core/serialization.h
```diff
@@ -0,0 +1,143 @@
...
+#include <fuse_core/uuid.h>
+
+#include <ros/time.h>
+
+#include <boost/archive/binary_iarchive.hpp>
+#include <boost/archive/binary_oarchive.hpp>
+#include <boost/archive/text_oarchive.hpp>
+#include <boost/archive/text_iarchive.hpp>
+#include <boost/serialization/access.hpp>
+#include <boost/serialization/array_wrapper.hpp>
```
efernandez: `#include <boost/serialization/array.hpp>` , so it compiles with Boost serialization `1.58.0`.

See https://github.com/locusrobotics/fuse/pull/83#issuecomment-521663781 for more details.

fuse_core/include/fuse_core/serialization.h
svwilliams: Yeah. This was one of those things I pulled from SO. I'm was a bit curious about the resizing thing for static-sized arrays, but it seems to work. It is something I plan to revisit at some future date, as much for curiosity as anything.

That said, I'm not sure that accepting a matrix by base type is the correct thing to do in this case. Since we are serializing, we need the full derived type. But the CRTP of MatrixBase may change things. I haven't thought it through.

There is an interesting suggestion here to use "Eigen's plugin feature"...which I didn't know existed.
https://stackoverflow.com/questions/18382457/eigen-and-boostserialize

Anyway, what is here I think is "good enough" for a first implementation. But if you want to make me a PR with better Eigen serialization support, that would be welcome.

fuse_core/CMakeLists.txt
svwilliams: True. But the `DEPENDS` seems sufficiently flexible to handle one without the other:
https://github.com/ros/catkin/blob/kinetic-devel/cmake/catkin_package.cmake#L168

I did a quick test to see if I was crazy:
```
find_package(Eigen3 REQUIRED)
set(depend_name "EIGEN3")
set(STEVE_INCLUDE_DIRS "")
set(STEVE_LIBRARIES "")
list(APPEND STEVE_INCLUDE_DIRS ${${depend_name}_INCLUDE_DIRS})
list(APPEND STEVE_LIBRARIES ${${depend_name}_LIBRARIES})
message("STEVE_INCLUDE_DIRS: ${STEVE_INCLUDE_DIRS}")
message("STEVE_LIBRARIES: ${STEVE_LIBRARIES}")
```
And I get:
```
STEVE_INCLUDE_DIRS: /usr/include/eigen3
STEVE_LIBRARIES:
```
Which seems like the desired outcome.

fuse_variables/include/fuse_variables/acceleration_angular_3d_stamped.h
svwilliams: Due to the nature of the serialization system, calling the base class `serialize()` function is not an option. The serializer need to know about the class hierarchy, which gets registered when you use `boost::serialization::base_object`. But using `N`/`SIZE` is definitely a good call.

fuse_core/include/fuse_core/serialization.h
efernandez: I completely agree this can be explored later.

fuse_variables/include/fuse_variables/acceleration_angular_3d_stamped.h
efernandez: Thanks for the clarification on calling the base class `serialize` method.

ayrton04 (approved):

PR #92, opened 2019-08-15
Added serialization to all the constraint classes

svwilliams:
Added serialization to all the constraint classes

### DIFF
--- a/fuse_core/include/fuse_core/local_parameterization.h
+++ b/fuse_core/include/fuse_core/local_parameterization.h
@@ -35,7 +35,9 @@
 #define FUSE_CORE_LOCAL_PARAMETERIZATION_H
 
 #include <fuse_core/macros.h>
+#include <fuse_core/serialization.h>
 
+#include <boost/serialization/access.hpp>
 #include <ceres/local_parameterization.h>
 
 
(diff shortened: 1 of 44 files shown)


fuse_core/include/fuse_core/local_parameterization.h
```diff
@@ -86,6 +88,21 @@ class LocalParameterization : public ceres::LocalParameterization
...
+  friend class boost::serialization::access;
+
+  /**
+   * @brief The Boost Serialize method that serializes all of the data members in to/out of the archive
+   *
+   * @param[in/out] archive - The archive object that holds the serialized class members
+   * @param[in] version - The version of the archive being read/written. Generally unused.
+   */
+  template<class Archive>
+  void serialize(Archive& /* archive */, const unsigned int /* version */)
```
efernandez, line 103: Why do we need this if there's no attributes to de/serialize?

efernandez (approved):

fuse_core/include/fuse_core/local_parameterization.h
svwilliams: If we serialize an object by base pointer (which is done in the MarginalConstraint class), the base class needs to have a serialize() method. There may be a Boost macro or switch somewhere that makes this less invasive, but I haven't found it.

efernandez: Thanks for clarifying. I was expecting that when I saw the children classes of this one. I'm not familiar enough with Boost serialization, so I don't know if there's a shorter way to do this. Either way, it LGTM.

ayrton04 (approved):

PR #93, opened 2019-08-15
Added serialization to the graph classes

svwilliams:
Added serialization to the graph classes

### DIFF
--- a/fuse_core/include/fuse_core/graph.h
+++ b/fuse_core/include/fuse_core/graph.h
@@ -36,17 +36,100 @@
 
 #include <fuse_core/constraint.h>
 #include <fuse_core/macros.h>
+#include <fuse_core/serialization.h>
 #include <fuse_core/transaction.h>
 #include <fuse_core/uuid.h>
 #include <fuse_core/variable.h>
 
+#include <boost/core/demangle.hpp>
 #include <boost/range/any_range.hpp>
+#include <boost/serialization/access.hpp>
+#include <boost/type_index/stl_type_index.hpp>
 #include <ceres/covariance.h>
 #include <ceres/solver.h>
 
 #include <ostream>
+#include <string>
 #include <utility>
 #include <vector>
+
+
+/**
+ * @brief Implementation of the serialize() and deserialize() member functions for derived classes
+ *
+ * Usage:
+ * @code{.cpp}
+ * class Derived : public Graph
+ * {
+ * public:
+ *   FUSE_GRAPH_SERIALIZE_DEFINITION(Derived);
+ *   // The rest of the derived graph implementation
+ * }
+ * @endcode
+ */
+#define FUSE_GRAPH_SERIALIZE_DEFINITION(...) \
+  void serialize(fuse_core::BinaryOutputArchive& archive) const override \
+  { \
+    archive << *this; \
+  }  /* NOLINT */ \
+  void serialize(fuse_core::TextOutputArchive& archive) const override \
+  { \
+    archive << *this; \
+  }  /* NOLINT */ \
+  void deserialize(fuse_core::BinaryInputArchive& archive) override \
+  { \
+    archive >> *this; \
+  }  /* NOLINT */ \
+  void deserialize(fuse_core::TextInputArchive& archive) override \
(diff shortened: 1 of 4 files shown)


fuse_core/include/fuse_core/graph.h
```diff
@@ -36,19 +36,102 @@
...
+  { \
+    return detail::type(); \
+  }
+
+/**
+* @brief Convenience function that creates the required pointer aliases, and type() method
+*
+* Usage:
+* @code{.cpp}
+* class Derived : public Constraint
```
efernandez: `class Derived : public Graph`

fuse_core/include/fuse_core/graph.h
```diff
@@ -36,19 +36,102 @@
...
+ * @brief Implements the type() member function using the suggested implementation
+ *
+ * Also creates a static detail::type() function that may be used without an object instance
+ *
+ * Usage:
+ * @code{.cpp}
+ * class Derived : public Graph
+ * {
+ * public:
+ *   FUSE_CONSTRAINT_TYPE_DEFINITION(Derived);
```
efernandez: `FUSE_GRAPH_TYPE_DEFINITION(Derived);`

fuse_core/include/fuse_core/graph.h
```diff
@@ -36,19 +36,102 @@
...
+
+/**
+* @brief Convenience function that creates the required pointer aliases, and type() method
+*
+* Usage:
+* @code{.cpp}
+* class Derived : public Constraint
+* {
+* public:
+*   FUSE_CONSTRAINT_DEFINITIONS(Derived);
```
efernandez: `FUSE_GRAPH_DEFINITIONS(Derived);`

fuse_core/include/fuse_core/graph.h
```diff
@@ -36,19 +36,102 @@
...
+*   // The rest of the derived constraint implementation
```
efernandez: `// The rest of the derived graph implementation`

fuse_core/include/fuse_core/graph.h
```diff
@@ -270,6 +361,73 @@ class Graph
...
+
+  /**
+   * @brief Serialize this Constraint into the provided binary archive
+   *
+   * This can/should be implemented as follows in all derived classes:
+   * @code{.cpp}
+   * archive << *this;
+   * @endcode
+   *
+   * @param[out] archive - The archive to serialize this constraint into
```
efernandez: `constraint` -> `graph` (same below)

fuse_graphs/include/fuse_graphs/hash_graph.h
```diff
@@ -330,8 +337,56 @@ class HashGraph : public fuse_core::Graph
...
+   * @param[in/out] archive - The archive object that holds the serialized class members
+   * @param[in] version - The version of the archive being read/written. Generally unused.
+   */
+  template<class Archive>
+  void serialize(Archive& archive, const unsigned int /* version */)
+  {
+    archive & boost::serialization::base_object<fuse_core::Graph>(*this);
+    archive & constraints_;
+    archive & constraints_by_variable_uuid_;
+    archive & problem_options_;
```
efernandez, line 357: I don't think we have to de/serialize the problem options. :thinking:

efernandez (approved):

fuse_graphs/include/fuse_graphs/hash_graph.h
svwilliams: Agreed that it is probably overkill. But if you want to serialize a graph and later load it back and have it behave the same way, then it seems required. And its only a hand full of values, so it's not that painful.

Really what I'm toying with in my head moving a bunch of the graph functionality out into free functions. The graph would build the Problem object (where you could pass in Problem::Options at call time), then optimizing and computing covariances would be generic functions that operate on the Problem. But I have bigger fish to fry at the moment.

efernandez: Fair enough

ayrton04 (approved):

PR #94, opened 2019-08-15
Added serialization for the Transaction class

svwilliams:
Added serialization for the Transaction class

### DIFF
--- a/fuse_core/test/test_transaction.cpp
+++ b/fuse_core/test/test_transaction.cpp
@@ -52,6 +52,7 @@ using fuse_core::UUID;
  *
  * Order of the stamps is not important. Extra stamps in the Transaction will return False.
  *
+ * @tparam TimeRange   A range or container with objects compatible with a "const ros::Time&" signature
  * @param expected     The set of expected stamps
  * @param transaction  The transaction to test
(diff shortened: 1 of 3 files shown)


fuse_core/test/test_transaction.cpp
```diff
@@ -51,16 +52,16 @@ using fuse_core::UUID;
...
- * @return                 True if the expected stamps, and only the expected stamps, exist in the
- *                         transaction, False otherwise.
+ * @param expected     The set of expected stamps
+ * @param transaction  The transaction to test
+ * @return             True if the expected stamps, and only the expected stamps, exist in the
+ *                     transaction, False otherwise.
  */
-bool testInvolvedStamps(
-  const std::vector<ros::Time>& expected_stamps, const Transaction& transaction)
+template <typename TimeRange>
```
efernandez, line 61: You can also add `@tparam TimeRange` to add a doxygen comment for this template param: http://www.doxygen.nl/manual/commands.html#cmdtparam

I guess you're switching to ranges, which is quite obvious. Not sure if it's worth adding a comment regarding the types expected.

efernandez (approved):

ayrton04 (approved):

PR #96, opened 2019-08-15
Added fuse SerializedXXX messages

svwilliams:
Added fuse SerializedXXX messages. The intent is to serialize a graph or transaction into the message buffer, send it across the ROS Comms wire, and deserialize it on the other end. The down side being the graph/transaction data will look like garbage if you echo it.

### DIFF
--- a/fuse_msgs/package.xml
+++ b/fuse_msgs/package.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>fuse_msgs</name>
+  <version>0.0.0</version>
+  <description>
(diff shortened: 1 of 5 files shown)


fuse_msgs/package.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>fuse_msgs</name>
+  <version>0.0.0</version>
```
efernandez: Remember to sync the version with the rest of the packages, so you don't have issues when you release a new version of the repo.

efernandez (approved):

ayrton04 (approved):

PR #97, opened 2019-08-15
Added deserializers for the Graph and Transaction messages

svwilliams:
Added deserializers for the Graph and Transaction messages, and a few other toys.

Alright. I believe this is the last of the bunch. I ran a quick integration test, and everything _appears_ to be running as expected.

The integration test contained a SerializedPublisher, which provides `/graph` and `/transaction` topics. Echoing the topics provides the expected mostly unhelpful information:
```
$ rostopic echo /graph --noarr
header: 
  seq: 0
  stamp: 
    secs: 4062
    nsecs: 570341615
  frame_id: ''
plugin_name: "fuse_graphs::HashGraph"
data: "<array type: uint8, length: 8031>"
---
header: 
  seq: 1
  stamp: 
    secs: 4062
    nsecs: 620715955
  frame_id: ''
plugin_name: "fuse_graphs::HashGraph"
data: "<array type: uint8, length: 11907>"
```

If you fire up a `fuse_echo` node instead, the graph and transaction are printed in a somewhat human-readable form:
```
$ rosrun fuse_core fuse_echo 
-------------------------
GRAPH:
received at: 4062.570195672
HashGraph
  constraints:
   - fuse_constraints::AbsoluteVelocityLinear2DStampedConstraint
  uuid: 53f42a3a-83b4-4769-9232-a82838ec2f30
  variable: 2b49ac71-6d10-5ff9-9b11-7d0b0c8859af
  mean: -0.000278797  2.84778e-12
  sqrt_info:   121.02 0.179513
       0  31621.8

   - fuse_constraints::AbsolutePose2DStampedConstraint
  uuid: 5aee3f68-3299-4e5c-88ba-5ff566601f26
  position variable: 3541141d-2136-56ec-854d-0209bc5c8628
  orientation variable: c077f205-a644-5afe-96b6-a4ba1c23530f
  mean:     0.182583 -0.000140056 -0.000355377
  sqrt_info:     19.0258   0.0750849 2.33595e-05
          0     52.0892     1.80317
          0           0     5.36031

...
```

I'm not going to pretend everything is perfect. In particular, I'm using pluginlib in a slightly unusual way. That may bite me in the future and require a refactor. But at least for this simple test, everything does seem to run as expected.

### DIFF
--- a/fuse_core/include/fuse_core/graph_deserializer.h
+++ b/fuse_core/include/fuse_core/graph_deserializer.h
@@ -75,18 +75,18 @@ public:
    * If no plugin is available for a contained Variable or Constraint, or an error occurs during deserialization,
    * an exception is thrown.
    *
-   * @param[IN]  msg  The SerializedGraph message to be deserialized
+   * @param[in]  msg  The SerializedGraph message to be deserialized
    * @return          A unique_ptr to a derived Graph object
    */
   fuse_core::Graph::UniquePtr deserialize(const fuse_msgs::SerializedGraph::ConstPtr& msg);
 
   /**
-   * @brief Deserialize a graph
+   * @brief Deserialize a SerializedGraph message into a fuse Graph object.
    *
    * If no plugin is available for a contained Variable or Constraint, or an error occurs during deserialization,
    * an exception is thrown.
    *
-   * @param[IN]  msg  The SerializedGraph message to be deserialized
+   * @param[in]  msg  The SerializedGraph message to be deserialized
    * @return          A unique_ptr to a derived Graph object
    */
   fuse_core::Graph::UniquePtr deserialize(const fuse_msgs::SerializedGraph& msg);
--- a/fuse_core/include/fuse_core/graph_deserializer.h
+++ b/fuse_core/include/fuse_core/graph_deserializer.h
@@ -1,0 +1,102 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2019, Locus Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_CORE_GRAPH_DESERIALIZER_H
+#define FUSE_CORE_GRAPH_DESERIALIZER_H
+
+#include <fuse_msgs/SerializedGraph.h>
+#include <fuse_core/constraint.h>
+#include <fuse_core/graph.h>
+#include <fuse_core/variable.h>
+#include <pluginlib/class_loader.h>
+
+
+namespace fuse_core
+{
+
+/**
+ * @brief Serialize a graph into a message
+ */
+void serializeGraph(const fuse_core::Graph& graph, fuse_msgs::SerializedGraph& msg);
+
+/**
+ * @brief Deserialize a graph
+ *
+ * The deserializer object loads all of the known Variable and Constraint libraries, allowing derived types contained
+ * within the graph to be properly deserialized. The libraries will be unloaded on destruction. As a consequence, the
+ * deserializer object must outlive any created graph instances.
+ */
+class GraphDeserializer
+{
+public:
+  /**
+   * @brief Constructor
+   */
+  GraphDeserializer();
+
+  /**
+   * @brief Destructor
+   */
+  ~GraphDeserializer();
+
+  /**
+   * @brief Deserialize a SerializedGraph message into a fuse Graph object.
+   *
+   * If no plugin is available for a contained Variable or Constraint, or an error occurs during deserialization,
+   * an exception is thrown.
+   *
+   * @param[IN]  msg  The SerializedGraph message to be deserialized
+   * @return          A unique_ptr to a derived Graph object
+   */
+  fuse_core::Graph::UniquePtr deserialize(const fuse_msgs::SerializedGraph::ConstPtr& msg);
+
+  /**
+   * @brief Deserialize a graph
+   *
+   * If no plugin is available for a contained Variable or Constraint, or an error occurs during deserialization,
+   * an exception is thrown.
+   *
+   * @param[IN]  msg  The SerializedGraph message to be deserialized
+   * @return          A unique_ptr to a derived Graph object
+   */
+  fuse_core::Graph::UniquePtr deserialize(const fuse_msgs::SerializedGraph& msg);
+
+private:
+  pluginlib::ClassLoader<fuse_core::Constraint> constraint_loader_;  //!< Pluginlib class loader for Constraint types
+  pluginlib::ClassLoader<fuse_core::Graph> graph_loader_;  //!< Pluginlib class loader for Graph types
+  pluginlib::ClassLoader<fuse_core::Variable> variable_loader_;  //!< Pluginlib class loader for Variable types
+};
+
+}  // namespace fuse_core
+
+#endif  // FUSE_CORE_GRAPH_DESERIALIZER_H
--- a/fuse_core/include/fuse_core/graph_deserializer.h
+++ b/fuse_core/include/fuse_core/graph_deserializer.h
@@ -65,11 +65,6 @@ public:
   GraphDeserializer();
 
   /**
-   * @brief Destructor
-   */
-  ~GraphDeserializer();
-
-  /**
    * @brief Deserialize a SerializedGraph message into a fuse Graph object.
    *
    * If no plugin is available for a contained Variable or Constraint, or an error occurs during deserialization,
@@ -92,9 +87,9 @@ public:
   fuse_core::Graph::UniquePtr deserialize(const fuse_msgs::SerializedGraph& msg);
 
 private:
+  pluginlib::ClassLoader<fuse_core::Variable> variable_loader_;  //!< Pluginlib class loader for Variable types
   pluginlib::ClassLoader<fuse_core::Constraint> constraint_loader_;  //!< Pluginlib class loader for Constraint types
   pluginlib::ClassLoader<fuse_core::Graph> graph_loader_;  //!< Pluginlib class loader for Graph types
-  pluginlib::ClassLoader<fuse_core::Variable> variable_loader_;  //!< Pluginlib class loader for Variable types
 };
 
 }  // namespace fuse_core
--- a/fuse_core/include/fuse_core/transaction_deserializer.h
+++ b/fuse_core/include/fuse_core/transaction_deserializer.h
@@ -1,0 +1,101 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2019, Locus Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_CORE_TRANSACTION_DESERIALIZER_H
+#define FUSE_CORE_TRANSACTION_DESERIALIZER_H
+
+#include <fuse_msgs/SerializedTransaction.h>
+#include <fuse_core/constraint.h>
+#include <fuse_core/transaction.h>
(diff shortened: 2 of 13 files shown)


fuse_core/src/fuse_echo.cpp
```diff
@@ -0,0 +1,94 @@
...
+ * Class that subscribes to the 'graph' and 'transaction' topic and prints the objects to stdout
+ */
+class FuseEcho
+{
+public:
+  explicit FuseEcho(const ros::NodeHandle& node_handle = ros::NodeHandle()) :
+    node_handle_(node_handle)
+  {
+    // Subscribe to the constraint topic
+    graph_subscriber_ = node_handle_.subscribe("graph", 1, &FuseEcho::graphCallback, this);
```
efernandez: I think the queue size should be larger than `1`. IMHO an `echo` tool should have a large queue, so no message is missed. Another option would be to add a ROS param `queue_size`, but sounds like too much.

fuse_core/src/fuse_echo.cpp
```diff
@@ -0,0 +1,94 @@
...
+    auto transaction = transaction_deserializer_.deserialize(msg);
+    transaction.print();
+  }
+};
+
+int main(int argc, char **argv)
+{
+  ros::init(argc, argv, "fuse_echo", ros::init_options::AnonymousName);
+
+  // Object that subscribes to the 'transaction' topic and prints the constraints and variables to stdout
```
efernandez: `// Object that subscribes to the 'graph' and 'transaction' topic and prints the constraints and variables to stdout`

fuse_core/src/graph_deserializer.cpp
```diff
@@ -0,0 +1,118 @@
...
+  // Create a Graph object using pluginlib. This will throw if the plugin name is not found.
+  // The unique ptr returned by pluginlib has a custom deleter. This makes it annoying to return
+  // back to the user as the output is not equivalent to fuse_core::Graph::UniquePtr. Instead, wrap an
+  // unmanaged raw pointer in a unique_ptr, and handle the library unloading in the destructor.
+  auto graph = fuse_core::Graph::UniquePtr(graph_loader_.createUnmanagedInstance(msg.plugin_name));
+  // Deserialize the message into the Graph. This will throw if something goes wrong during deserialization.
+  std::stringstream stream;
+  std::copy(msg.data.begin(), msg.data.end(), std::ostream_iterator<char>(stream));
+  {
+    BinaryInputArchive archive(stream);
```
efernandez, line 92: Sorry, but I think I missed whether there's support for different input archive types or not. Could the stream be text instead of binary? I'm just wondering. I think this is fine, but if the graph could be in text format in the `stream`, then this will fail.

fuse_core/include/fuse_core/graph_deserializer.h
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * @brief Deserialize a SerializedGraph message into a fuse Graph object.
+   *
+   * If no plugin is available for a contained Variable or Constraint, or an error occurs during deserialization,
+   * an exception is thrown.
+   *
+   * @param[IN]  msg  The SerializedGraph message to be deserialized
+   * @return          A unique_ptr to a derived Graph object
+   */
+  fuse_core::Graph::UniquePtr deserialize(const fuse_msgs::SerializedGraph::ConstPtr& msg);
```
efernandez, line 76: I think this could be `static`: `static fuse_core::Graph::UniquePtr deserialize(const fuse_msgs::SerializedGraph::ConstPtr& msg);` (same below)

fuse_core/include/fuse_core/transaction_deserializer.h
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * @brief Deserialize a SerializedTransaction message into a fuse Transaction object.
+   *
+   * If no plugin is available for a contained Variable or Constraint, or an error occurs during deserialization,
+   * an exception is thrown.
+   *
+   * @param[IN]  msg  The SerializedTransaction message to be deserialized
+   * @return          A fuse Transaction object
+   */
+  fuse_core::Transaction deserialize(const fuse_msgs::SerializedTransaction::ConstPtr& msg);
```
efernandez, line 76: I think this could be `static`: `fuse_core::Transaction deserialize(const fuse_msgs::SerializedTransaction::ConstPtr& msg);` (same below)

fuse_core/src/graph_deserializer.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+
+namespace fuse_core
+{
+
+void serializeGraph(const fuse_core::Graph& graph, fuse_msgs::SerializedGraph& msg)
+{
+  std::stringstream stream;
+  // Serialize graph into the stream
+  // The archive to not flushed until it goes out of scope
```
efernandez: `// Scope the archive to not flush until it goes out of scope`

fuse_core/src/transaction_deserializer.cpp
```diff
@@ -0,0 +1,109 @@
...
+  return deserialize(*msg);
+}
+
+fuse_core::Transaction TransactionDeserializer::deserialize(const fuse_msgs::SerializedTransaction& msg)
+{
+  // The Transaction object is not a plugin and has no derived types. That makes it much easier to use.
+  auto transaction = fuse_core::Transaction();
+  // Deserialize the message into the Variable. This will throw if something goes wrong in the deserialization.
+  std::stringstream stream;
+  std::copy(msg.data.begin(), msg.data.end(), std::ostream_iterator<char>(stream));
```
efernandez: I think you could using Boost iostreams to make this more efficient: https://stackoverflow.com/a/12646850 (same for the graph)

fuse_publishers/src/serialized_publisher.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+SerializedPublisher::SerializedPublisher() :
+  fuse_core::AsyncPublisher(1)
+{
+}
+
+void SerializedPublisher::onInit()
+{
+  // Advertise the topics
+  graph_publisher_ = node_handle_.advertise<fuse_msgs::SerializedGraph>("graph", 1);
```
efernandez: IMHO the `queue_size` should be a ROS param, because the user might want a larger value than the default. Personally, I'd set the default queue size to `10` (or more). I think it's better that the default behaviour favors publishing all messages, i.e. no messages are dropped if we publish too fast here.

efernandez: IMO the publishers should be in the private namespace. IIR you can simply replace `node_handle_` with `private_node_handle_`.

efernandez (approved):

efernandez:
@ayrton04 I wonder if the `SerializedPublisher` should also have a `device_name` (and `device_id`), as the `Odometry2DPublisher` :thinking:

fuse_core/src/graph_deserializer.cpp
svwilliams: There is serialization support for both the text and binary archive types. If you are using serialization in your own application, you can select either archive type.

However, the functions that serializes data into the message explicitly use the binary format, so the deserialize from a message function also uses the binary archive. I have not seen a method for detecting the archive type at runtime, but it _may_ be possible.

We can also discuss expanding the deserializer to fulfill other use-cases. I just don't know what those use-cases are yet. The SerializedPublisher-FuseEcho pair is really just a toy to demonstrate that the serialization system is functioning correctly.

fuse_core/include/fuse_core/graph_deserializer.h
svwilliams: I need to test, but I don't _think_ that would work. A static method would allow me to call `deserialize()` without first creating an object. I believe I need to load the shared libraries that contain the variables and constraints before deserializing the graph. At least, that how it works in my head.

fuse_core/src/transaction_deserializer.cpp
svwilliams: Boost has everything.

I did a quick attempt, but I ran into type issues. The archive only accepts streams of type `char`. However the serialized message is defined as `uint8`/`unsigned char`. And fascinatingly, switching the message to `int8`/`signed char` also doesn't work because a `char` is not the same type as a `signed char`. I could switch the message buffer to `string`, but then I pretty sure it would print even if you use `rostopic echo --noarr`. I'll get this working eventually, just not today.

fuse_core/include/fuse_core/graph_deserializer.h
efernandez: Yes, that's true, but there's an issue right now. You should have a singleton to load the shared libraries only once, and more importantly, unload them only once.

Currently, `fuse_echo` crashes on shutdown because of that, with this message:
``` bash
terminate called after throwing an instance of 'class_loader::LibraryUnloadException'
  what():  Attempt to unload library that class_loader is unaware of.
```

I commented out the transaction deserializer code in `fuse_echo` and it didn't crash, so I'm quite sure the issue is that when the graph deserializer (call after the transaction deserializer because it's before in the attributes list) is destroyed it unloads the shared libraries that has already been unloaded by the transaction desrializer. Indeed, the core dump I get shows the crash coming from the graph deserializer destructor, on `pluginlib` unloading methods.

fuse_core/include/fuse_core/graph_deserializer.h
```diff
@@ -0,0 +1,102 @@
...
+void serializeGraph(const fuse_core::Graph& graph, fuse_msgs::SerializedGraph& msg);
+
+/**
+ * @brief Deserialize a graph
+ *
+ * The deserializer object loads all of the known Variable and Constraint libraries, allowing derived types contained
+ * within the graph to be properly deserialized. The libraries will be unloaded on destruction. As a consequence, the
+ * deserializer object must outlive any created graph instances.
+ */
+class GraphDeserializer
```
ayrton04, line 59: Would it make sense to mark this class as `final`?

fuse_core/src/graph_deserializer.cpp
```diff
@@ -0,0 +1,118 @@
...
+  return deserialize(*msg);
+}
+
+fuse_core::Graph::UniquePtr GraphDeserializer::deserialize(const fuse_msgs::SerializedGraph& msg)
+{
+  // Create a Graph object using pluginlib. This will throw if the plugin name is not found.
+  // The unique ptr returned by pluginlib has a custom deleter. This makes it annoying to return
+  // back to the user as the output is not equivalent to fuse_core::Graph::UniquePtr. Instead, wrap an
+  // unmanaged raw pointer in a unique_ptr, and handle the library unloading in the destructor.
+  auto graph = fuse_core::Graph::UniquePtr(graph_loader_.createUnmanagedInstance(msg.plugin_name));
```
ayrton04, line 87: I'm guessing this call prevents this method from being marked `const`?

ayrton04 (approved):
Pending Enrique's issues.

ayrton04:
@efernandez We can do that. What would be the advantage?

efernandez:
@ayrton04 Both the `GraphDeserializer` and `TransactionDeserializer` destructors unload the plugins. That means some plugins are unloaded twice, which seems to be the root cause of the crash on shutdown.

One solution could be a singleton, so both the `GraphDeserializer` and `TransactionDeserializer` create a static instance of it in their constructors. And when the `static` singleton gets out of scope on shutdown, its destructor would unload all plugins only once.

I can play with that and provide a PR on top of this one if you want.

ayrton04:
Sorry, being dense: you're suggesting that's a solution to this?

> @ayrton04 I wonder if the SerializedPublisher should also have a device_name (and device_id), as the Odometry2DPublisher

svwilliams:
Working on the deserializer double-unload issue today. Hopefully I can come up with something clever.

svwilliams:
So....Deserializers....This is a bit weird. I'm not sure exactly what to do about it.

The pluginlib system already contains a singleton-style loader here: https://github.com/ros/class_loader/blob/melodic-devel/src/class_loader_core.cpp

Each time a library is loaded, it adds an owner for each class in that library. When the last owner is removed, the library is unloaded. This is exactly what we want to happen.

The issue arises with the fuse library structure. There is a `libfuse_variables.so` library and a `libfuse_constraints.so` library. However, because the constraints need to know about the variables, 
`libfuse_constraints.so` is linked against `libfuse_variables.so`. When the constraints library is loaded, the class loader also finds the variables.

Some debug printing I added:
```
[ INFO] [/fuse_echo_1566351260173492716]: GraphDeserializer::GraphDeserializer()  Loading library for fuse_constraints::AbsoluteAccelerationAngular2DStampedConstraint
class_loader::ClassLoader::ClassLoader()  Creating ClassLoader, library = /home/swilliams/locus_dev/devel/lib//libfuse_constraints.so, address = 0x558d6d44c3a0
ClassLoaderCore::loadLibrary() /home/swilliams/locus_dev/devel/lib//libfuse_constraints.so : 0x558d6d44c3a0
ClassLoaderCore::loadLibrary()   isLibraryLoadedByAnybody? 0
  - fuse_constraints::AbsoluteAccelerationAngular2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::AbsoluteAccelerationLinear2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::AbsoluteOrientation2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::AbsoluteOrientation3DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::AbsoluteOrientation3DStampedEulerConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::AbsolutePose2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::AbsolutePose3DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::AbsolutePosition2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::AbsolutePosition3DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::AbsoluteVelocityAngular2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::AbsoluteVelocityLinear2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::MarginalConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::RelativeAccelerationAngular2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::RelativeAccelerationLinear2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::RelativeOrientation2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::RelativeOrientation3DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::RelativePose2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::RelativePose3DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::RelativePosition2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::RelativePosition3DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::RelativeVelocityAngular2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_constraints::RelativeVelocityLinear2DStampedConstraint
    * 0x558d6d44c3a0
  - fuse_variables::AccelerationAngular2DStamped
    * 0x558d6d44c3a0
  - fuse_variables::AccelerationAngular3DStamped
    * 0x558d6d44c3a0
  - fuse_variables::AccelerationLinear2DStamped
    * 0x558d6d44c3a0
  - fuse_variables::AccelerationLinear3DStamped
    * 0x558d6d44c3a0
  - fuse_variables::Orientation2DStamped
    * 0x558d6d44c3a0
  - fuse_variables::Orientation3DStamped
    * 0x558d6d44c3a0
  - fuse_variables::Position2DStamped
    * 0x558d6d44c3a0
  - fuse_variables::Position3DStamped
    * 0x558d6d44c3a0
  - fuse_variables::VelocityAngular2DStamped
    * 0x558d6d44c3a0
  - fuse_variables::VelocityAngular3DStamped
    * 0x558d6d44c3a0
  - fuse_variables::VelocityLinear2DStamped
    * 0x558d6d44c3a0
  - fuse_variables::VelocityLinear3DStamped
    * 0x558d6d44c3a0
ClassLoaderCore::loadLibrary() end
class_loader::ClassLoader::ClassLoader()  Finished creating ClassLoader, library = /home/swilliams/locus_dev/devel/lib//libfuse_constraints.so, address = 0x558d6d44c3a0
```
And because all the variables are loaded already, when we load the `libfuse_variables.so` library, there are no new classes associated with it.
```
[ INFO] [/fuse_echo_1566351260173492716]: GraphDeserializer::GraphDeserializer()  Loading library for fuse_variables::AccelerationAngular2DStamped
class_loader::ClassLoader::ClassLoader()  Creating ClassLoader, library = /home/swilliams/locus_dev/devel/lib//libfuse_variables.so, address = 0x558d6d490fc0
ClassLoaderCore::loadLibrary() /home/swilliams/locus_dev/devel/lib//libfuse_variables.so : 0x558d6d490fc0
ClassLoaderCore::loadLibrary()   isLibraryLoadedByAnybody? 0
ClassLoaderCore::loadLibrary() end
class_loader::ClassLoader::ClassLoader()  Finished creating ClassLoader, library = /home/swilliams/locus_dev/devel/lib//libfuse_variables.so, address = 0x558d6d490fc0
```
And because there are no classes associated with the `libfuse_variables.so` library, the loading and unloading is no tracked properly.

I just pushed a commit that changes the order of library loading in the deserializers. By loading the variables first and then the constraints, there are classes associated with the `libfuse_variables.so` library. This allows the library usage tracking to work as expected, and class destruction completes without issue.

I'm not in love with this as the final solution. If some object creates a Constraints plugin loader before it creates a deserializer, we are right back in this same place. I feel like the correct solution is to modify the library compilation is some way, possibly with some forward-declares or such, so the `libfuse_constraints.so` does not contain the variable classes. But I'm getting to the edge of what I understand about compilation and linking. If someone has some suggestions on a remedy, I'll gladly try them out. Otherwise, I may make a ticket for this issue and look into it later.

fuse_core/include/fuse_core/graph_deserializer.h
efernandez: @ayrton04 I was proposing a solution for this, but from the last comments from @svwilliams it seems that the issue was bigger than I initially thought.

ayrton04: Gotcha. My first comment to you was about this comment you made originally:

> @ayrton04 I wonder if the SerializedPublisher should also have a device_name (and device_id), as the Odometry2DPublisher 🤔

What would be the advantage of that?

efernandez:
A little late but I want to give my :+1: to this.

FYI I've been busy debugging some issues related with fuse at user level:

1. IMU sending zero covariance : this caused the problem to have `inf` in an angular velocity constraint, but printing the graph (still running w/o serialization) didn't help me finding the sensor model that produced it because that information is not in the contraints. I wonder if it's worth adding that information, at least to the `print` method. I'm mentioning it here because for that it'd have to be a field of the class and be serialized as well, but I'm not sure that'd be a good idea. Thoughts?

2. Failed to compute the covariance for some blocks : after setting a pose the covariance couldn't be computed. This is the first time it happens to me so far, so it's weird. I saw this on a bag file and I can't reproduce it. With the graph serialization I think I could publish the graph and bag it, so I can see how it looks like. I wonder what's your plan for loading the graph?, as that would help to bring the system to the conditions prior to an error, making things super easy to triage.

svwilliams:
1. You are thinking of a "source" or similar string field to be populated with the sensor model's `name()`? I like it.
2. I haven't really thought through how the graph saving/loading should happen in practice. Maybe  an optional parameter for the optimizer with a file path? It would deserialize the graph before starting everything else up? That is more of a "continue processing" mechanism than a debug mechanism. For debugging, maybe we need a `fuse_wtf` command that loads up the graph and looks for problems -- disconnected graphs, bad covariances, prints the detailed optimization summary, etc.

efernandez:
1. Yes, sth like a `source` field populated with the sensor/motion model's `name()`.

2. I need to think a little more about this. I like the idea of a `fuse_wtf` command. Personally, I'd avoid dumping things into a file if we could just rely on the graph being published, which in principle should be enough, since it has all the information. I'm just a little concerned of the graph not being published for all changes that happen to it, e.g. on `reset` when a `set_pose` request happens (not sure).

PR #98, opened 2019-08-22
[RST-2340] Add serialization support to fuse

svwilliams:
This PR adds serialization to the API of the Variable, Constraint, Graph, and Transaction classes. Support for additional class types can be added as needed.

Additionally, a ROS msgs have been added to transport serialized Graph and Transaction objects using standard ROS comms mechanisms. To demonstrate this ability, a fuse Publisher was created to serialize and publish the updated graph, and a stand alone utility, fuse_echo, was created to subscribe, deserialize, and print the graph.

All commits in this PR have been reviewed independently.

### DIFF
--- a/fuse_publishers/src/serialized_publisher.cpp
+++ b/fuse_publishers/src/serialized_publisher.cpp
@@ -1,0 +1,87 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2019, Locus Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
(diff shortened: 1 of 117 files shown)


fuse_publishers/src/serialized_publisher.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+void SerializedPublisher::notifyCallback(
+  fuse_core::Transaction::ConstSharedPtr transaction,
+  fuse_core::Graph::ConstSharedPtr graph)
+{
+  auto stamp = ros::Time::now();
+  if (graph_publisher_.getNumSubscribers() > 0)
+  {
+    fuse_msgs::SerializedGraph msg;
+    msg.header.stamp = stamp;
```
efernandez, line 77: Could you also set `msg.header.frame_id = "map";`? (same below for the transaction)

I don't know if it's possible to retrieve the frame of the graph / transaction somehow from their classes. If they don't have that field, I think for now this should be fine, or a ROS param to set the frame.

This is important to visualize the graph in `rviz`, for example.

efernandez (changes requested):

efernandez (approved):

svwilliams:
All commits in this PR have been reviewed independently. Merging.

PR #99, opened 2019-08-23
Add fuse_viz pkg with rviz SerializedGraph display

efernandez:
This adds a `fuse_viz` pkg with and `rviz` `fuse_msgs::SerializedGraph` display. This is a very basic implementation for now, but it sets a baseline for improvements and discussions.

This PR is now targeting `devel` and it's been rebased.

### DIFF
--- a/fuse/package.xml
+++ b/fuse/package.xml
@@ -19,6 +19,7 @@
   <exec_depend>fuse_optimizers</exec_depend>
   <exec_depend>fuse_publishers</exec_depend>
   <exec_depend>fuse_variables</exec_depend>
+  <exec_depend>fuse_viz</exec_depend>
 
   <export>
     <metapackage />
--- a/fuse_viz/CMakeLists.txt
+++ b/fuse_viz/CMakeLists.txt
@@ -1,0 +1,89 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(fuse_viz)
+
+set(build_depends
+  fuse_msgs
+  fuse_variables
+  rviz
+)
+
+find_package(catkin REQUIRED COMPONENTS
+  ${build_depends}
+)
+
+find_package(Qt5 COMPONENTS Core Widgets REQUIRED)
+set(QT_LIBRARIES Qt5::Widgets)
+
+add_definitions(-DQT_NO_KEYWORDS)
+
+###########
+## Build ##
+###########
+
+add_compile_options(-std=c++14 -Wall -Werror)
+
+catkin_package(
+  INCLUDE_DIRS include
+  LIBRARIES ${PROJECT_NAME}
+  CATKIN_DEPENDS
+    ${build_depends}
+  DEPENDS
+    QT
+)
+
+qt5_wrap_cpp(moc_files
+  include/fuse_viz/serialized_graph_display.h
+)
+
+set(source_files
+  src/serialized_graph_display.cpp
+  ${moc_files}
+)
+
+add_library(${PROJECT_NAME}
+  ${source_files}
+)
+add_dependencies(${PROJECT_NAME}
+  ${catkin_EXPORTED_TARGETS}
+)
+target_include_directories(${PROJECT_NAME}
+  PUBLIC
+    include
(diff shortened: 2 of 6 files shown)


svwilliams (approved):
Some thoughts on the Graph display:
* I assume you have future plans for the graph visualization. At the moment it just displays the robot poses spheres/axes. This limited result could have been accomplished with a fuse::Publisher that outputs a geometry_msgs::PoseArray or a visualizations::MarkerArray.

And on the DOT converter:
* Since you have created a static graph deserializer, I'm not sure wrapping the DotWriter in a class is required. It could just be a free function. Maybe.
* If you separate out the deserialization of the graph from the conversion to dot format, you could make one `write()` signature accept a `const Graph&`. Then a second signature could accept a `SerializedGraphConstPtr`. That second version would deserialize the graph then call the first. This makes the dot conversion a little more generic and potentially usable in other places.
* In addition to the stand-alone `SerializedGraphDOTWriterNode`, you could create a `DOTPublisher` that plugs in directly to the fuse framework. This would avoid the serialization-deserialization step, but would require the converter to be running constantly. So they would have slightly different use-cases.

efernandez:
Yes, I plan to add edges for differential mode, which should connect the variables I'm displaying right now.

Will address your comments for the dot format converter when I'm back from vacation on Friday, thanks.

The reason the dot converter is stand-alone is because I've not been able to find any real-time renderer for dot graphs. It could still be a publisher, but probably not very useful unless we find an efficient way to render its output. Or just make the publisher write to disk, but sounds awkward. :thinking:

ayrton04:
http://gensoft.pasteur.fr/docs/graphviz/2.38.0/pdf/libguide.pdf

> As with reading, Graphviz provides some specialized functions for rendering. Of note is
```
gvRenderData (GVC_t *gvc, Agraph_t* g, char *format, char **result,
unsigned int *length)
```
> which writes the output of the rendering onto an allocated character buffer. A pointer to this buffer is
returned in *result and the number of bytes written is stored in length. After using the buffer, the
memory should be freed by the application.

So perhaps these two methods would get it done (with whatever method is required to generate the first argument to `gvRenderData`).

```
Agraph_t* G = agmemread(char* your_dot_data_as_char_array);
gvRenderData (GVC_t *gvc, Agraph_t* g, char *format, char **result,
unsigned int *length)
```

ayrton04 (approved):

svwilliams:
Just verifying you are ready for me to merge this?
Actually, never mind. Looks like there are still some planned work.

efernandez:
@ayrton04 Thanks for the suggestion, but I'm afraid it'd be too slow.

I did some quick tests from the command line using `xdot` with different filters and it looks like the layout is pretty bad for all of them, and most importantly they run very slow. Maybe it's just the time it takes to load the GUI, but I'm afraid it's not only that. In my system I have the following filters: dot, neato, twopi, circo, or fdp. `neato` is probably the best one in both speed and layout quality/spacing.

You can use the attached file to run `xdot -f neato graph_100.txt` to get an idea. If you still think it's worth building a GraphViz real-time visualization tool, I can try implementing something.

[graph_100.dot.zip](https://github.com/locusrobotics/fuse/files/3570139/graph_100.dot.zip)

What I have in mind, if it could actually be implemented efficiently in real-time, is a `DOTPublisher` that would actually NOT publish anything, but instead it'd render the graph into a GUI. But being a publisher plugin it'd have to run together with the node running pose estimation, in the same machine. I think it's better to have a separate tool in this case, so we can run the visualizer remotely, consuming serialized graph messages. I'm not sure if we're all thinking about the same tool here. :smiley: Thoughts?

efernandez:
I wonder if I should move the DOT graph stuff into another PR, so the `rviz` display plugin can be merged as is for now. Thoughts? :smiley:

svwilliams:
Entirely up to you. No rush whatsoever.

efernandez:
@ayrton04 @svwilliams This PR is now rebased and targeting `devel`, and it only has the `rviz` display.

PR #100, opened 2019-08-27
[RST-1951] speed optimizations

svwilliams:
A couple of speed optimizations
* Changed the way random UUIDs are generated. The Boost random number generator appears to be significantly slower than the C++11 version. This changes cut total CPU time by almost half in my test cases.
* Small tweaks to copying Eigen matrices around within the marginalization code. Not sure exactly why manually copying entries is faster than entire blocks, but it cut runtime of the marginalization section by >20%


ayrton04 (approved):

fuse_core/include/fuse_core/uuid.h
```diff
@@ -183,24 +167,26 @@ namespace uuid
...
-                     id.end(),
-                     iter);
-    return generate(namespace_string, buffer.data(), buffer.size());
-  }
+  UUID generate(const std::string& namespace_string, const ros::Time& stamp, const UUID& id);
 }  // namespace uuid
 
 }  // namespace fuse_core
 
+namespace std
```
svwilliams, line 175: Not really a speed optimization. Just something nice I stole from Matt and the zone engine.

fuse_core/src/uuid.cpp
```diff
@@ -0,0 +1,154 @@
...
+// a source language processor.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE, TITLE AND NON-INFRINGEMENT. IN NO EVENT
+// SHALL THE COPYRIGHT HOLDERS OR ANYONE DISTRIBUTING THE SOFTWARE BE LIABLE
+// FOR ANY DAMAGES OR OTHER LIABILITY, WHETHER IN CONTRACT, TORT OR OTHERWISE,
+// ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+UUID generate()
```
svwilliams, line 87: This is identical to the boost function `uuid operator()()`, only using a stdlib random number generator.
https://www.boost.org/doc/libs/1_65_0/boost/uuid/random_generator.hpp

fuse_core/src/uuid.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+  // set version
+  // must be 0b0100xxxx
+  *(u.begin() + 6) &= 0x4F;  // 0b01001111
+  *(u.begin() + 6) |= 0x40;  // 0b01000000
+
+  return u;
+}
+
+UUID generate(const std::string& namespace_string, const ros::Time& stamp)
```
svwilliams, line 121: These two function implementations I just moved from the header. Since I have a cpp file now, these might as well live here too.

fuse_constraints/src/marginalize_variables.cpp
```diff
@@ -405,22 +405,30 @@ LinearTerm marginalizeNext(const std::vector<LinearTerm>& linear_terms)
...
     const auto& linear_term = linear_terms[term_index];
     auto row_offset = row_offsets[term_index];
-
     for (size_t i = 0ul; i < linear_term.variables.size(); ++i)
     {
       const auto& A = linear_term.A[i];
       auto dense = index_to_dense[linear_term.variables[i]];
       auto column_offset = column_offsets[dense];
-      Ab.block(row_offset, column_offset, A.rows(), A.cols()) = A;
+      for (int row = 0; row < A.rows(); ++row)
```
svwilliams, line 418: Not sure exactly why manually copying entries is faster than using Eigen block operations, but this cut out a significant amount of CPU time. The equivalent operation below has less impact, but is still faster.

fuse_core/src/uuid.cpp
```diff
@@ -0,0 +1,154 @@
...
+{
+
+namespace uuid
+{
+
+// Use the stdlib random number generators. They seem to be significantly faster than the Boost equivalents.
+// Implementation mimics original Boost UUID random generator implementation here:
+// https://www.boost.org/doc/libs/1_65_0/boost/uuid/random_generator.hpp
+//
+// Boost random_generator.hpp header file  ----------------------------------------------//
```
svwilliams, line 58: I have no idea what the proper license treatment is. I added a copy here as a good faith attempt.

DLu (approved):
These optimizations make me sad (because you have to do them, not that you're fixing them)

PR #101, opened 2019-08-27
[RST-2427] Added a 'source' field to the constraints

svwilliams:
Added a 'source' field to the constraints. This is an API-breaking change and hence touches a lot of files.

### DIFF
--- a/fuse_constraints/include/fuse_constraints/absolute_constraint_impl.h
+++ b/fuse_constraints/include/fuse_constraints/absolute_constraint_impl.h
@@ -48,10 +48,11 @@ namespace fuse_constraints
 
 template<class Variable>
 AbsoluteConstraint<Variable>::AbsoluteConstraint(
+  const std::string& source,
   const Variable& variable,
   const fuse_core::VectorXd& mean,
   const fuse_core::MatrixXd& covariance) :
-    fuse_core::Constraint{variable.uuid()},
+    fuse_core::Constraint(source, {variable.uuid()}),  // NOLINT(whitespace/braces)
     mean_(mean),
     sqrt_information_(covariance.inverse().llt().matrixU())
 {
@@ -62,11 +63,12 @@ AbsoluteConstraint<Variable>::AbsoluteConstraint(
 
 template<class Variable>
 AbsoluteConstraint<Variable>::AbsoluteConstraint(
+  const std::string& source,
   const Variable& variable,
   const fuse_core::VectorXd& partial_mean,
   const fuse_core::MatrixXd& partial_covariance,
   const std::vector<size_t>& indices) :
-    fuse_core::Constraint{variable.uuid()}
+    fuse_core::Constraint(source, {variable.uuid()})  // NOLINT(whitespace/braces)
 {
   assert(partial_mean.rows() == static_cast<int>(indices.size()));
   assert(partial_covariance.rows() == static_cast<int>(indices.size()));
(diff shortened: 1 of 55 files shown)


ayrton04 (approved):

fuse_constraints/include/fuse_constraints/absolute_constraint_impl.h
```diff
@@ -48,10 +48,11 @@ namespace fuse_constraints
 
 template<class Variable>
 AbsoluteConstraint<Variable>::AbsoluteConstraint(
+  const std::string& source,
   const Variable& variable,
   const fuse_core::VectorXd& mean,
   const fuse_core::MatrixXd& covariance) :
-    fuse_core::Constraint{variable.uuid()},
+    fuse_core::Constraint(source, {variable.uuid()}),  // NOLINT(whitespace/braces)
```
efernandez, line 55: I think this should also work: `fuse_core::Constraint{source, variable.uuid()},` (same in other instances below)

fuse_constraints/include/fuse_constraints/relative_constraint_impl.h
```diff
@@ -48,11 +48,12 @@ namespace fuse_constraints
 
 template<class Variable>
 RelativeConstraint<Variable>::RelativeConstraint(
+  const std::string& source,
   const Variable& variable1,
   const Variable& variable2,
   const fuse_core::VectorXd& delta,
   const fuse_core::MatrixXd& covariance) :
-    fuse_core::Constraint{variable1.uuid(), variable2.uuid()},
+    fuse_core::Constraint(source, {variable1.uuid(), variable2.uuid()}),  // NOLINT(whitespace/braces)
```
efernandez, line 56: As mentioned before, I think this should work: `fuse_core::Constraint{source, variable1.uuid(), variable2.uuid()},` (same below)

fuse_constraints/src/absolute_orientation_3d_stamped_constraint.cpp
```diff
@@ -40,33 +40,38 @@
...
 namespace fuse_constraints
 {
 
 AbsoluteOrientation3DStampedConstraint::AbsoluteOrientation3DStampedConstraint(
+  const std::string& source,
   const fuse_variables::Orientation3DStamped& orientation,
   const fuse_core::Vector4d& mean,
   const fuse_core::Matrix3d& covariance) :
-    fuse_core::Constraint{orientation.uuid()},
+    fuse_core::Constraint(source, {orientation.uuid()}),  // NOLINT(whitespace/braces)
```
efernandez, line 54: `fuse_core::Constraint{source, orientation.uuid()},`

efernandez (approved):
LGTM, but it seems to me that you're missing direct initialization (https://en.cppreference.com/w/cpp/language/direct_initialization) with list initialization (https://en.cppreference.com/w/cpp/language/list_initialization), or maybe aggregate initialization (https://en.cppreference.com/w/cpp/language/aggregate_initialization) :thinking:

fuse_constraints/include/fuse_constraints/absolute_constraint_impl.h
svwilliams: That particular incantation results in:
```
error: no matching function for call to ‘std::vector<boost::uuids::uuid>::vector(boost::uuids::uuid&)’
```
This does work, but I'm not sure it has any advantages over the constructor `()` version:
```
fuse_core::Constraint{source, {variable.uuid()}}
```
I could do something with variadic templates:
https://stackoverflow.com/questions/25721869/initializer-list-combined-with-other-parameters
```
public:
  template <class ...Args>
  Constraint(const std::string& source, Args... args) : Constraint(source, {args...}) {}
private:
  Constraint(const std::string& source, std::initializer_list<UUID> variable_uuid_list) :
    source_(source),
    variables_(variable_uuid_list)
  {}
```
but I'm not sure it is worth the trouble.

efernandez: Fair enough. I don't think it's worth the trouble, and I didn't realized there's an `std::initializer_list` for the UUIDs.

PR #103, opened 2019-08-31
[RST-2437] Ensure that all variables are updated

svwilliams:
Ensure that all variables are updated when the update_variables flag is set. Previously, only variables involved in motion model segments that needed to be generated were being updated. Variables involved in previously generated segments were left untouched. This could result in incorrect initial state guesses being used for those variables.

As an example, imagine a sensor transaction that connects two adjacent states at T1 and T2, and a second transaction that connects T1 to T3. If both of these transactions are added to the graph at the same time:
* Transaction1 is processed. A motion model segment is generated from T1 to T2. Since states T1 and T2 are involved in the newly generated motion model segment, their states are updated. No other states are involved in this transaction, so all states have been updated.
* Transaction2 is processed. A motion model is generated from T2 to T3. Since a motion model was previously generated from T1 to T2, the motion model does not need to regenerate that segment. Since states T2 and T3 are involved in the newly generated motion model segment, their states are updated. However, the transaction also involves the state at T1. It was not involved in the generated motion model segment, so its state has not been updated. The original state from Transaction2 overwrites the updated state from Transaction1. This can lead to a poor initial guess being sent to the optimizer.

With this PR, if the `update_variables` flag is set, any variable from a previously generated motion model segment that is also involved in the current transaction is updated from the value stored in the motion model history.

### DIFF
--- a/fuse_core/src/timestamp_manager.cpp
+++ b/fuse_core/src/timestamp_manager.cpp
@@ -123,9 +123,9 @@ void TimestampManager::query(
             if (std::any_of(
                   transaction_variables.begin(),
                   transaction_variables.end(),
-                  [&variable](const fuse_core::Variable& input_variable)
+                  [variable_uuid = variable->uuid()](const auto& input_variable)
                   {
-                    return input_variable.uuid() == variable->uuid();
+                    return input_variable.uuid() == variable_uuid;
                   }))  // NOLINT
             {
               motion_model_transaction.addVariable(variable, update_variables);
--- a/fuse_core/src/timestamp_manager.cpp
+++ b/fuse_core/src/timestamp_manager.cpp
@@ -81,7 +81,7 @@ void TimestampManager::query(
   Transaction motion_model_transaction;
   std::set<ros::Time> augmented_stamps(stamps.begin(), stamps.end());
   auto first_stamp = *augmented_stamps.begin();
-  ros::Time last_stamp = *augmented_stamps.rbegin();
+  auto last_stamp = *augmented_stamps.rbegin();
   {
     auto begin = motion_model_history_.upper_bound(first_stamp);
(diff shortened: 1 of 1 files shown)


fuse_core/src/timestamp_manager.cpp
```diff
@@ -78,18 +79,9 @@ void TimestampManager::query(
...
-    auto iter = stamps.begin();
-    while (std::next(iter) != stamps.end())
-    {
-      ++iter;
-    }
-    last_stamp = *iter;
-  }
   std::set<ros::Time> augmented_stamps(stamps.begin(), stamps.end());
+  auto first_stamp = *augmented_stamps.begin();
+  ros::Time last_stamp = *augmented_stamps.rbegin();
```
efernandez: Maybe do `auto last_stamp` as you do with `first_stamp`

efernandez (approved):

ayrton04 (approved):

fuse_core/src/timestamp_manager.cpp
```diff
@@ -121,6 +113,25 @@ void TimestampManager::query(
...
+        {
+          // Add the motion model version of the variables involved in this motion model segment
+          // This ensures that the variables in the final transaction will be overwritten with the motion model version
+          auto transaction_variables = transaction.addedVariables();
+          for (const auto& variable : history_iter->second.variables)
+          {
+            if (std::any_of(
+                  transaction_variables.begin(),
+                  transaction_variables.end(),
+                  [&variable](const fuse_core::Variable& input_variable)
```
efernandez: In C++14 you can do:
``` c++
[variable_uuid = variable->uuid()](const auto& input_variable)
{
  return input_variable.uuid() == variable_uuid;
}
```

PR #109, opened 2019-10-15
Publish odometry with timer and allow to predict it

efernandez:
* Add `predict_odom` parameter that:
  1. Disables publishing the odometry in the `notifyCallback`
  2. Enables publishing the odometry in the `tfPublishTimerCallback`, predicting it into the current time.
     The Jacobian of the motion model used in the predict function is computed so the odometry covariance can be propagated.
     The odometry cached in the class is updated with the predicted pose, covariance and time stamp.
* Add `process_noise_diagonal` parameter that defines the diagonal of the process noise covariance that is added to the propagated covariance computing after predicting the odometry into the current time.
  The process noise covariance has 8x8 size, so it can represent all the state, although the acceleration is ignored at the moment.
* Add new `predict` free functions to compute the Jacobian of the motion model wrt the state. Two versions are provided:
  1. One that takes each single component and a plain array of double to
     the Jacobian.
  2. Another that takes aggregate objects like `tf2_2d::Transform` and an `fuse_core::Matrix8d` for the Jacobian.
* Cache the stamp of the latest covariance successfully computed in the `notifyCallback`, so it's possible to know if the covariance is valid and up to date in the `tfPublishTimerCallback`.
  If the covariance is invalid or not up to date, no time is wasted propagating it.

### DIFF
--- a/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
+++ b/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
@@ -73,8 +73,7 @@ public:
   {
     nh.getParam("publish_tf", publish_tf);
     nh.getParam("predict_to_current_time", predict_to_current_time);
-    nh.getParam("predict_odom", predict_odom);
-    nh.getParam("tf_publish_frequency", tf_publish_frequency);
+    nh.getParam("publish_frequency", publish_frequency);
 
     std::vector<double> process_noise_diagonal(8, 0.0);
     nh.param("process_noise_diagonal", process_noise_diagonal, process_noise_diagonal);
@@ -133,8 +132,7 @@ public:
 
   bool publish_tf { true };
   bool predict_to_current_time { false };
-  bool predict_odom { false };
-  double tf_publish_frequency { 10.0 };
+  double publish_frequency { 10.0 };
   fuse_core::Matrix8d process_noise_covariance;   //!< Process noise covariance matrix
   ros::Duration tf_cache_time { 10.0 };
   ros::Duration tf_timeout { 0.1 };
--- a/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
+++ b/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
@@ -44,6 +44,7 @@
 
 #include <ceres/covariance.h>
 
+#include <algorithm>
 #include <cassert>
 #include <string>
 #include <vector>
@@ -81,6 +82,12 @@ public:
     if (process_noise_diagonal.size() != 8)
     {
       throw std::runtime_error("Process noise diagonal must be of length 8!");
+    }
+
+    if (std::any_of(process_noise_diagonal.begin(), process_noise_diagonal.end(),
+                    [](const auto& v) { return v < 0.0; }))  // NOLINT(whitespace/braces)
+    {
+      throw std::runtime_error("All process noise diagonal entries must be positive!");
     }
 
     process_noise_covariance = fuse_core::Vector8d(process_noise_diagonal.data()).asDiagonal();
--- a/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
+++ b/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
@@ -37,6 +37,7 @@
 #include <fuse_models/parameters/parameter_base.h>
 
 #include <fuse_core/ceres_options.h>
+#include <fuse_core/eigen.h>
 
 #include <ros/console.h>
 #include <ros/node_handle.h>
@@ -60,6 +61,8 @@ namespace parameters
 struct Odometry2DPublisherParams : public ParameterBase
 {
 public:
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
   /**
    * @brief Method for loading parameter values from ROS.
    *
@@ -69,7 +72,18 @@ public:
   {
     nh.getParam("publish_tf", publish_tf);
     nh.getParam("predict_to_current_time", predict_to_current_time);
+    nh.getParam("predict_odom", predict_odom);
     nh.getParam("tf_publish_frequency", tf_publish_frequency);
+
+    std::vector<double> process_noise_diagonal(8, 0.0);
+    nh.param("process_noise_diagonal", process_noise_diagonal, process_noise_diagonal);
+
+    if (process_noise_diagonal.size() != 8)
+    {
+      throw std::runtime_error("Process noise diagonal must be of length 8!");
+    }
+
+    process_noise_covariance = fuse_core::Vector8d(process_noise_diagonal.data()).asDiagonal();
 
     double tf_cache_time_double = tf_cache_time.toSec();
     nh.getParam("tf_cache_time", tf_cache_time_double);
@@ -112,7 +126,9 @@ public:
 
   bool publish_tf { true };
   bool predict_to_current_time { false };
+  bool predict_odom { false };
   double tf_publish_frequency { 10.0 };
+  fuse_core::Matrix8d process_noise_covariance;   //!< Process noise covariance matrix
   ros::Duration tf_cache_time { 10.0 };
   ros::Duration tf_timeout { 0.1 };
   int queue_size { 1 };
--- a/fuse_models/include/fuse_models/unicycle_2d_predict.h
+++ b/fuse_models/include/fuse_models/unicycle_2d_predict.h
@@ -36,6 +36,7 @@
 
 #include <ceres/jet.h>
 #include <fuse_core/util.h>
+#include <fuse_core/eigen.h>
 #include <tf2_2d/tf2_2d.h>
 
 
@@ -100,6 +101,105 @@ inline void predict(
   fuse_core::wrapAngle2D(yaw2);
 }
 
+enum StateIndex : uint8_t
+{
+  X,
+  Y,
+  YAW,
+  V_X,
+  V_Y,
+  V_YAW,
+  A_X,
+  A_Y
+};
+
+/**
+ * @brief Given a state and time delta, predicts a new state
+ * @param[in] position1_x - First X position
+ * @param[in] position1_y - First Y position
+ * @param[in] yaw1 - First orientation
+ * @param[in] vel_linear1_x - First X velocity
+ * @param[in] vel_linear1_y - First Y velocity
+ * @param[in] vel_yaw1 - First yaw velocity
+ * @param[in] acc_linear1_x - First X acceleration
+ * @param[in] acc_linear1_y - First Y acceleration
+ * @param[in] dt - The time delta across which to predict the state
+ * @param[in] position2_x - Second X position
+ * @param[in] position2_y - Second Y position
+ * @param[in] yaw2 - Second orientation
+ * @param[in] vel_linear2_x - Second X velocity
+ * @param[in] vel_linear2_y - Second Y velocity
+ * @param[in] vel_yaw2 - Second yaw velocity
+ * @param[in] acc_linear2_x - Second X acceleration
+ * @param[in] acc_linear2_y - Second Y acceleration
+ * @param[in] jacobians - Jacobians wrt the state
+ */
+// TODO(efernandez) fuse this with the templated predict function so we can use the analytic Jacobian in the cost
+// function. See http://ceres-solver.org/analytical_derivatives.html
+inline void predict(
+  const double position1_x,
+  const double position1_y,
+  const double yaw1,
+  const double vel_linear1_x,
+  const double vel_linear1_y,
+  const double vel_yaw1,
+  const double acc_linear1_x,
+  const double acc_linear1_y,
+  const double dt,
+  double& position2_x,
+  double& position2_y,
+  double& yaw2,
+  double& vel_linear2_x,
+  double& vel_linear2_y,
+  double& vel_yaw2,
+  double& acc_linear2_x,
+  double& acc_linear2_y,
+  // TODO(efernandez) double** jacobians didn't play well with Eigen::Map<>, at least directly
+  double* jacobians)
+{
+  const double sy = ceres::sin(yaw1);
+  const double cy = ceres::cos(yaw1);
+
+  const double half_dt2 = 0.5 * dt * dt;
+  const double delta_x = vel_linear1_x * dt + acc_linear1_x * half_dt2;
+  const double delta_y = vel_linear1_y * dt + acc_linear1_y * half_dt2;
+
+  const double delta_x_rot = cy * delta_x - sy * delta_y;
+  const double delta_y_rot = sy * delta_x + cy * delta_y;
+
+  position2_x = position1_x + delta_x_rot;
+  position2_y = position1_y + delta_y_rot;
+  yaw2 = yaw1 + vel_yaw1 * dt;
+  vel_linear2_x = vel_linear1_x + acc_linear1_x * dt;
+  vel_linear2_y = vel_linear1_y + acc_linear1_y * dt;
+  vel_yaw2 = vel_yaw1;
+  acc_linear2_x = acc_linear1_x;
+  acc_linear2_y = acc_linear1_y;
+
+  fuse_core::wrapAngle2D(yaw2);
+
+  if (!jacobians)
+  {
+    return;
+  }
+
+  Eigen::Map<fuse_core::Matrix8d> J(jacobians);
+  J.setIdentity();
+  J(X, YAW) = -delta_y_rot;
+  J(X, V_X) = cy * dt;
+  J(Y, V_X) = sy * dt;
+  J(X, V_Y) = -J(Y, V_X);
+  J(X, A_X) = 0.5 * J(X, V_X) * dt;
+  J(X, A_Y) = 0.5 * J(X, V_Y) * dt;
+  J(Y, YAW) = delta_x_rot;
+  J(Y, V_Y) = J(X, V_X);
+  J(Y, A_X) = -J(X, A_Y);
+  J(Y, A_Y) = J(X, A_X);
+  J(YAW, V_YAW) = dt;
+  J(V_X, A_X) = dt;
+  J(V_Y, A_Y) = dt;
+}
+
 /**
  * @brief Given a state and time delta, predicts a new state
  * @param[in] position1 - First position (array with x at index 0, y at index 1)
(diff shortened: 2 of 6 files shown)


efernandez:
@svwilliams All `roslint` and `tests` are passing for `fuse_models` pkg locally for me. Could you share the test/s failure output here, please?

svwilliams:
Best I can tell, all of the CI tests have passed. I'll get the review done
later today.

efernandez:
nvm, I removed my comment about the CI test failures because I figured out the `roslint` warnings and I fixed them.

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
```diff
@@ -69,8 +72,19 @@ struct Odometry2DPublisherParams : public ParameterBase
   {
     nh.getParam("publish_tf", publish_tf);
     nh.getParam("predict_to_current_time", predict_to_current_time);
+    nh.getParam("predict_odom", predict_odom);
     nh.getParam("tf_publish_frequency", tf_publish_frequency);
 
+    std::vector<double> process_noise_diagonal(8, 0.0);
+    nh.param("process_noise_diagonal", process_noise_diagonal, process_noise_diagonal);
+
+    if (process_noise_diagonal.size() != 8)
```
svwilliams, line 81: Maybe also verify all the entries are positive

fuse_models/src/odometry_2d_publisher.cpp
```diff
@@ -254,21 +262,114 @@ void Odometry2DPublisher::tfPublishTimerCallback(const ros::TimerEvent& event)
...
-      event.current_real.toSec() - odom_output_.header.stamp.toSec(),
-      pose,
-      velocity_linear,
-      unused_yaw_vel,
-      unused_acc);
-
-    trans.header.stamp = event.current_real;
+    const double dt = event.current_real.toSec() - odom_output_.header.stamp.toSec();
+
+    if (params_.predict_odom)
```
svwilliams: Here, you only publish if both `params_.predict_to_current_time` and `params_.predict_odom` are true. But on line 171 (https://github.com/locusrobotics/fuse/pull/109/files#diff-ec05e4b7290a6166c67ff2570f66529fR171), you disable publishing when only `params_.predict_odom` is true. Thus, if you have `params_.predict_to_current_time == false` and `params_.predict_odom == true`, you will never publish an odom message.

Reviewing the `robot_localization` package, it appears that the odom and tf messages are tied together -- the odom and tf messages always publish the same state at the same time. I wonder if we should move the odom publication into the timer callback, and have the `params_.predict_to_current_time` setting control both odom and tf?

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -100,6 +101,105 @@ inline void predict(
...
+  if (!jacobians)
+  {
+    return;
+  }
+
+  Eigen::Map<fuse_core::Matrix8d> J(jacobians);
+  J.setIdentity();
+  J(X, YAW) = -delta_y_rot;
+  J(X, V_X) = cy * dt;
+  J(Y, V_X) = sy * dt;
```
svwilliams, line 190: Can you move this one down with its `J(Y, ...)` friends?
```
J(X, V_X) = cy * dt;
J(X, V_Y) = -sy * dt;
...
J(Y, V_X) = -J(X, V_Y);
J(Y, V_Y) = J(X, V_X);
```

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -100,6 +101,105 @@ inline void predict(
...
+ * @param[in] position2_y - Second Y position
+ * @param[in] yaw2 - Second orientation
+ * @param[in] vel_linear2_x - Second X velocity
+ * @param[in] vel_linear2_y - Second Y velocity
+ * @param[in] vel_yaw2 - Second yaw velocity
+ * @param[in] acc_linear2_x - Second X acceleration
+ * @param[in] acc_linear2_y - Second Y acceleration
+ * @param[in] jacobians - Jacobians wrt the state
+ */
+// TODO(efernandez) fuse this with the templated predict function so we can use the analytic Jacobian in the cost
```
svwilliams, line 137: I would really like to get these two functions merged.

This could be done by easily by defaulting the `jacobian` variable to null.
```
template<typename T>
inline void predict(
  const T position1_x,
  ...
  T& acc_linear2_y,
  T* jacobian = nullptr
{
  ...
  if (!jacobian) return;
  ...
}
```
This would completely replace the existing templated function. The other two signatures (that involve tf2_2d objects) would call this function, providing a non-null jacobian when appropriate.

Another option that seems more C++/type-safe would be to use a `boost::optional`. I really wish `std::optional` was included in C++14.
```
template<typename T>
inline void predict(
  const T position1_x,
  ...
  T& acc_linear2_y,
  boost::optional<Eigen::Matrix<T, 8, 8>&> jacobian
{
  ...
  if (!jacobian) return;
  ...
}
```
where I'm not exactly sure what Eigen type should be passed in. An EigenBase<T, 8, 8> maybe???? I'm just making things up now. In the case of using this function to compute Ceres cost function Jacobians, Ceres will provide a C-style double array. Within the cost function implementation, you could use `Eigen::Map<>` to convert the double array into an Eigen compatible type. I feel that should be possible, but don't know the exact type required off the top of my head.

Ultimately I'm fine with either approach. I just don't like having the duplicate code that exists now.

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -100,6 +101,105 @@ inline void predict(
...
+  J(X, V_Y) = -J(Y, V_X);
+  J(X, A_X) = 0.5 * J(X, V_X) * dt;
+  J(X, A_Y) = 0.5 * J(X, V_Y) * dt;
+  J(Y, YAW) = delta_x_rot;
+  J(Y, V_Y) = J(X, V_X);
+  J(Y, A_X) = -J(X, A_Y);
+  J(Y, A_Y) = J(X, A_X);
+  J(YAW, V_YAW) = dt;
+  J(V_X, A_X) = dt;
+  J(V_Y, A_Y) = dt;
```
svwilliams, line 200: Can you write a unit test that uses the Ceres autodiff system to prove the Jacobians are correct? I mean, they look correct, but I've been known to make a sign error on occasion.

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -100,6 +101,105 @@ inline void predict(
...
+  double& yaw2,
+  double& vel_linear2_x,
+  double& vel_linear2_y,
+  double& vel_yaw2,
+  double& acc_linear2_x,
+  double& acc_linear2_y,
+  // TODO(efernandez) double** jacobians didn't play well with Eigen::Map<>, at least directly
+  double* jacobians)
+{
+  const double sy = ceres::sin(yaw1);
```
ayrton04, line 160: +1 for merging the methods, but if we don't, then there's no reason to use the `ceres` versions of these, right?

fuse_models/src/odometry_2d_publisher.cpp
ayrton04: If we're going for feature parity, that would make sense.

ayrton04 (approved):
LGTM, Stephen's feedback notwithstanding.

fuse_models/include/fuse_models/unicycle_2d_predict.h
efernandez: I was trying to save one extra `-` operation. If you still think it's better to move it, I'd do it.

fuse_models/include/fuse_models/unicycle_2d_predict.h
efernandez: I tried that and it didn't work because `predict` is used inside a cost function and Ceres requires the `jacobian` to be a `T**` double pointer: http://ceres-solver.org/analytical_derivatives.html

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
efernandez: Done.

BTW I also wasn't sure of the default value for the process noise diagonal. I set it to `0.0`, that in principle shouldn't be allowed for a covariance matrix, but in the case of the process noise covariance it doesn't have any negative impact AFAIK.

Also note this is similar to https://github.com/locusrobotics/fuse/blob/devel/fuse_models/src/unicycle_2d.cpp#L99-L107, where there's no check for the diagonal entries.

fuse_models/include/fuse_models/unicycle_2d_predict.h
efernandez: Done

fuse_models/src/odometry_2d_publisher.cpp
efernandez: Yep, that would make the params easier to understand. If you both agree, I can do the following:
1. Remove the odometry message publication from the `notifyCallback`
2. Remove the `predict_odom` param
3. Publish the TF and odometry message from `tfPublishTimerCallback` only, and always
4. Since both the TF and odometry messsage are published from the timer callback, rename it to `publishTimerCallback`, and I could also rename the other related variables, and even the `tf_publish_frequency` parameter
5. If `predict_to_current_time` is enabled, predict both the TF and the odometry message

BTW I was thinking of also publishing the accelerations in another topic. That could be part of another PR, but it'd be certainly easier if we do simplify things as explained above.

efernandez:
I've addressed some of your review comments and replied to the others. I'm going to wait for your feedback before I continue implementing more changes on this PR.

Thanks for your review. This PR was a first iteration on this topic, so I was indeed looking for that kind of feedback. :smiley:

fuse_models/src/odometry_2d_publisher.cpp
svwilliams: That all sounds good to me. And I'd put the accel publish in a separate PR, if you don't mind.

fuse_models/include/fuse_models/unicycle_2d_predict.h
svwilliams: ¯\\\_(ツ)\_\/¯
Not a big deal one way or the other, now that we have unit test validation of the code.

fuse_models/test/test_unicycle_2d_predict.cpp
```diff
@@ -322,6 +323,119 @@ TEST(Predict, predictObjects)
...
+                jet_acc_linear2_x.v,
+                jet_acc_linear2_y.v;
+
+  J_autodiff.transposeInPlace();
+
+  const Eigen::IOFormat HeavyFmt(
+      Eigen::FullPrecision, 0, ", ", ";\n", "[", "]", "[", "]");
+
+  const fuse_core::Matrix8d J_diff = J_autodiff - J;
+  EXPECT_TRUE((J_diff.array().abs() < std::numeric_limits<double>::epsilon()).all())
```
svwilliams: FYI: I got tired of writing similar code constantly, so I made `EXPECT_MATRIX_NEAR`
https://github.com/locusrobotics/fuse/blob/devel/fuse_core/include/fuse_core/eigen_gtest.h

Feel free to tweak the output format, etc. if you have better suggestions.

fuse_models/include/fuse_models/unicycle_2d_predict.h
svwilliams: Right. I guess what I was expecting was `predict()` would be wrapped in a derived ceres::CostFunction. The way Ceres populates jacobians is not directly compatible with the new predict() function anyway. So something vaguely along the lines of:
```
class Unicycle2DStateCostFunction : public ceres::CostFunction
{
public:
  bool Evaluate(
    double const* const* parameters,
    double* residuals,
    double** jacobians) const override
  {
    boost::optional<Eigen::Matrix8d> jacobian_matrix = boost::none;
    if (jacobians)
    {
      jacobian_matrix = Eigen::Matrix8d();
    }

    predict(
      position1,
      yaw1,
      vel_linear1,
      vel_yaw1,
      acc_linear1,
      dt_,
      position_pred,
      yaw_pred,
      vel_linear_pred,
      vel_yaw_pred,
      acc_linear_pred,
      jacobian_matrix);  // Only computes jacobians if Ceres wants them

    // Populate the Ceres-requested jacobian blocks
    // I'm having to think about what goes here too hard for a code review.
    // The predict() function computes State2 as a function of State1.
    // But the cost function computes (conceptually) "error = State2 - State1"
    // I'm sure there is a way to translate the predict() jacobian into a Ceres
    // jacobian. Regardless, Ceres wants the jacobian in blocks, not as one
    // giant matrix, so some copying will be required.
    if (jacobians)
    {
      if (jacobians[0])
      {
        Eigen::Map<fuse_core::MatrixXd>(jacobians[0], 8, 2) = jacobian_matrix->block(0, 0, 8, 2);
        // Maybe???
      }
      if (jacobians[1])
      {
        ...
      }
      ...
    }
  }
```

fuse_models/test/test_unicycle_2d_predict.cpp
efernandez: Pretty useful. Updated

fuse_models/src/odometry_2d_publisher.cpp
efernandez: Done.

- [x] Remove the odometry message publication from the notifyCallback
- [x] Remove the predict_odom param
- [x] Publish the TF and odometry message from tfPublishTimerCallback only, and always
- [x] Since both the TF and odometry messsage are published from the timer callback, rename it to `publishTimerCallback` and also the timer attribute to `publish_timer_`.
- [x] Also rename the `tf_publish_frequency` parameter to `publish_frequency`. I did this on a separate commit in case we don't want that.
- [x] If predict_to_current_time is enabled, predict both the TF and the odometry message
- [x] :exclamation: I also notice the `publish_tf` parameter wasn't used, so I 've disable broadcasting the TF if it's `false`.

The PR should be reviewed again to confirm all this is good. It should be easier to review now, indeed. I still need to merge the `predict` function and use it as a `ceres::CostFunction`, but I'll have to leave that for next week. I also need to test now that everything still works as intended.

fuse_models/include/fuse_models/unicycle_2d_predict.h
efernandez: Will try merging the methods.

fuse_models/include/fuse_models/unicycle_2d_predict.h
efernandez: Will do that next week. All other changes requested are in. I'm going to validate things still work as intended.

fuse_models/src/odometry_2d_publisher.cpp
```diff
@@ -245,63 +248,137 @@ void Odometry2DPublisher::tfPublishTimerCallback(const ros::TimerEvent& event)
...
-      ROS_WARN_STREAM_THROTTLE(5.0, "Could not lookup the " << params_.base_link_frame_id << "->" <<
-        params_.odom_frame_id << " transform. Error: " << e.what());
 
-      return;
-    }
+    odom_pub_.publish(odom_output_);
   }
 
-  tf_broadcaster_.sendTransform(trans);
+  if (params_.publish_tf)
```
efernandez, line 346: `publish_tf` wasn't used before. That was a bug I just discovered.

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
```diff
@@ -69,7 +73,23 @@ struct Odometry2DPublisherParams : public ParameterBase
...
+    {
+      throw std::runtime_error("Process noise diagonal must be of length 8!");
+    }
+
+    if (std::any_of(process_noise_diagonal.begin(), process_noise_diagonal.end(),
+                    [](const auto& v) { return v < 0.0; }))  // NOLINT(whitespace/braces)
```
svwilliams, line 87:
```suggestion
                    [](const auto& v) { return v <= 0.0; }))  // NOLINT(whitespace/braces)
```
You're probably duplicating code I wrote elsewhere, 0's really shouldn't be allowed either.

svwilliams (commented):
Changes look good. All that's left is to merge the predict functions in some usable way.

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
efernandez: If I don't allow `0.0`, then the default value for `process_noise_diagonal` can't be all `0.0`s as it's now. Then the question is what value should I use in that case. Maybe something like `1e-6`? But the diagonal represents different state components, with different units, so some should be smaller than others, like the velocities or accelerations. :thinking: 

The reason I allowed `0.0`s in this case is because it's the process noise covariance, so if the user doesn't specify it, then it'd be like there's no process noise, that it's not recommended, but the math won't blow up. Maybe I can add a `ROS_WARN` saying that it's not recommended to have zeros in the process noise covariance diagonal.

Regarding the check, I can only find this in unicycle 2D motion model, but it doesn't perform the check I do here. Maybe it should also have a check similar to mine.
https://github.com/locusrobotics/fuse/blob/9933456ecc24ba9b649a8a2885be3f852306efee/fuse_models/src/unicycle_2d.cpp#L99-L105

fuse_models/src/odometry_2d_publisher.cpp
```diff
@@ -245,63 +248,137 @@ void Odometry2DPublisher::tfPublishTimerCallback(const ros::TimerEvent& event)
...
+      odom_output_.twist.covariance[35] = covariance(5, 5);
     }
-    catch (const std::exception& e)
-    {
-      ROS_WARN_STREAM_THROTTLE(5.0, "Could not lookup the " << params_.base_link_frame_id << "->" <<
-        params_.odom_frame_id << " transform. Error: " << e.what());
 
-      return;
-    }
+    odom_pub_.publish(odom_output_);
```
efernandez: @svwilliams @ayrton04 I just fixed this. I have this instruction inside the `if (params_.predict_to_current_time)` block, so if that parameter was `false`, the odometry was never published. Now it's always published, either with the optimization or predicted value. Let me know if you think something is wrong with this.

efernandez: I wrote the comment before pushing the changes, but you can simply search for `publish(` and se e that now is outside that `if` block.

fuse_models/src/odometry_2d_publisher.cpp
```diff
@@ -245,63 +248,137 @@ void Odometry2DPublisher::tfPublishTimerCallback(const ros::TimerEvent& event)
...
-      geometry_msgs::TransformStamped map_to_odom;
-      tf2::doTransform(base_to_odom, map_to_odom, trans);
-      map_to_odom.child_frame_id = params_.odom_frame_id;
-      trans = map_to_odom;
+      try
+      {
+        auto base_to_odom = tf_buffer_->lookupTransform(
+          params_.base_link_frame_id,
+          params_.odom_frame_id,
+          trans.header.stamp,
```
efernandez, line 363: @svwilliams @ayrton04 This was a bug! Before the time stamp passed to `lookupTransform` was `latest_stamp_`, but if `predict_to_current_time` is `true`, this should be `event.current_real` instead. Using `trans.header.stamp` we get the correct stamp in both cases.

Because of this, with a robot turning in place the pose estimation overshoots. This is easy to see looking at a LIDAR scan alignment wrt the map. After this fix, the scans aligned perfectly. :smiley: 

Let me know if you want this on another PR, only with this change, as it looks like a big issue.

efernandez:
@svwilliams I still need to merge the `predict` functions, but I also found a couple of issues. See my comments in the code. I've thoroughly tested this brunch in sim and real data. I'll continue doing so, but now I'm more confident on the code contributed here.

I'll try to get the `predict` functions merged asap, hopefully over the weekend. But maybe it's better that this PR gets merged and I create another one after, so it's easier to review.

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
svwilliams: I expect the math will blow up actually. Inside the motion model cost function, the error is normalized by the square root information matrix. If there are any zero terms in the process noise, the covariance cannot be inverted into an information matrix.

Although this is not that process noise, I guess.

efernandez: Exactly, the `process_noise_diagonal` set here is NOT used in: https://github.com/locusrobotics/fuse/blob/devel/fuse_models/src/unicycle_2d.cpp#L246 and https://github.com/locusrobotics/fuse/blob/devel/fuse_models/src/unicycle_2d_state_kinematic_constraint.cpp#L81, where a covariance with zeros will certainly blow up.

This one is only used in https://github.com/locusrobotics/fuse/pull/109/files#diff-ec05e4b7290a6166c67ff2570f66529fR320, that will simply make that addition a nop.

That being said, if we agree on some initial values, or some WARN messages, I could add that, so the user is aware of a potentially bad configuration.

Ideally, the publisher should be using the same process noise covariance as the unicycle_2d motion model. Indeed, I think the publisher should allow the user to specify the motion model to use for prediction, but that would be another story, for another PR.

svwilliams:
Merging this PR, and expecting a second "predict" PR at some point in the future. :)

svwilliams (approved):

PR #112, opened 2019-10-29
fix compilation in Kinetic

facontidavide:


svwilliams:
There is a build server running that tests Xenial/Kinetic builds, but I believe that build server has a set of default compile flags that get added. So these errors slipped through the cracks. So thanks for this PR.

I am going to attempt to fix this the "correct" way by specifying the needed `target_compile_features()`. If I can work it out, I'll post a link to an alternate PR for you to try in a day or two. And if I can't, I'll get this merged.

efernandez:
This looks more like it's trying to disable gnu extensions, which it's something I encountered trying to use Ceres 2.0, although I'm not using it at the moment: https://cmake.org/cmake/help/v3.15/variable/CMAKE_CXX_STANDARD.html, https://cmake.org/cmake/help/v3.15/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD, https://cmake.org/cmake/help/v3.15/prop_tgt/CXX_EXTENSIONS.html#prop_tgt:CXX_EXTENSIONS

If that's the intention, we should do `set(CMAKE_CXX_EXTENSIONS OFF)`, as it's already done in `fuse_constraints` and `fuse_core`.

:+1: for using `target_compile_features`. IIRC only newer versions of CMake propagate this stuff properly though. FYI I'm using version `3.5.1` in my system. I wonder which version are you guys using.

svwilliams:
@facontidavide Can I get some more details about the compilation issue you are experiencing?

I created a virtual machine running Ubuntu 16.04 + ROS Kinetic, made a workspace containing the fuse packages and tf2_2d, and installed all dependencies using rosdep. Everything compiled without error, and all of the unit tests passed.

I want to make sure I can reproduce the issue, so that I know any compilation changes I make will fix it.

@efernandez `target_compile_features` was introduced in Cmake 3.1. Ubuntu 16.04's default version of CMake is 3.5.1. I feel safe increasing the required cmake version to 3.1.

facontidavide:
For some reason it was complaining about variadic lambdas. But this happened only on one of my two laptops, so it might be hard to reproduce

svwilliams:
My attempt at using `target_compile_features()` was not particularly successful. I am reverting to the suggestion here to set `CMAKE_CXX_STANDARD` instead. I have a [test branch](https://github.com/locusrobotics/fuse/tree/kinetic-compile), if you have time to give it a try. It starts with your PR and cleans up some additional compile flags. If that works for you, I'll add my changes to this PR and get it merged in.

svwilliams (approved):

ayrton04 (approved):

PR #113, opened 2019-10-29
Use measurement stamps for transformed variables

sjphilli:
This PR fixes an issue where the graph is incorrectly updated due to sensor measurement transactions with inconsistent timestamps. This is minor, but it came up in a test case and _I don't think_ it's intended behaviour.

Currently there is a discrepancy in how transactions and variables are timestamped for the built-in sensor models where transforms are applied (e.g., pose message received in frame F1 with target frame F2). For transactions, the original measurement stamp is used to add involved stamps, however for variables, the transformed stamp (based on transform time) is used. 
This is problematic in cases where [older](https://github.com/locusrobotics/fuse/blob/9933456ecc24ba9b649a8a2885be3f852306efee/fuse_models/include/fuse_models/common/sensor_proc.h#L217) transforms are used - i.e., subsequent motion models will generate (and stamp) variables based on the transaction's involved stamps, resulting in different uuids than any corresponding measured variables - the end result being that some variables are added to the graph twice, where the measurement 'version' of the variable is not connected.

### DIFF
--- a/fuse_models/include/fuse_models/common/sensor_proc.h
+++ b/fuse_models/include/fuse_models/common/sensor_proc.h
@@ -273,8 +273,8 @@ inline bool processAbsolutePoseWithCovariance(
   tf2::fromMsg(transformed_message.pose.pose, absolute_pose_2d);
 
   // Create the pose variable
-  auto position = fuse_variables::Position2DStamped::make_shared(transformed_message.header.stamp, device_id);
-  auto orientation = fuse_variables::Orientation2DStamped::make_shared(transformed_message.header.stamp, device_id);
+  auto position = fuse_variables::Position2DStamped::make_shared(pose.header.stamp, device_id);
+  auto orientation = fuse_variables::Orientation2DStamped::make_shared(pose.header.stamp, device_id);
   position->x() = absolute_pose_2d.x();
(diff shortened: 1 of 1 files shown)


svwilliams:
You are correct. That is problematic.

In addition to the changes you are proposing, I'm wondering if transforming data at non-header timestamps should be disallowed?
https://github.com/locusrobotics/fuse/blob/c269114f4e9804aad9518619c2392068601d1c3f/fuse_models/include/fuse_models/common/sensor_proc.h#L213

Because I'm not sure what timestamp should be associated with data measured at time t1, but modified by a transforms from time t2. I assume I have not hit this bug because my tf transforms have been valid for the sensor measurement time. What is your scenario that the transforms are not available? Is it a fluke where you would like/tolerate the generation of an error message, or is it "standard operating procedure" for some reason?

svwilliams (approved):

ayrton04 (approved):

sjphilli:
Yeah, this only came up because of a bug in how my tfs were being stamped (in other code).

I think disallowing transforms at non-header stamps is a good idea. That would remove the t1/t2 dilemma, and I can't really think of any use cases that would be heavily depend on that behaviour.

PR #115, opened 2019-11-06
Predict jacobians per parameter block

efernandez:
This is a continuation of https://github.com/locusrobotics/fuse/pull/109 that provides the following:
1. Compute the jacobians of the motion model `predict` free function
    and a new cost function for the `unicycle_2d` state, that is equivalent to
    the cost functor, following the documentation regarding the jacobians
    layout: http://ceres-solver.org/nnls_modeling.html#costfunction . That is,
    the jacobians are computed per parameter blocks.

2. Add a unit test for the new cost function. The gradient checker is disabled because
    if we call `Probe` on the gradient checker the check passes successfully,
    but the probe results produces at segfault when it's destroyed at the
    end of the `TEST` scope.
    
    This also happens if we create an dummy cost function and probe the
    gradient checker on it, taking the gradient checker test from ceres
    source code. This suggest either an issue in the gradient checker itself
    or likely some compile or link issue.

    Fortunately, a similar check is done with automatic differentiation, w/o any issues.
    And it's still possible to enable the gradient checker check (currently commented out)
    locally, and confirm that it also passes.

3. Update previous code, specially to convert between `Eigen` matrices that
    store the full jacobian and ceres jacobians layout.

After this PR we should be able to merge the `predict` functions and the cost function and functor following http://ceres-solver.org/interfacing_with_autodiff.html . This could be discussed as part of this PR, but IMHO it'd make the diff too large and make the review harder. But there certainly a few things that are still in WIP status.

### DIFF
--- a/fuse_core/include/fuse_core/eigen.h
+++ b/fuse_core/include/fuse_core/eigen.h
@@ -63,6 +63,9 @@ using Matrix7d = Eigen::Matrix<double, 7, 7, Eigen::RowMajor>;
 using Matrix8d = Eigen::Matrix<double, 8, 8, Eigen::RowMajor>;
 using Matrix9d = Eigen::Matrix<double, 9, 9, Eigen::RowMajor>;
 
+template <typename Scalar, int RowsAtCompileTime, int ColsAtCompileTime>
+using Matrix = Eigen::Matrix<Scalar, RowsAtCompileTime, ColsAtCompileTime, Eigen::RowMajor>;
+
 }  // namespace fuse_core
 
 #endif  // FUSE_CORE_EIGEN_H
--- a/fuse_models/include/fuse_models/unicycle_2d_predict.h
+++ b/fuse_models/include/fuse_models/unicycle_2d_predict.h
@@ -39,6 +39,8 @@
 #include <fuse_core/eigen.h>
 #include <tf2_2d/tf2_2d.h>
 
+#include <array>
+
 
 namespace fuse_models
 {
@@ -54,14 +56,14 @@ namespace fuse_models
  * @param[in] acc_linear1_x - First X acceleration
  * @param[in] acc_linear1_y - First Y acceleration
  * @param[in] dt - The time delta across which to predict the state
- * @param[in] position2_x - Second X position
- * @param[in] position2_y - Second Y position
- * @param[in] yaw2 - Second orientation
- * @param[in] vel_linear2_x - Second X velocity
- * @param[in] vel_linear2_y - Second Y velocity
- * @param[in] vel_yaw2 - Second yaw velocity
- * @param[in] acc_linear2_x - Second X acceleration
- * @param[in] acc_linear2_y - Second Y acceleration
+ * @param[out] position2_x - Second X position
+ * @param[out] position2_y - Second Y position
+ * @param[out] yaw2 - Second orientation
+ * @param[out] vel_linear2_x - Second X velocity
+ * @param[out] vel_linear2_y - Second Y velocity
+ * @param[out] vel_yaw2 - Second yaw velocity
+ * @param[out] acc_linear2_x - Second X acceleration
+ * @param[out] acc_linear2_y - Second Y acceleration
  */
 template<typename T>
 inline void predict(
@@ -101,18 +103,6 @@ inline void predict(
   fuse_core::wrapAngle2D(yaw2);
 }
 
-enum StateIndex : uint8_t
-{
-  X,
-  Y,
-  YAW,
-  V_X,
-  V_Y,
-  V_YAW,
-  A_X,
-  A_Y
-};
-
 /**
  * @brief Given a state and time delta, predicts a new state
  * @param[in] position1_x - First X position
@@ -124,18 +114,16 @@ enum StateIndex : uint8_t
  * @param[in] acc_linear1_x - First X acceleration
  * @param[in] acc_linear1_y - First Y acceleration
  * @param[in] dt - The time delta across which to predict the state
- * @param[in] position2_x - Second X position
- * @param[in] position2_y - Second Y position
- * @param[in] yaw2 - Second orientation
- * @param[in] vel_linear2_x - Second X velocity
- * @param[in] vel_linear2_y - Second Y velocity
- * @param[in] vel_yaw2 - Second yaw velocity
- * @param[in] acc_linear2_x - Second X acceleration
- * @param[in] acc_linear2_y - Second Y acceleration
- * @param[in] jacobians - Jacobians wrt the state
+ * @param[out] position2_x - Second X position
+ * @param[out] position2_y - Second Y position
+ * @param[out] yaw2 - Second orientation
+ * @param[out] vel_linear2_x - Second X velocity
+ * @param[out] vel_linear2_y - Second Y velocity
+ * @param[out] vel_yaw2 - Second yaw velocity
+ * @param[out] acc_linear2_x - Second X acceleration
+ * @param[out] acc_linear2_y - Second Y acceleration
+ * @param[out] jacobians - Jacobians wrt the state
  */
-// TODO(efernandez) fuse this with the templated predict function so we can use the analytic Jacobian in the cost
-// function. See http://ceres-solver.org/analytical_derivatives.html
 inline void predict(
   const double position1_x,
   const double position1_y,
@@ -154,10 +142,10 @@ inline void predict(
   double& vel_yaw2,
   double& acc_linear2_x,
   double& acc_linear2_y,
-  // TODO(efernandez) double** jacobians didn't play well with Eigen::Map<>, at least directly
-  double* jacobians)
+  double** jacobians)
 {
-  const double sy = ceres::sin(yaw1);
+  // There are better models for this projection, but this matches the one used by r_l.
+  const double sy = ceres::sin(yaw1);  // Should probably be sin((yaw1 + yaw2) / 2), but r_l uses this model
   const double cy = ceres::cos(yaw1);
 
   const double half_dt2 = 0.5 * dt * dt;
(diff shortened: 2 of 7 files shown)


fuse_core/include/fuse_core/eigen.h
```diff
@@ -63,6 +63,9 @@ using Matrix7d = Eigen::Matrix<double, 7, 7, Eigen::RowMajor>;
 using Matrix8d = Eigen::Matrix<double, 8, 8, Eigen::RowMajor>;
 using Matrix9d = Eigen::Matrix<double, 9, 9, Eigen::RowMajor>;
 
+template <typename Scalar, int RowsAtCompileTime, int ColsAtCompileTime>
```
efernandez, line 66: This allows to declare row-major matrices with `fuse_core::Matrix<S, R, C>`, which is quite convenient when setting the jacobians for each parameter block.

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -54,14 +56,14 @@ namespace fuse_models
...
  * @param[in] dt - The time delta across which to predict the state
- * @param[in] position2_x - Second X position
- * @param[in] position2_y - Second Y position
- * @param[in] yaw2 - Second orientation
- * @param[in] vel_linear2_x - Second X velocity
- * @param[in] vel_linear2_y - Second Y velocity
- * @param[in] vel_yaw2 - Second yaw velocity
- * @param[in] acc_linear2_x - Second X acceleration
- * @param[in] acc_linear2_y - Second Y acceleration
+ * @param[out] position2_x - Second X position
```
efernandez, line 59: The doxygen was wrong. These are `out`put arguments.

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -101,18 +103,6 @@ inline void predict(
   fuse_core::wrapAngle2D(yaw2);
 }
 
-enum StateIndex : uint8_t
```
efernandez, line 104: Now I use the comma initializer to set the jacobians, so I don't need this `enum`.

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -124,18 +114,16 @@ enum StateIndex : uint8_t
...
- * @param[in] position2_x - Second X position
- * @param[in] position2_y - Second Y position
- * @param[in] yaw2 - Second orientation
- * @param[in] vel_linear2_x - Second X velocity
- * @param[in] vel_linear2_y - Second Y velocity
- * @param[in] vel_yaw2 - Second yaw velocity
- * @param[in] acc_linear2_x - Second X acceleration
- * @param[in] acc_linear2_y - Second Y acceleration
- * @param[in] jacobians - Jacobians wrt the state
+ * @param[out] position2_x - Second X position
```
efernandez, line 117: The doxygen was wrong. These are `out`put arguments.

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -154,10 +142,10 @@ inline void predict(
   double& vel_yaw2,
   double& acc_linear2_x,
   double& acc_linear2_y,
-  // TODO(efernandez) double** jacobians didn't play well with Eigen::Map<>, at least directly
-  double* jacobians)
+  double** jacobians)
```
efernandez, line 145: This is now equivalent to the `Evaluate` argument for the `jacobians` in a `ceres::CostFunction`, so it's easier to interact with this from the cost function. The cost function might (and actually does) provide more jacobians than the ones set here. Indeed, here only the first 5 parameter blocks are set, the other 5 are set in the cost function, as well as updating the ones set here.

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -154,10 +142,10 @@ inline void predict(
...
 {
-  const double sy = ceres::sin(yaw1);
+  // There are better models for this projection, but this matches the one used by r_l.
```
efernandez, line 147: This is simply the same comment as in the `predict` function that doesn't compute the `jacobians`, in preparation to remove the other one and only have this one in a future PR.

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -178,26 +166,70 @@ inline void predict(
...
 
-  if (!jacobians)
+  if (jacobians)
   {
-    return;
-  }
+    // Jacobian wrt position1
+    if (jacobians[0])
+    {
+      Eigen::Map<fuse_core::Matrix<double, 8, 2>> jacobian(jacobians[0]);
```
efernandez, line 174: I tried using `Eigen::ColMajor` to make the comma intializer be in two lines, but that didn't work.

If you want, I could add some comments to make it easier to know that each row and column represent in the comma initializer below.

Note I'm also trying to provide the most efficient implementation. Another option would be to set the `jacobians` directly using `jacobians[R][C]` instead of using an `Eigen::Map` and the comma initializer. I guess that might be a little faster. Not sure if it'd be cleaner/easier to read. Thoughts?

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -281,6 +313,23 @@ inline void predict(
   double acc_linear_x_pred {};
   double acc_linear_y_pred {};
 
+  // fuse_core::Matrix8d is Eigen::RowMajor, so we cannot use pointers to the columns where each parameter block starts.
```
efernandez, line 316: With `RowMajor` matrices this is the only way to make things work with ceres jacobians layout. It's a little bit ugly I have to do this here because the concepts of parameter blocks comes from the `ceres::Problem` definition, but it impacts the `predict` function contract. I don't see what else I can do, but I could probably encapsulate this code in an utility function (or maybe a class) that does this conversion in both directions.

Another option would be to use `ColMajor` matrices for the full jacobian, but `ceres` still expects `RowMajor` order for the sub-matrices that represent each parameter block jacobian, so we might still run into some issues/ugliness/inefficiencies.

If we keep the full jacobian a `RowMajor` (which I think it's what we should do for now, so we don't break the API), the conversion could also be done by changing to `ColMajor` first. I didn't try that though, and I'm not 100% it'd make things easier or more efficient.

fuse_models/include/fuse_models/unicycle_2d_predict.h
```diff
@@ -299,7 +348,9 @@ inline void predict(
     vel_yaw2,
     acc_linear_x_pred,
     acc_linear_y_pred,
-    jacobian.data());
+    jacobians.data());
+
+  jacobian << J[0], J[1], J[2], J[3], J[4];
```
efernandez, line 353: This is the trivial conversion back to the `RowMajor` full jacobian.

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -0,0 +1,285 @@
...
+ * where, the matrix A is fixed, the state variables are provided at two discrete time steps, and proj is a function
+ * that projects the state variables from time t1 to time t2. In case the user is interested in implementing a cost
+ * function of the form
+ *
+ *   cost(X) = (X - mu)^T S^{-1} (X - mu)
+ *
+ * where, mu is a vector and S is a covariance matrix, then, A = S^{-1/2}, i.e the matrix A is the square root
+ * information matrix (the inverse of the covariance).
+ */
+class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2, 1, 2, 2, 1, 2, 1, 2>
```
efernandez, line 85: This is certainly duplicating things from the cost functor. Either on this PR or in a future one (which I prefer), we should follow http://ceres-solver.org/interfacing_with_autodiff.html in order to remove the code duplication but still allow to use the analytic and automatic differentiation (and even numeric differentiation, required by `ceres::GradientChecker`), which is important for unit testing purposes.

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -0,0 +1,285 @@
...
+    fuse_core::wrapAngle2D(residuals[2]);
+
+    // Scale the residuals by the square root information matrix to account for
+    // the measurement uncertainty.
+    Eigen::Map<fuse_core::Vector8d> residuals_map(residuals);
+    residuals_map.applyOnTheLeft(A_);
+
+    if (jacobians)
+    {
+      // It might be possible to simplify the code below implementing something like this but using compile-time
```
efernandez, line 168: This is a TODO comment. Something that could be explored, although it might require some helper `struct`s to implement the template recursion. Not sure if it'd be cleaner at the end. It'd be definitely cleaner here, but it'd require that template recursion magic somewhere, and it'd probably require access to `A_`, at least.

fuse_models/src/unicycle_2d_state_kinematic_constraint.cpp
```diff
@@ -103,8 +102,16 @@ void Unicycle2DStateKinematicConstraint::print(std::ostream& stream) const
 
 ceres::CostFunction* Unicycle2DStateKinematicConstraint::costFunction() const
 {
-  return new ceres::AutoDiffCostFunction<Unicycle2DStateCostFunctor, 8, 2, 1, 2, 1, 2, 2, 1, 2, 1, 2>(
-    new Unicycle2DStateCostFunctor(dt_, sqrt_information_));
+  // Here we return a cost function that computes the analytic derivatives/jacobians, but we could use automatic
```
efernandez, line 105: I don't know if there's an easy way to allow to switch from analytic to automatic differentiation at runtime, but I guess we don't really need that. I might be convenient though for implementing the unit test that compares the jacobians with one and the other differentiation method, for validation purpose of the analytic ones.

fuse_models/test/test_unicycle_2d_predict.cpp
```diff
@@ -344,7 +345,19 @@ TEST(Predict, predictJacobians)
   double acc_linear2_x = 0.0;
   double acc_linear2_y = 0.0;
 
-  fuse_core::Matrix8d J;
+  const std::array<size_t, 5> block_sizes = {2, 1, 2, 1, 2};
```
efernandez, line 348: Updating so the full jacobian matrix is properly converted into the layout `ceres` expects. This is a conversion similar to the one mentioned before in on of the `predict` free function. In this case the `num_parameter_blocks` is a compile time constant, so we can use an `std::array` to store the jacobians per parameter block. In the `predict` method is the same, although there we use some `static` variables.

If we encapsulate the conversion somewhere else, I think it'd have to be a templated class, so we can indicate the parameter block sizes. Thoughts?

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+  const double* parameters[] =
+  {
+    position1, yaw1, vel_linear1, vel_yaw1, acc_linear1,
+    position2, yaw2, vel_linear2, vel_yaw2, acc_linear2
+  };
+
+  fuse_core::Vector8d residuals;
+
+  const auto& block_sizes = cost_function.parameter_block_sizes();
```
efernandez, line 80: Again, here we convert from `ceres` jacobian layout to a full jacobian. But in this case we take the parameter block sizes from the cost function. This is NOT available at compile time, so we must use an `std::vector` instead of an `std::array`. If we create a conversion util, it should ideally support both the compile (array) and runtime (vector) cases. I believe this is an important point that should be considered. Maybe this could also be done at compile time if we access the parameter block sizes from the class, rather than an instantiated object. :thinking:

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
```diff
@@ -0,0 +1,141 @@
...
+  // are not zero for position2.x = -4.389e-16 and yaw2 = -2.776e-16
+  EXPECT_MATRIX_NEAR(fuse_core::Vector8d::Zero(), residuals, 1e-15);
+
+  // Check jacobians are correct using a gradient checker
+  ceres::NumericDiffOptions numeric_diff_options;
+  ceres::GradientChecker gradient_checker(&cost_function, NULL, numeric_diff_options);
+
+  // We cannot use std::numeric_limits<double>::epsilon() tolerance because the worst relative error is 5.26356e-10
+  ceres::GradientChecker::ProbeResults probe_results;
+  // TODO(efernandez) probe_results segfaults when it's destroyed at the end of this TEST function, but Probe actually
```
efernandez, line 106: This is really unfortunate. It'd be great if you could give it a try in your system. I usually add a print a the end of the `TEST` to see if the segfault happens after, when objects are being destroyed.

My gut feeling is that this test should link against something else from `ceres`. I built and ran the `gradient_checker_test` from `ceres` and it didn't segfault, but then I took almost the same code from that test and tried it inside `fuse_models` and it crashed with the segfault described in the comment.

efernandez (commented):

fuse_models/include/fuse_models/unicycle_2d_predict.h
svwilliams: I'm all for using the most efficient method. If you happen to do some timing tests and that shows switching to direct initialization is faster, then have at it. Otherwise this seems completely fine.

And if you want to label the jacobian terms, that's great. But it should also be fairly clear what they are. And I actually prefer the 8-row version because that is the correct shape of the jacobian block.

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
svwilliams: I'll take a look at the gradient checker at some point, but I don't want to hold up this PR for it.

fuse_models/src/unicycle_2d_state_kinematic_constraint.cpp
svwilliams: I'm somewhat curious to see the performance difference between the analytic derivatives and the automatic derivatives. As long as the analytic derivatives are faster, then I see no particular need to switch back and forth.

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
svwilliams: I thought this was going to be easy...until I tried to write it up. Leaving this as a TODO item seems reasonable. It won't actually improve efficiency any; just reduce the amount of code duplication.

fuse_models/include/fuse_models/unicycle_2d_predict.h
svwilliams: This seems fine. I can't think of a way to convert without doing a copy. If a profiler ever says this consumes an important amount of CPU time, we can revisit.

fuse_core/include/fuse_core/eigen.h
svwilliams: Thanks. I've been meaning to do something like this.

svwilliams (approved):

fuse_models/include/fuse_models/unicycle_2d_predict.h
ayrton04: Unrelated: given that this package is no longer `fuse_rl`, I wonder if we care about trying to replicate how `r_l` behaves.

ayrton04 (approved):

fuse_models/include/fuse_models/unicycle_2d_predict.h
svwilliams: I'm going to leave that decision to you, Tom. But in a separate PR.

PR #116, opened 2019-11-11
Wait for reset service existence

efernandez:
There are some cases where the `reset` service might not be available yet if the `set_pose` is set very early when the node is launched. This waits for the `reset` service to be available in the `onInit` method, so it's guaranteed to be available later when it's called.

### DIFF
--- a/fuse_models/src/unicycle_2d_ignition.cpp
+++ b/fuse_models/src/unicycle_2d_ignition.cpp
@@ -230,6 +230,12 @@ void Unicycle2DIgnition::process(const geometry_msgs::PoseWithCovarianceStamped&
   // Tell the optimizer to reset before providing the initial state
   if (!params_.reset_service.empty())
   {
+    // Wait for the reset service
+    while (!reset_client_.waitForExistence(ros::Duration(10.0)) && ros::ok())
+    {
+      ROS_WARN_STREAM("Waiting for '" << reset_client_.getService() << "' service to become avaiable.");
+    }
+
     auto srv = std_srvs::Empty();
     if (!reset_client_.call(srv))
     {


fuse_models/src/unicycle_2d_ignition.cpp
```diff
@@ -88,6 +88,11 @@ void Unicycle2DIgnition::onInit()
   if (!params_.reset_service.empty())
   {
     reset_client_ = node_handle_.serviceClient<std_srvs::Empty>(ros::names::resolve(params_.reset_service));
+
+    while (!reset_client_.waitForExistence(ros::Duration(30.0)) && ros::ok())
```
efernandez: The `30.0`s timeout dictates the `ROS_WARN_STREAM` print rate. Maybe it could be set to a lower value like `10.0`.

The check for `ros::ok()` is because `waitForExistence` could also `return false` if the node has been shutdown: http://docs.ros.org/indigo/api/roscpp/html/classros_1_1ServiceClient.html#a530e3f1d55cf50ab00b8e5bbd9e8230a

efernandez: This actually doesn't work because it's called before the `reset` service is advertised in the `FixedLagSmoother` constructor, and happens in the base class `Optimizer` constructor, so it creates some sort of "deadlock". I'm trying other places, but it looks like it's not possible to guarantee the service is available yet in the init/start methods. I guess worst case I could add a check in the `process`  method.

efernandez: I've updated the PR to wait for the `reset` service in the `process` method, right before the service client is called.

svwilliams: Yeah. I have a future task for myself to rework the optimizer construction sequence. I feel like we should completely initialize the derived optimizer before loading the plugins. But with the current inheritance scheme, that is not really possible.

fuse_models/src/unicycle_2d_ignition.cpp
```diff
@@ -230,6 +230,12 @@ void Unicycle2DIgnition::process(const geometry_msgs::PoseWithCovarianceStamped&
   // Tell the optimizer to reset before providing the initial state
   if (!params_.reset_service.empty())
   {
+    // Wait for the reset service
+    while (!reset_client_.waitForExistence(ros::Duration(30.0)) && ros::ok())
```
svwilliams: 10s does seem less extreme, if you want to make that change.

svwilliams (approved):

fuse_models/src/unicycle_2d_ignition.cpp
efernandez: Changed to `10.0`s

ayrton04 (approved):

PR #118, opened 2019-11-19
Add fuse_loss pkg with plugin-based loss functions

efernandez:
* Add `fuse_loss` pkg with basic ceres loss functions wrapped into plugins
* Update `fuse_core::Constraint` to allow setting its loss function with a `loss` accessor and mutator
* Add ROS parameters to set the loss function of all constraints set by all `fuse_models`
* Add unit test for `fuse_core::Loss` with a dummy `ExampleLoss` child class
* Update constraint and graph optimization unit tests adding loss functions
* Add unit test for basic loss functions provided in `fuse_loss`
* Add singleton `LossLoader` class to load loss functions with `pluginlib` `ClassLoader` during paremeter loading from different `fuse_models`, so it outlive the objects it instantiate
* Update graph and transaction deserializers so they also load all `fuse_core::Loss` plugin declared in the current workspace
* Don't let ceres take ownership of the loss functions, so they don't get destroyed with the problem because they shouldn't be remove until the constraint and loss that handles is destroyed
* Add missed `fuse_variables` depend missed in one fuse pkg

I tested this with Ceres 2.0 / master on top of https://github.com/locusrobotics/fuse/pull/117:
* CPU% and MEM% usage look similar. I added different types of loss functions to multiple different constraints of three odometry sources and the motion model
* I used `fuse_echo` to validate the transactions have loss functions in the added constraints, which I think it's a sufficient test for now
* I've noticed that when I `Ctrl+C` the launch file from where I run fuse (I'm running the `fixed_lag_smoother`), ROS escalates to SIGTERM because it takes longer to stop the node. I suspect this is because of the way I'm loading unique instances of loss functions with `pluginlib`. I'm implicitly casting the `UniquePtr` from `pluginlib` into an `std::shared_ptr`, which is allowed in C++ standard, but the `UniquePtr` has a non-trivial destructor that does some extra work that I think I'm losing when I cast it into an `std::shared_ptr`. This is a relatively minor issue, but I can look into fixing it.


fuse_constraints/src/marginalize_variables.cpp
```diff
@@ -331,7 +331,6 @@ LinearTerm linearize(
     double squared_norm = result.b.squaredNorm();
     double rho[3];
     loss_function->Evaluate(squared_norm, rho);
-    delete loss_function;
```
efernandez, line 334: The `loss_function` is now handled by an `std::shared_ptr<Loss>` object

fuse_core/include/fuse_core/constraint.h
```diff
@@ -262,29 +263,38 @@ class Constraint
...
+    loss_ = loss;
+  }
+
+  /**
+   * @brief Read-only access to the Ceres loss function.
+   *
+   * @return A base pointer to an instance of a derived ceres::LossFunction.
    */
-  virtual ceres::LossFunction* lossFunction() const
+  ceres::LossFunction* lossFunction() const
```
efernandez, line 292: I didn't change the name of this method so I didn't have to update the places where it's used. For that reason the name I used for the class that wraps it up is `fuse_core::Loss` instead of `fuse_core::LossFunction`, and similarly, the attribute here is `loss_` instead of `loss_function_`, and same for the `loss()` accessor and mutator. What do you think about this names?

fuse_core/include/fuse_core/loss_loader.h
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * @brief Constructor
+   */
+  LossLoader()
+    : loss_loader_("fuse_core", "fuse_core::Loss")
+  {
+  }
+
+private:
+  pluginlib::ClassLoader<fuse_core::Loss> loss_loader_;  //!< Pluginlib class loader for Loss types
```
efernandez, line 94: I need a singleton for this because I need to guarantee it outlives the objects it creates. Otherwise a `SEVERE WARNING` is printed when it gets destroyed before the objects it has instantiated.

I use this from the `initialize` method in the `Loss` children, from a helper free function that loads the loss function parameters from the server. So the other option would be to have one loader per sensor/motion model, which doesn't look better than this approach to me. Thoughts?

fuse_graphs/src/hash_graph.cpp
```diff
@@ -53,6 +53,8 @@ namespace fuse_graphs
 HashGraph::HashGraph(const HashGraphParams& params) :
   problem_options_(params.problem_options)
 {
+  // We don't want Ceres to take ownership of the loss functions
+  problem_options_.loss_function_ownership = ceres::Ownership::DO_NOT_TAKE_OWNERSHIP;
```
efernandez: This is important because we recreate the problem every time the graph is optimized or we get the covariance from the solver. And even if we didn't recreate the problem, now we need to guarantee the `ceres::LossFunction`s don't get destroyed until the `fuse_core::Loss` wrapper that handles it is destroyed, which happens when the `fuse_core::Constraint` that has that loss function is not used anymore, e.g. marginalized out.

fuse_graphs/test/example_loss.h
```diff
@@ -0,0 +1,91 @@
...
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_GRAPHS_TEST_EXAMPLE_LOSS_H  // NOLINT{build/header_guard}
+#define FUSE_GRAPHS_TEST_EXAMPLE_LOSS_H  // NOLINT{build/header_guard}
```
efernandez, line 35: This is the same as the `ExampleLoss` class in `fuse_core/test`. Not should how to share them. Should I export the `fuse_core` one so I can use it in `fuse_graphs`?

fuse_loss/fuse_plugins.xml
```diff
@@ -0,0 +1,72 @@
+<library path="lib/libfuse_loss">
+  <class type="fuse_loss::TrivialLoss" base_class_type="fuse_core::Loss">
+    <description>
+    This corresponds to no robustification. It is not normally necessary to use this, as setting not loss function
```
efernandez, line 4: I summarized what I believe it's the important information from the ceres documentation, mainly from the source code comments: https://github.com/ceres-solver/ceres-solver/blob/master/include/ceres/loss_function.h

fuse_loss/include/fuse_loss/arctan_loss.h
```diff
@@ -0,0 +1,127 @@
...
+{
+
+/**
+ * @brief The ArctanLoss loss function.
+ *
+ * This class encapsulates the ceres::ArctanLoss class, adding the ability to serialize it and load it dynamically.
+ *
+ * See the Ceres documentation for more details. http://ceres-solver.org/nnls_modeling.html#lossfunction
+ */
+class ArctanLoss : public fuse_core::Loss
```
efernandez, line 57: This and the other `Loss` children classes are very thin wrappers on top of the corresponding `ceres::LossFunction`. I refrained from creating a pre-processor macro to instantiate them because I think that would have been ugly, especially because the `TolerantLoss` function has two arguments instead of one, which is what most loss functions have. Do you think there's a better approach? TBH I don't expect the loss functions to change or the number of them grow, so I think this should be fine. However, it's true there's a lot of code duplication.

fuse_loss/include/fuse_loss/arctan_loss.h
```diff
@@ -0,0 +1,127 @@
...
+   * @param[out] stream The stream to write to. Defaults to stdout.
+   */
+  void print(std::ostream& stream = std::cout) const override;
+
+  /**
+   * @brief Parameter 'a' accessor.
+   *
+   * @return Parameter 'a'.
+   */
+  double a() const
```
efernandez, line 107: I added this accessor and mutator below for testing purposes

fuse_loss/include/fuse_loss/tolerant_loss.h
```diff
@@ -0,0 +1,150 @@
...
+   * @param[in] b Parameter 'b'.
+   */
+  void b(const double b)
+  {
+    b_ = b;
+  }
+
+private:
+  double a_{ 1.0 };  //<! TolerantLoss parameter 'a'. See Ceres documentation for more details
+  double b_{ 0.1 };  //<! TolerantLoss parameter 'b'. See Ceres documentation for more details
```
efernandez, line 145: Not sure what the default value for `b_` should be

fuse_loss/test/test_huber_loss.cpp
```diff
@@ -0,0 +1,162 @@
...
+ */
+#include <fuse_core/serialization.h>
+#include <fuse_loss/huber_loss.h>
+
+#include <ceres/autodiff_cost_function.h>
+#include <ceres/problem.h>
+#include <ceres/solver.h>
+#include <gtest/gtest.h>
+
+TEST(HuberLoss, Constructor)
```
efernandez, line 42: I just realized I only create a unit test for the `HuberLoss`, and not for the others. Honestly, the unit test for the others would be almost the same, so I'm not sure if it's worth having it or not. If you think it's good idea or you don't have a strong opinion, I can create them. I'd basically copy this and create the others changing the loss function used and updating the parameters they take, as well as the values, so the optimization takes makes sense.

These classes aren't really giving a lot more functionality than the `ceres::LossFunction`s, so I probably shouldn't be testing much here. But I still thing the optimization test below is a good thing to have, either on this pkg or in another one inside the `fuse` repo.

Thoughts?

fuse_models/include/fuse_models/parameters/parameter_base.h
```diff
@@ -93,6 +94,22 @@ inline std::vector<size_t> loadSensorConfig(const ros::NodeHandle& nh, const std
...
+    return {};
+  }
+
+  std::string loss_type;
+  getParamRequired(nh, name + "/type", loss_type);
+
+  auto loss = fuse_core::createUniqueLoss(loss_type);
+  loss->initialize(nh.resolveName(name));
+
+  return loss;
```
efernandez, line 110: This is important:
* `fuse_core::createUniqueLoss` returns a `pluginlib::UniquePtr<T>` where `T` is `fuse_core::Loss`: https://github.com/ros/pluginlib/blob/melodic-devel/pluginlib/include/pluginlib/class_loader.hpp#L57
* Therefore, the `loss` variable would be of type `pluginlib::UniquePtr<T>`
* But the `return` type of `loadLossConfig` is `fuse_core::Loss::SharedPtr`, i.e. an `std::shared_ptr<T>` pointer
* `pluginlib::UniquePtr<T>` is actually an alias of `class_loader::ClassLoader::UniquePtr<T>`: `pluginlib::UniquePtr<T>`, which is defined as an `std::unique_ptr<T, class_loader::DeleterType<T>>;`
* The `return` statement here does an implicit cast from the `std::unique_ptr` to the `std::shared_ptr`
* I'm not sure if that passes the non-default destructor `class_loader::DeleterType<T>`
* In case the non-default destructor is not passed to the `std::shared_ptr`, I wonder if this is could explain that the node takes longer to shutdown and ROS escalates to `SIGTERM`. No crash happens though.

BTW the `class_loader::ClassLoader` class has a method to create an instance handle by an `std::shared_ptr`: https://github.com/ros/class_loader/blob/melodic-devel/include/class_loader/class_loader.hpp#L127, but curiously, the `pluginlib::ClassLoader` only has a method for an `std::unique_ptr`: https://github.com/ros/pluginlib/blob/melodic-devel/pluginlib/include/pluginlib/class_loader.hpp#L128, or for `boost::shared_ptr`: https://github.com/ros/pluginlib/blob/melodic-devel/pluginlib/include/pluginlib/class_loader.hpp#L109 :confused:

fuse_optimizers/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ set(build_depends
   fuse_constraints
   fuse_core
   fuse_graphs
+  fuse_variables
```
efernandez, line 9: This could be on another PR. I simply noticed that `fuse_optimizers` uses `fuse_variables` here: https://github.com/locusrobotics/fuse/blob/devel/fuse_optimizers/src/variable_stamp_index.cpp#L133

It's only part of the implementation though. It's not in the header file.

I can move this to another PR if you want. It got into this commit when I was rebasing all the commits I did during the development of this feature.

efernandez (commented):
Added a bunch of comments to help with the review

efernandez:
I just updated this PR with `roslint` fixes and a bug in the `ExampleLoss` class I used in the unit tests in `fuse_graphs`. All tests pass for me locally, but I'm actually testing with this PR on top of #117 because I have Ceres 2.0 installed in my system atm. So maybe it's something related to that, but I doubt it.

@svwilliams Could you paste here the checks that are failing, please?

fuse_models/include/fuse_models/parameters/parameter_base.h
efernandez: According to this test snippet, the custom deleter is passed to the `std::shared_ptr`, so in principle everything should be fine: https://wandbox.org/permlink/A8XqXg6x6a2bRvPY

svwilliams:
It appears to be a bunch of roslint checks
```
Scanning dependencies of target roslint_fuse_loss
test_rosdistro_fuse_PR-118/package/fuse_loss/src/arctan_loss.cpp:48:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/src/arctan_loss.cpp
test_rosdistro_fuse_PR-118/package/fuse_loss/src/cauchy_loss.cpp:48:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/src/cauchy_loss.cpp
test_rosdistro_fuse_PR-118/package/fuse_loss/src/huber_loss.cpp:48:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/src/huber_loss.cpp
test_rosdistro_fuse_PR-118/package/fuse_loss/src/softlone_loss.cpp:48:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/src/softlone_loss.cpp
test_rosdistro_fuse_PR-118/package/fuse_loss/src/tolerant_loss.cpp:49:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/src/tolerant_loss.cpp
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/src/trivial_loss.cpp
test_rosdistro_fuse_PR-118/package/fuse_loss/src/tukey_loss.cpp:48:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/src/tukey_loss.cpp
test_rosdistro_fuse_PR-118/package/fuse_loss/test/test_huber_loss.cpp:60:  Single-parameter constructors should be marked explicit.  [runtime/explicit] [5]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/test/test_huber_loss.cpp
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/arctan_loss.h:60:  Constructors callable with one argument should be marked explicit.  [runtime/explicit] [5]
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/arctan_loss.h:74:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/arctan_loss.h
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/cauchy_loss.h:60:  Constructors callable with one argument should be marked explicit.  [runtime/explicit] [5]
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/cauchy_loss.h:74:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/cauchy_loss.h
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/huber_loss.h:60:  Constructors callable with one argument should be marked explicit.  [runtime/explicit] [5]
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/huber_loss.h:74:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/huber_loss.h
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/softlone_loss.h:60:  Constructors callable with one argument should be marked explicit.  [runtime/explicit] [5]
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/softlone_loss.h:74:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/softlone_loss.h
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/tolerant_loss.h:61:  Constructors callable with one argument should be marked explicit.  [runtime/explicit] [5]
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/tolerant_loss.h:75:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/tolerant_loss.h
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/trivial_loss.h:72:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/trivial_loss.h
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/tukey_loss.h:60:  Constructors callable with one argument should be marked explicit.  [runtime/explicit] [5]
test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/tukey_loss.h:74:  Add #include <string> for string  [build/include_what_you_use] [4]
Done processing test_rosdistro_fuse_PR-118/package/fuse_loss/include/fuse_loss/tukey_loss.h
```

efernandez:
Oh, shoot! I forgot to test the new pkg `fuse_loss` in my lass pass with `make roslint` and `make run_tests`. Thanks. It should pass now.

paulbovbel:
FYI @svwilliams, if you'd like to make the unit test results visible for external contributors, you can add a travis test config.

A file like https://github.com/ros-perception/perception_pcl/blob/melodic-devel/.travis.yml should be enough.

efernandez:
Rebased on top of `devel` and updated `fuse_loss/CMakeLists.txt` to match the changes in https://github.com/locusrobotics/fuse/pull/119

fuse_constraints/src/absolute_orientation_3d_stamped_constraint.cpp
```diff
@@ -88,6 +88,12 @@ void AbsoluteOrientation3DStampedConstraint::print(std::ostream& stream) const
          << "  orientation variable: " << variables().at(0) << "\n"
          << "  mean: " << mean().transpose() << "\n"
          << "  sqrt_info: " << sqrtInformation() << "\n";
+
+  if (loss())
```
svwilliams, line 92: I keep thinking I should add some reusable base-class printing functionality. Maybe one day....

fuse_core/CMakeLists.txt
```diff
@@ -276,6 +276,26 @@ if(CATKIN_ENABLE_TESTING)
...
+      include
+      ${Boost_INCLUDE_DIRS}
+      ${catkin_INCLUDE_DIRS}
+      ${CERES_INCLUDE_DIRS}
+      ${CMAKE_CURRENT_SOURCE_DIR}
+      ${EIGEN3_INCLUDE_DIRS}
+  )
+  target_link_libraries(test_loss
+    ${PROJECT_NAME}
+  )
```
svwilliams, line 298: May or may not need to add:
```
  set_target_properties(test_loss
    PROPERTIES
      CXX_STANDARD 14
      CXX_STANDARD_REQUIRED YES
  )
```
I blindly added it to all of the unit tests because I didn't feel like tracking down exactly which ones needed it and which ones didn't.

fuse_core/include/fuse_core/loss.h
```diff
@@ -0,0 +1,307 @@
...
+  friend class boost::serialization::access;
+
+  /**
+   * @brief The Boost Serialize method that serializes all of the data members in to/out of the archive
+   *
+   * @param[in/out] archive - The archive object that holds the serialized class members
+   * @param[in] version - The version of the archive being read/written. Generally unused.
+   */
+  template<class Archive>
+  void serialize(Archive& /* archive */, const unsigned int /* version */)
```
svwilliams, line 294: Since the base class owns the `loss_function_` pointer, isn't it the base class's responsibility to serialize the loss_function_ pointer?

fuse_core/include/fuse_core/loss_loader.h
svwilliams: I was going to ask, since pluginlib itself uses some singletons internally. But this seems reasonable.

fuse_core/test/example_loss.h
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * @brief The Boost Serialize method that serializes all of the data members in to/out of the archive
+   *
+   * @param[in/out] archive - The archive object that holds the serialized class members
+   * @param[in] version - The version of the archive being read/written. Generally unused.
+   */
+  template<class Archive>
+  void serialize(Archive& archive, const unsigned int /* version */)
+  {
+    archive & boost::serialization::base_object<fuse_core::Loss>(*this);
```
svwilliams, line 85: The Ceres loss function pointer of the base class remains unserialized here.

fuse_graphs/test/example_loss.h
svwilliams: I have yet to find a way I like to handle this sort of thing. The code is littered with ExampleVariable and ExampleConstraint classes in several packages. And testing plugins is even worse. I say keep the copies for now. If I ever discover a way to handle this better, I can clean these up too.

fuse_graphs/src/hash_graph.cpp
svwilliams: That explains a question I had about returning the member pointer above.

From an interface standpoint, if the implementation of all Loss classes relies on this ownership setting, I feel like this setting needs to be enforced in the Graph base class...which is currently not practical. I'm making myself a ticket to fix that in the future. I'm curious if all of the objects should be set to DO_NOT_TAKE_OWNERSHIP? That might offer some performance advantages if the Constraint classes can create a single CostFunction object, instead of creating a new cost function object every time the graph is optimized.

fuse_core/include/fuse_core/loss.h
```diff
@@ -0,0 +1,307 @@
...
+   * @brief Print a human-readable description of the loss function to the provided stream.
+   *
+   * @param  stream The stream to write to. Defaults to stdout.
+   */
+  virtual void print(std::ostream& stream = std::cout) const = 0;
+
+  /**
+   * @brief Return a raw pointer to the ceres::LostFunction encapsulated
+   *
+   * The Ceres interface requires a raw pointer. Ceres will take ownership of the pointer and promises to properly
```
svwilliams, line 214: Further down you set the Ceres DO_NOT_TAKE_OWNERSHIP flag in the Graph. As such, this comment is wrong and needs to be updated.

fuse_core/include/fuse_core/loss.h
```diff
@@ -0,0 +1,307 @@
...
+ */
+class Loss
+{
+public:
+  SMART_PTR_ALIASES_ONLY(Loss);
+
+  /**
+   * @brief Constructor
+   *
+   * @param[in] loss_function - A base pointer to an instance of a derive ceres::LostFunction, or a nullptr (default)
```
svwilliams: Since this is a raw pointer, can you specify if we are transferring ownership of the pointer to Loss class or not.
And if we are transferring ownership, shouldn't we delete the pointer in the destructor.
And if we are not, I'm curious how the lifetime of the pointer is managed by the sensor/motion model.

fuse_core/test/example_loss.h
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * @brief Dummy loss implementation for testing
+ */
+class ExampleLoss : public fuse_core::Loss
+{
+public:
+  FUSE_LOSS_DEFINITIONS(ExampleLoss);
+
+  explicit ExampleLoss(const double a = 1.0) : fuse_core::Loss(new ceres::HuberLoss(a)), a(a)
```
svwilliams: This signature seems to indicate the base class owns the loss function pointer. However, nothing is currently deleting the loss function pointer in the base class.

fuse_core/test/test_loss.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <gtest/gtest.h>
+
+
+TEST(Loss, Constructor)
+{
+  const double a{ 0.3 };
+  ExampleLoss loss(a);
+  ASSERT_EQ(a, loss.a);
+  ASSERT_NE(nullptr, loss.lossFunction());
+}
```
svwilliams, line 53: I don't suppose we can add some serialization tests here? Particularly that check the `ceres::LossFunction` pointer, because I do not think it is currently being serialized.

fuse_loss/fuse_plugins.xml
svwilliams: This is great....much better than what I would have done. :)

fuse_optimizers/CMakeLists.txt
svwilliams: This is fine. No need to make extra work for anyone.

fuse_core/include/fuse_core/constraint.h
```diff
@@ -262,29 +263,38 @@ class Constraint
...
+  {
+    return loss_;
+  }
+
+  /**
+   * @brief Set the constraint loss function
+   *
+   * @param[in] loss - The loss function
+   */
+  void loss(const Loss::SharedPtr& loss)
```
svwilliams: The Constraint class is definitely sharing ownership of the provided Loss object. This should be indicated by accepting the shared ptr by value:
```
void loss(Loss::SharedPtr loss)
{
  loss_ = std::move(loss);
}
```
My reference for all things shared_ptr: https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/

Honestly, I would much rather the Constraint class have exclusive ownership of the Loss object via a unique_ptr, but the complexities of pluginlib makes that somewhere between difficult to impossible.

fuse_core/include/fuse_core/constraint.h
svwilliams: I'm not in love with `Loss` as a class/function name, but I don't have any good suggestions either. So I guess leave it unless Tom has a brilliant alternative name.

fuse_core/include/fuse_core/constraint.h
```diff
@@ -262,29 +263,38 @@ class Constraint
...
-   * with standard least-squares optimization. By providing a different loss function, the penalty for the constraint's
-   * error can be altered.
-   *
-   * This is generally done to reduce the effect of outlier measurements that made it into the optimization problem.
-   * It is always better to remove outliers before they make it into the optimization problem, but no method is perfect.
-   * Using robust loss functions can significantly improve the results and stability of the solution in the presence
-   * of outlier measurements.
+   * @brief Read-only access to the loss function.
    *
    * The Ceres interface requires a raw pointer. Ceres will take ownership of the pointer and promises to properly
```
svwilliams: I'm not sure this description really fits the `loss()` function

fuse_core/include/fuse_core/constraint.h
svwilliams: I'm also curious if it would make sense to supply the loss function via the constructor and skip these accessor functions all together. The downside being that all derived constraint constructors would also need to be updated to accept a loss function.

fuse_core/include/fuse_core/graph_deserializer.h
```diff
@@ -89,6 +89,7 @@ class GraphDeserializer
 private:
   pluginlib::ClassLoader<fuse_core::Variable> variable_loader_;  //!< Pluginlib class loader for Variable types
   pluginlib::ClassLoader<fuse_core::Constraint> constraint_loader_;  //!< Pluginlib class loader for Constraint types
+  pluginlib::ClassLoader<fuse_core::Constraint> loss_loader_;  //!< Pluginlib class loader for Loss types
```
svwilliams: I assume this should be:
```suggestion
  pluginlib::ClassLoader<fuse_core::Loss> loss_loader_;  //!< Pluginlib class loader for Loss types
```

fuse_core/include/fuse_core/transaction_deserializer.h
```diff
@@ -89,6 +89,7 @@ class TransactionDeserializer
 private:
   pluginlib::ClassLoader<fuse_core::Variable> variable_loader_;  //!< Pluginlib class loader for Variable types
   pluginlib::ClassLoader<fuse_core::Constraint> constraint_loader_;  //!< Pluginlib class loader for Constraint types
+  pluginlib::ClassLoader<fuse_core::Constraint> loss_loader_;  //!< Pluginlib class loader for Loss types
```
svwilliams: Again, assuming this should be:
```suggestion
  pluginlib::ClassLoader<fuse_core::Loss> loss_loader_;  //!< Pluginlib class loader for Loss types
```

fuse_loss/include/fuse_loss/arctan_loss.h
svwilliams: This is how I would have implemented it as well. All of the code here is just boilerplate stuff; no "real work" is duplicated. And the fact that the various loss functions have only one parameter, and that parameter is called `a` is more coincidence than by design. The different loss functions could have easily had different parameter names; then the code duplication would look a lot more necessary.

fuse_models/include/fuse_models/parameters/parameter_base.h
svwilliams: Yeah. I've run into this unique ptr deleter function issue before. I'm sure there is a very logical and well-considered reason why shared_ptrs can hold a deleter function but unique_ptrs require an additional template argument, but it's really annoying.

fuse_loss/test/test_huber_loss.cpp
svwilliams: Given the "thin wrapper"-ness of the `fuse_core::Loss` class, testing just one of the classes seems reasonable. I would add an additional check to the serialization unit to verify the `ceres::LossFunction*` survives the serialize-deserialize process. As mentioned earlier, I don't think it currently does.

And the only other unit test I would like to see is an optimization of a graph where the final value is different when the loss function is used, versus when it is not. That would verify that the loss function is correctly making it all the way into the Ceres problem. That would belong in the fuse_graph package, I suppose. Here we are manually creating the problem, so we are not testing the real code path used by the optimizers.

fuse_loss/include/fuse_loss/arctan_loss.h
```diff
@@ -0,0 +1,129 @@
...
+  void serialize(Archive& archive, const unsigned int /* version */)
+  {
+    archive & boost::serialization::base_object<fuse_core::Loss>(*this);
+    archive & a_;
+  }
+};
+
+}  // namespace fuse_loss
+
+BOOST_CLASS_EXPORT_KEY(fuse_loss::ArctanLoss);
```
svwilliams, line 144: This macro is obviously included somewhere, but the "include what you use" principle would recommend that the boost header be included here as well. Similarly, the pluginlib macro header should be included in the cpp files.

svwilliams (commented):
Having reviewed (most) of the code, I have a few design questions/notes:
(0) I'd prefer the Constraint classes have exclusive ownership of the loss functions. However, the implementation details of pluginlib make that nearly impossible. I am begrudgingly accepting that holding a shared_ptr is the only workable solution.
(1) The `fuse_core::Loss` class interface deviates from the rest of the fuse design in that the returned Ceres object pointer is owned by this class and not by Ceres. In all other similar cases (variable parameterizations, constraint cost functions), the fuse classes generate a brand new ceres object and return a pointer to that object to Ceres, letting Ceres worry about lifetime management. The same _could_ be done here by making the `lossFunction()` method pure virtual in the base class and constructing a new `ceres::LossFunction` object within the derived implementation. But I do not want to dismiss the current design too hastily, as there is potentially some computation benefit to _not_ creating the same object over and over again. Perhaps I'll want to adopt this same pattern for the other classes as well.
(2) Holding a base class `ceres::LossFunction` pointer in the fuse `fuse_core::Loss` base class slightly limits the implementation details of derived classes. By holding the pointer in the base class you get to define a default implementation of `lossFunction()`, but I think that is about it. If you leave that to the derived class, you have the opportunity for the derived class to hold a stack-allocated member object, implementing the `lossFunction()` as `ceres::LossFunction* lossFunction() const { return &member_variable_; }` (where const-correctness may need to be adjusted to make Ceres happy). This has the added advantage of not needing to delete a pointer in the destructor (currently missing from the code), and not having to worry about who owns the raw pointer. Since the selection of the loss function is delegated to named fuse classes anyway (`fuse_loss::HuberLoss`, etc.), I don't think you lose any functionality.
(3) As written, I don't think the serialization system works. Nothing is currently serializing the `ceres::LossFunction*` pointer held in the `fuse_core::Loss` base class. Thus, when a `fuse_core::Loss` object is deserialized, that pointer is either going to be null or uninitialized gibberish. I believe Boost Serialization supports serializing an object via a base class pointer, provided all of the derived classes have been properly registered and serialization functions have been written for them. So, to get serialization working I think you need to write Boost Serialization methods for each type of the derived `ceres::LossFunction`, and then serialize the base class pointer in the `serialize()` method of the `fuse_core::Loss` base class (or, if you use stack-allocated objects as described in (2), the derived `ceres::LossFunction` member variable would need to be serialized by the derived `fuse_core::Loss` class). I believe Boost Serialization supports non-intrusive serialization functions, which would allow the serialization functions to be written without modifying the Ceres classes. Alternatively, if the derived `fuse_core::Loss` objects always return a new `ceres::LossFunction` object pointer as in (1), then the derived `fuse_core::Loss` classes would not contain a `ceres::LossFunction` pointer or object, and hence no serialization methods would be needed for the Ceres classes. That is appealing from a "less work required" point of view, but it does give up the possible computation advantage of not creating and deleting copies of the `ceres::LossFunction` classes every time the graph is optimized.
(4) You have previously expressed an interest in supporting the [`ceres::ComposedLoss`](http://ceres-solver.org/nnls_modeling.html#_CPPv2N5ceres12ComposedLossE) function. That may represent a challenge; I haven't thought through how that would work with the current `fuse_core::Loss` classes. You might be able to create a `fuse_loss::ComposedLoss` class that holds shared_ptrs to two child `fuse_core::Loss` objects. The `initialize()` method would then recurse to initialize the two children. Once the children have been created, a `ceres::ComposedLoss` object could be created using the `lossFunction()` methods of the children. Maybe? It at least seems plausible.
(5) I'm curious if it would make sense to supply the loss function via the `fuse_core::Constraint` constructor, instead of as an additional post-construction step. The downside being that all derived constraint constructors would also need to be updated to accept a loss function.
(6) I'm delegating approval of all class and parameter names to Tom.
(7) I have asked about the license preamble and not received any response. Feel free to switch out to your Clearpath preamble. That is actually included in other files already, so why not these.

fuse_core/include/fuse_core/constraint.h
efernandez: I initially call it `LossFunction`, but it's not too different from `Loss` and at the end I took the shorter one. I don't have a strong opinion on this, but I prefer `Loss` over `LossFunction`. I'm all ears for other naming suggestions. :ear:

fuse_core/include/fuse_core/loss.h
efernandez: True, I just gave that a try, but I relaized that in order to do that I need to provide Boost serialization support to `ceres::LossFunction`s, which I don't think it's possible, or at least it's not trivial.

I've removed the `loss_function_` attribute from the `Loss` base class and now it's the children responsibility to return a `ceres::LossFunction*` from the `lossFunction()` method, that now is pure virtual.

The ownership of the `ceres::LossFunction` has also been changed to `TAKE_OWNERSHIP` (Ceres default), although it could be changed to `DO_NOT_TAKE_OWNERSHIP`.

## Thoughts on supporting `DO_NOT_TAKE_OWNERSHIP`

First of all, this  could be done in the future, in another PR.

I believe the hardest part to do so would be the Boost serialization:
* I think it'd required a split `save` and `load`, so we can construct the `ceres::LossFunction` in the `load` method only.
* It'd be constructed based on the other members in the children classes (typically the `a` argument, which would be the only thing de/serialized).
* The children would have an `std::shared_ptr<ceres::LossFunction> loss_function_` (or `std::unique_ptr`) member and they'd return a RAW pointer to it with `.get()` in `lossFunction()`.

fuse_core/test/example_loss.h
efernandez: Changed. There's no need to serialized it after my last changes because now the children implement the `lossFunction()` pure virtual method returning a `ceres::LossFunction` pointer that is built and allocated in the heap, as I think it was the original intend. Note this could be changed in a future PR to make it more efficient.

Example implementation for the `HuberLoss` children:
``` c++
ceres::LossFunction* HuberLoss::lossFunction() const
{
  return new ceres::HuberLoss(a_);
}
```

fuse_core/include/fuse_core/transaction_deserializer.h
efernandez: Fixed.

When I tested it with `fuse_echo` it was printing the loss function values properly, so maybe this wasn't really needed. Either way, I think we can look at whether all these loaders are needed or not later.

fuse_graphs/src/hash_graph.cpp
efernandez: I think using `DO_NOT_TAKE_OWNERSHIP` would allow some optimizations gains. That could be done in a future PR.

I've done some changes that now make the loss functions behave as the constraints, so ownership is on the code that calls `lossFunction()`.

I also added this to the `Loss` base class:
``` c++
  static constexpr ceres::Ownership Ownership =
      ceres::Ownership::TAKE_OWNERSHIP;  //<! The ownership of the ceres::LossFunction* returned by lossFunction()
```

This way the code that uses it can check the ownership that's being used and act accordingly. I've updated all the code and tests to take this into account. This means that switching to `DO_NOT_TAKE_OWNERSHIP` only requires a change here, in principle.

fuse_graphs/test/example_loss.h
efernandez: :+1:

fuse_loss/fuse_plugins.xml
efernandez: :smiley:

fuse_loss/include/fuse_loss/arctan_loss.h
efernandez: :+1:

fuse_loss/test/test_huber_loss.cpp
efernandez: You were right. I just extended the test to check the `ceres::LossFunction*` survices the serialize-deserialize process.

I check:
1. It's not NULL, which it's not enough because it could get constructed again with default values, instead of the originals
2. It satisfies some invariantes for the inlier region
3. It satisfies some invariantes for the outlier region

There's already a unit test I added that exercise and checks the error changes depending on whether the loss functions are used or not. That test is actually in `fuse_loss`. Not sure if you want me to move it somewhere else. :thinking: Note that I updated some existing tests in both `fuse_constraints` and `fuse_graphs`. I think they already test what you're saying.

fuse_models/include/fuse_models/parameters/parameter_base.h
efernandez: Considering this good for now and marking this conversation as resolved.

fuse_core/include/fuse_core/loss.h
efernandez: Changed back to default `TAKE_OWNERSHIP`

fuse_core/test/example_loss.h
efernandez: Ownership is now on the `lossFunction` caller and there's no `loss_function_` member now.

This could be changed in a future PR as an optimization.

fuse_core/include/fuse_core/graph_deserializer.h
efernandez: Fixed

fuse_constraints/src/marginalize_variables.cpp
efernandez: Changed. This is back to the original code. Actually like this:
``` c++
    if (fuse_core::Loss::Ownership == ceres::Ownership::TAKE_OWNERSHIP)
    {
      delete loss_function;
    }
```

fuse_loss/include/fuse_loss/arctan_loss.h
efernandez: Assuming this is fine, marking as resolved.

fuse_loss/include/fuse_loss/tolerant_loss.h
efernandez: Considering this good enough and likely irrelevant. Marking as resolved.

fuse_core/include/fuse_core/loss.h
efernandez: Changed, not relevant anymore.

efernandez:
@svwilliams I've addressed all your comments and marked all conversions that have been addressed as resolved. I've even hidden the comments I wrote that have been also fixed or are outdated.

I answer your main comments below (:heavy_check_mark: means I've addressed that; :question: means I made a decision and I'd like your opinion/review/approval on that).
Let me know if I'm missing something. Now you should be able to do a second review pass on this. I'll wait for your feedback.

> (0) I'd prefer the Constraint classes have exclusive ownership of the loss functions. However, the implementation details of pluginlib make that nearly impossible. I am begrudgingly accepting that holding a shared_ptr is the only workable solution.

:heavy_check_mark:

> (1) The fuse_core::Loss class interface deviates from the rest of the fuse design in that the returned Ceres object pointer is owned by this class and not by Ceres. In all other similar cases (variable parameterizations, constraint cost functions), the fuse classes generate a brand new ceres object and return a pointer to that object to Ceres, letting Ceres worry about lifetime management. The same could be done here by making the lossFunction() method pure virtual in the base class and constructing a new ceres::LossFunction object within the derived implementation. But I do not want to dismiss the current design too hastily, as there is potentially some computation benefit to not creating the same object over and over again. Perhaps I'll want to adopt this same pattern for the other classes as well.

:heavy_check_mark: I've changed it back to `TAKE_OWNERSHIP`, but also thought of switching back to `DO_NOT_TAKE_OWNERSHIP` in the future, so my last changes should make that easier.

> (2) Holding a base class ceres::LossFunction pointer in the fuse fuse_core::Loss base class slightly limits the implementation details of derived classes. By holding the pointer in the base class you get to define a default implementation of lossFunction(), but I think that is about it. If you leave that to the derived class, you have the opportunity for the derived class to hold a stack-allocated member object, implementing the lossFunction() as ceres::LossFunction* lossFunction() const { return &member_variable_; } (where const-correctness may need to be adjusted to make Ceres happy). This has the added advantage of not needing to delete a pointer in the destructor (currently missing from the code), and not having to worry about who owns the raw pointer. Since the selection of the loss function is delegated to named fuse classes anyway (fuse_loss::HuberLoss, etc.), I don't think you lose any functionality.

:question: Regarding the part the suggests having a member variable, I believe this is possible, but the Boost serialization part should be split into `load` and `save` to update that mermber variable.

This will also allow to avoid allocating memory from the heap every time `lossFunction()` is called, which should give us some optimization gains. I could try that in a future PR.

What do you think?

> (3) As written, I don't think the serialization system works. Nothing is currently serializing the ceres::LossFunction* pointer held in the fuse_core::Loss base class. Thus, when a fuse_core::Loss object is deserialized, that pointer is either going to be null or uninitialized gibberish. I believe Boost Serialization supports serializing an object via a base class pointer, provided all of the derived classes have been properly registered and serialization functions have been written for them. So, to get serialization working I think you need to write Boost Serialization methods for each type of the derived ceres::LossFunction, and then serialize the base class pointer in the serialize() method of the fuse_core::Loss base class (or, if you use stack-allocated objects as described in (2), the derived ceres::LossFunction member variable would need to be serialized by the derived fuse_core::Loss class). I believe Boost Serialization supports non-intrusive serialization functions, which would allow the serialization functions to be written without modifying the Ceres classes. Alternatively, if the derived fuse_core::Loss objects always return a new ceres::LossFunction object pointer as in (1), then the derived fuse_core::Loss classes would not contain a ceres::LossFunction pointer or object, and hence no serialization methods would be needed for the Ceres classes. That is appealing from a "less work required" point of view, but it does give up the possible computation advantage of not creating and deleting copies of the ceres::LossFunction classes every time the graph is optimized.

:heavy_check_mark:  Should be fixed now.

> (4) You have previously expressed an interest in supporting the ceres::ComposedLoss function. That may represent a challenge; I haven't thought through how that would work with the current fuse_core::Loss classes. You might be able to create a fuse_loss::ComposedLoss class that holds shared_ptrs to two child fuse_core::Loss objects. The initialize() method would then recurse to initialize the two children. Once the children have been created, a ceres::ComposedLoss object could be created using the lossFunction() methods of the children. Maybe? It at least seems plausible.

:question: I need to think about this again. TBH I'd probably be fine with the basic loss functions. The compossed and other complex ones would be nice to have. I think this could also be addressed in a future PR.

Sounds good?

> (5) I'm curious if it would make sense to supply the loss function via the fuse_core::Constraint constructor, instead of as an additional post-construction step. The downside being that all derived constraint constructors would also need to be updated to accept a loss function.

:heavy_check_mark: I'm personally fine with the setter/mutator for now.

> (6) I'm delegating approval of all class and parameter names to Tom.

:heavy_check_mark: Sounds good.

> (7) I have asked about the license preamble and not received any response. Feel free to switch out to your Clearpath preamble. That is actually included in other files already, so why not these.

:heavy_check_mark: Ok, done. Keeping the BSD license, just updating the copyright for the new files.

fuse_loss/include/fuse_loss/arctan_loss.h
efernandez: Done. Also moved the pluginlib header up.

fuse_core/include/fuse_core/constraint.h
efernandez: I've updated the mutator to take the shared pointer by value and `std::move` into the attribute.

IMO I don't see a big need to change the constructor API. And I wouldn't do that in this PR, which already changes many things.

fuse_core/include/fuse_core/constraint.h
efernandez: Updated.

fuse_constraints/src/absolute_orientation_3d_stamped_constraint.cpp
efernandez: That would be great. Not sure what you're thinking of though. Maybe if you share some ideas here or in a separate issue, I can consider contributing something in that direction.

I guess you mean calling the base class `print`, but I personally think that will lead to issue with the order things are printed.

Something that would be useful though, it's some sort of indentation level that is inherited.

Either way, I don't think these are very important things. :smiley:

fuse_core/CMakeLists.txt
efernandez: Done

fuse_core/test/test_loss.cpp
efernandez: Done, including checks for the inlier and outlier region invariants that it should satisfy.

I mentioned this on another conversion that I left unresolved.

svwilliams:
> Regarding the part the suggests having a member variable, I believe this is possible, but the Boost serialization part should be split into load and save to update that mermber variable.

Agreed. Implementing the `save` and `load` serialization methods the ceres::LossFunction objects can be done in a separate PR...if at all.

> I need to think about this again. TBH I'd probably be fine with the basic loss functions....

Also agreed. Composed loss functions can wait for a separate PR...if at all. I have no particular need for them currently.

fuse_core/include/fuse_core/variable.h
```diff
@@ -388,7 +388,7 @@ class Variable
 };
 
 /**
- * Stream operator implementation used for all derived Constraint classes.
+ * Stream operator implementation used for all derived Variable classes.
```
svwilliams, line 391: Thank you

fuse_loss/test/test_huber_loss.cpp
svwilliams: This looks good. Thank you.

svwilliams (approved):
Works for me. Thanks for all of this.

fuse_core/include/fuse_core/constraint.h
ayrton04: The problem with naming it is that it's a really specific concept, so finding something that doesn't involve the word `loss` is going to be hard. This works for me.

ayrton04 (approved):
LGTM. Nice work!

svwilliams:
Just wanted to double-check this is ready to be merged. If so, I'll get that done today.

efernandez:
@svwilliams Yes, it's ready to be merged.

PR #121, opened 2019-11-21
Benchmark unicycle_2d state cost function

efernandez:
This benchmarks the analytic vs automatic differentiation cost functions for the `unicycle_2d` state motion model implemented in https://github.com/locusrobotics/fuse/pull/115, giving the following results:

``` bash
$ sudo cpupower frequency-set --governor performance
Setting cpu: 0
Setting cpu: 1
Setting cpu: 2
Setting cpu: 3
Setting cpu: 4
Setting cpu: 5
Setting cpu: 6
Setting cpu: 7
Setting cpu: 8
Setting cpu: 9
Setting cpu: 10
Setting cpu: 11

$ rosrun fuse_models benchmark_unicycle_2d_state_cost_function
2019-11-21 11:29:36
Running /home/efernandez/dev/ws/cpr_ws/devel/lib/fuse_models/benchmark_unicycle_2d_state_cost_function
Run on (12 X 2201 MHz CPU s)
CPU Caches:
  L1 Data 32K (x6)
  L1 Instruction 32K (x6)
  L2 Unified 256K (x6)
  L3 Unified 9216K (x1)
Load Average: 0.99, 1.25, 1.38
----------------------------------------------------------------------------------------------------------
Benchmark                                                                Time             CPU   Iterations
----------------------------------------------------------------------------------------------------------
Unicycle2DStateCostFunction/AnalyticUnicycle2DCostFunction             384 ns          384 ns      1587923
Unicycle2DStateCostFunction/AutoDiffUnicycle2DStateCostFunction       1491 ns         1491 ns       472094
```

refs CORE-15104

For this to get compiled you need to pass `-DCATKIN_ENABLE_BENCHMARKING=1` to `cmake` or `catkin --cmake-args`.

You also need the `benchmark` debian installed in your system for this to build successfully. You probably need to add it to your CI buildfarm for the build test to pass in this PR.

The benchmark shows the analytic version is almost 4 times faster than the automatic differentiation one. This is too good to be true, so maybe I'm not testing things properly, i.e maybe some initialization stuff is being benchmark for the automatic differentiation version, but I think I'm testing it right.

We could also try using plain arrays in the analytic version. I'd like to try that and benchmark it, but I'm not sure how to organize the code to do so. Should I create a new class that implements the analytic jacobians computation with plain array, so I have two classes (apart from the automatic differentiation) that I can benchmark simultaneously in the same workspace?

FYI @svwilliams

### DIFF
--- a/fuse_models/benchmark/benchmark_unicycle_2d_state_cost_function.cpp
+++ b/fuse_models/benchmark/benchmark_unicycle_2d_state_cost_function.cpp
@@ -1,0 +1,157 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2019, Clearpath Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#include <fuse_models/unicycle_2d_state_cost_function.h>
+#include <fuse_models/unicycle_2d_state_cost_functor.h>
+
+#include <benchmark/benchmark.h>
+
+#include <ceres/autodiff_cost_function.h>
+#include <Eigen/Dense>
+
+#include <vector>
+
+class Unicycle2DStateCostFunction : public benchmark::Fixture
+{
+public:
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
+  Unicycle2DStateCostFunction()
+    : jacobians(num_parameter_blocks)
+    , J(num_parameter_blocks)
+  {
+    for (size_t i = 0; i < num_parameter_blocks; ++i)
+    {
+      J[i].resize(num_residuals, block_sizes[i]);
+      jacobians[i] = J[i].data();
+    }
+  }
+
+  // Analytic cost function
+  static constexpr double dt{ 0.1 };
+  static const fuse_core::Matrix8d sqrt_information;
+
+  static const fuse_models::Unicycle2DStateCostFunction cost_function;
+
+  // Parameters
+  static const double* parameters[];
+
+  // Residuals
+  fuse_core::Vector8d residuals;
+
+  static const std::vector<int32_t>& block_sizes;
+  static const size_t num_parameter_blocks;
+
+  static const size_t num_residuals;
+
+  // Jacobians
+  std::vector<double*> jacobians;
+
+private:
+  // Cost function process noise and covariance
+  static const double process_noise_diagonal[];
+
+  static const fuse_core::Matrix8d covariance;
+
+  // Parameter blocks
+  static const double position1[];
+  static const double yaw1[];
+  static const double vel_linear1[];
+  static const double vel_yaw1[];
+  static const double acc_linear1[];
+
+  static const double position2[];
+  static const double yaw2[];
+  static const double vel_linear2[];
+  static const double vel_yaw2[];
+  static const double acc_linear2[];
+
+  // Jacobian matrices
+  std::vector<fuse_core::MatrixXd> J;
+};
+
+// Cost function process noise and covariance
+const double Unicycle2DStateCostFunction::process_noise_diagonal[] = { 1e-3, 1e-3, 1e-2, 1e-6, 1e-6, 1e-4, 1e-9, 1e-9 };
+
+const fuse_core::Matrix8d Unicycle2DStateCostFunction::covariance =
+    fuse_core::Vector8d(process_noise_diagonal).asDiagonal();
+
+// Parameter blocks
+const double Unicycle2DStateCostFunction::position1[] = { 0.0, 0.0 };
+const double Unicycle2DStateCostFunction::yaw1[] = {0.0};
+const double Unicycle2DStateCostFunction::vel_linear1[] = {1.0, 0.0};
+const double Unicycle2DStateCostFunction::vel_yaw1[] = {1.570796327};
+const double Unicycle2DStateCostFunction::acc_linear1[] = {1.0, 0.0};
+
+const double Unicycle2DStateCostFunction::position2[] = {0.105, 0.0};
+const double Unicycle2DStateCostFunction::yaw2[] = {0.1570796327};
+const double Unicycle2DStateCostFunction::vel_linear2[] = {1.1, 0.0};
+const double Unicycle2DStateCostFunction::vel_yaw2[] = {1.570796327};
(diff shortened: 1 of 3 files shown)


fuse_models/benchmark/benchmark_unicycle_2d_state_cost_function.cpp
```diff
@@ -0,0 +1,156 @@
...
+    , J(num_parameter_blocks)
+  {
+    for (size_t i = 0; i < num_parameter_blocks; ++i)
+    {
+      J[i].resize(num_residuals, block_sizes[i]);
+      jacobians[i] = J[i].data();
+    }
+  }
+
+  // Analytic cost function
```
efernandez, line 60: I tried not to use global variables. For that reason I have a `benchmark::Fixture`. I also used a bunch of `static` attributes, so I'm not sure if this is cleaner/better at the end.

efernandez: In addition, since almost everything is `static`, I didn't have to implement any `SetUp` and `TearDown` methods.
But I wonder if I should maybe do some sort of cleanup, like `J.clear()` and `jacobians.clear()` to make sure things are working as expected. That would change the constructor, and it won't really test things are producing the correct results though. At the end I left it as is because I think the unit test for the analytic jacobians is already doing the job of testing it. This indeed shares some code with that unit test. I'm open to any suggestion to make things cleaner and more modular, if you think it's feasible and worth it.

efernandez (commented):
Add some comments to guide the review, mostly with some design concerns I had.

efernandez:
I just fixed all `roslint` errors but the checks still failed. I suspect it's because the `benchmark` deb is not available/installed in your buildfarm. @svwilliams Could you past the error here and confirm, please?

svwilliams:
```
Cannot locate rosdep definition for [benchmark]
```
I don't see `benchmark` in the standard rosdep definition [here](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml)

fuse_models/benchmark/benchmark_unicycle_2d_state_cost_function.cpp
svwilliams: * I agree that the unit tests verify the correct values are computed, and the benchmark test measures the runtime. No need to mix the responsibilities.
* I didn't know about the Google Benchmark suite before now. This is great that you formalized the test.
* I'm not terribly concerned about sharing code between unit tests or between unit tests and benchmark tests. The code here is mostly just initialization anyway. No need to over complicate the structure.
* I'm not sure it really matters one way or the other about the static members. Assuming the benchmark suite works the same way as the gtest suite, the test fixture constructor should get called before the timing actually occurs, and each benchmark would use a clean test fixture object. I would expect you would get the exact same results with non-static member variables that were initialized in the constructor or setup() method. It is really just personal style preference at this point.

fuse_models/package.xml
```diff
@@ -13,6 +13,7 @@
 
   <build_depend>roslint</build_depend>
 
+  <depend>benchmark</depend>
```
svwilliams: The `benchmark` rosdep entry does not exist in the standard ROS [repo](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml). If you remove it, I expect the buildfarm to build successfully, as it does not know about the `-DCATKIN_ENABLE_BENCHMARKING` flag. I'll consult with our build team, but I suspect benchmark results from our CI system would be largely meaningless anyway. And I can install libbenchmark and run the benchmarks when needed.

svwilliams (approved):

fuse_models/package.xml
efernandez: True, we have it in a private `rosdep` set. You'd only need this:
``` bash
benchmark:
  ubuntu: benchmark
```
The debian is called `benchmark`. I actually created the debian we use, but I think it comes with versions of Ubuntu newer than Xenial. The debian name might change though.

Let me know if you want to give that a try. Otherwise, I can simply remove the dependency, and that should let the checks pass. You'll have to test things locally though, to confirm this builds and the benchmark runs.

svwilliams: I believe this is the official ubuntu package: [libbenchmark-dev](https://packages.ubuntu.com/bionic/libbenchmark-dev)
At least for Ubuntu Bionic. I don't see a package for Xenial.

We do indeed have our own rosdep set, and I'll look at getting this added to it. But since this needs to be buildable without access to our private rosdep files, I'm going to ask that package dependency be removed.

I have successfully run the benchmarks locally, and see a similar large performance gain from the analytic derivatives. Thanks for this.

efernandez: I commented out the `depend` entry for `benchmark` and added a comment explaining why and what needs to happen to uncomment it. I think we could actually send a PR to add it to the rosdep.

efernandez:
FYI @svwilliams I've created https://github.com/ros/rosdistro/pull/23163 to request adding `benchmark` to the official rosdep.

fuse_models/package.xml
```diff
@@ -13,6 +13,22 @@
...
+       ROS Melodic officially supports Ubuntu Bionic:
+       https://www.ros.org/reps/rep-0003.html
+
+       This could be uncommented if we add a benchmark entry like this to the official rosdep:
+
+       benchmark:
+         ubuntu:
+           bionic: libbenchmark-dev
+  -->
+  <!--<depend>benchmark</depend>-->
```
gavanderhoorn: Should this be a `test_depend`?

efernandez: Good point. IMHO it's good to have the ability to run the benchmark on the robot platforms, so in that sense it's better to have the benchmark executable available not only in a test build. Note that in the `CMakeLists.txt` I also added an `install` rule for it.

Either way, this is behind another cmake flag, so it actually doesn't compile by default and it wouldn't be available. `-DCATKIN_ENABLE_BENCHMARKING=1` must be passed to `cmake` or `catkin build --cmake-args` for that.

That being said, I don't have a strong opinion, so I can change it if you think there's no point on having a benchmark generally available.

gavanderhoorn: > Either way, this is behind another cmake flag, so it actually doesn't compile by default and it wouldn't be available. `-DCATKIN_ENABLE_BENCHMARKING=1` must be passed to `cmake` or `catkin build --cmake-args` for that.

true, but the dependency is installed regardless. Seems strange to do that if `DCATKIN_ENABLE_BENCHMARKING` isn't also set by default.

The contributed `rosdep` rule only adds a mapping for Ubuntu, so on other systems users either need to `skip-keys` or edit the manifest to make `rosdep` happy.

In my experience benchmarks are considered part of the tests, and those don't typically get built nor installed. Unless the user asks for them (by setting an option to `true` or `on` fi).

ayrton04: > IMHO it's good to have the ability to run the benchmark on the robot platforms, so in that sense it's better to have the benchmark executable available not only in a test build.

So it sounds like this isn't being built from source on the robot platform in question, or you could just put the new `CATKIN_ENABLE_BENCHMARKING` block inside the `CATKIN_ENABLE_TESTING` block, I think.

Would expanding the rosdep platforms for the dependency address your concern, @gavanderhoorn?

gavanderhoorn: > you could just put the new `CATKIN_ENABLE_BENCHMARKING` block inside the `CATKIN_ENABLE_TESTING` block

That would indeed perhaps be the best thing to do.

> Would expanding the rosdep platforms for the dependency address your concern,

As we don't have optional dependencies in ROS package manifests, anything you add will become a required dependency. I'm not too worried about this one in particular, but it will make setting up a build environment for `fuse` a little more complex.

efernandez: I see your concerns. I think in brief you're asking to keep things simply, so what about:

1. Change `<depend>` into `<test_depend>` for `benchmark`
2. Remove the `CATKIN_ENABLE_BENCHMARKING` block and move the benchmark inside `CATKIN_ENABLE_TESTING`. It won't run, but it'll get build in that case.
3. With the benchmark inside the `CATKIN_ENABLE_TESTING` block, I can remove the `install` rule for the benchmark. I should be able to find where the executable ends up, so there must be a way to call it.

This means that in order to run the benchmark on a robot you'll have to:

1. Create a workspace with `fuse`
2. Run `catkin build` in order to get the `build/fuse_models` folder. Not sure if there's a faster way.
3. Go into the `build/fuse_models` folder and from there there should be a `make` target for the benchmark, that we can build alone
4. Finally, we should be able to run the benchmark

If you agree on this, I can make the changes to unblock this PR. :smiley:

svwilliams: Works for me. The executable should live in: `{catkin_workspace}/devel/lib/fuse_models`
At least if you are using `catkin build`.

efernandez: Updated. Now:
1. There's a `<test_depend>` on `benchmark` as opposed to a `<depend>`
2. I updated the comment in the `package.xml` with the link to the existing PR in the official rosdistro, which has been approved but not merged yet.

- I'll remove that entire comment and uncomment the `<test_depend>` when that gets merged, although it also means you need to sync the rosdistro in your buildfarm for it to work.

- This will make the benchmark compile, but it won't run as a test. I guess we could consider that this way we are at least building the benchmark, i.e. testing it builds, which is better than nothing.

3. I removed the `CATKIN_ENABLE_BENCHMARKING` block and moved the benchmark inside (at the botom) the `CATKIN_ENABLE_TESTING` block.

- However, `catkin build` seems to default to `CATKIN_ENABLE_TESTING=ON`, so it tries to build the benchmark. Note that because it's not a `gtest`, the benchmark target is built immediately. There's no need for `make run_tests`. For this reason, if the `benchmark` package is NOT available in the system it'd fail to build.

4. To solve that issue, I made the `find_package` on `benchmark` not `REQUIRED` (now is `QUIET`), and the benchmark is built only if `benchmark_FOUND` is set to `TRUE`.

I tested this by removing the pkgconfig and cmake benchmark stuff from my system, so the `benchmark` package could NOT be found. It worked:

- No cmake errors/warnings because `benchmark` wasn't found, thanks to the `QUIET` mode in `find_package`
- The benchmark wasn't built

Then I brought the removed files back, touched the `CMakeLists.txt` file to force a re-build and `catkin build` again. After that, the benchmark was built. The resulting executable goes here (one is a symlink):
``` bash
~/ws/devel$ find | grep benchmark
./lib/fuse_models/benchmark_unicycle_2d_state_cost_function
./.private/fuse_models/lib/fuse_models/benchmark_unicycle_2d_state_cost_function
```

And it can be run with `rosrun fuse_models benchmark_unicycle_2d_state_cost_function`.

ayrton04 (approved):

PR #124, opened 2019-12-10
Test depend on benchmark

efernandez:
The `rosdep` entry for `benchmark` is now available for:
* Debian : expect for Stretch or older distros
* Ubuntu : expect for Xenial or older distros
* Fedora

See https://github.com/ros/rosdistro/pull/23163

### DIFF
--- a/fuse_models/package.xml
+++ b/fuse_models/package.xml
@@ -37,25 +37,7 @@
 
   <exec_depend>message_runtime</exec_depend>
 
-  <!-- The official rosdep doesn't have an entry for Google's benchmark deb:
-       https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml
-
-       There's an official deb named libbenchmark-dev available since Ubuntu Bionic:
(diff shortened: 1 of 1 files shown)


efernandez:
@svwilliams For the PR checks to pass you'd likely have to update the `rosdistro` in your buildfarm.

This won't work for Ubuntu Xenial though. If you're running the checks against this Ubuntu version, then simply comment and close this PR.

svwilliams:
Our buildfarm runs for both Xenial and Bionic, and the Xenial build fails for the expected reasons. Thus far the only solution I've come across is creating separate devel-kinetic and devel-melodic branches...which I really don't want to do at this stage. I'll leave this open for now, in case someone has a suggestion on how to handle optional dependencies with rosdep.

efernandez:
Maybe we could close this PR and create an issue pointing to it, to revisit whenever it's more convenient. :thinking:

svwilliams:
Works for me. Closing this PR. See Issue #127.

PR #125, opened 2020-01-03
Validate partial measurements

efernandez:
This runs the following validation check for all partial measurements:

1. All `mean` values must be **finite**
2. The `covariance` must be **symmetric**
3. The `covariance` must be **positive-definite**

If any of the validation checks fail, and `std::runtime_error` exception is thrown.

In this implementation the exception is NOT caught, but it should be trivial to catch and do something like `ROS_WARN_STREAM("<message>: " << ex.what());` and correct the values/covariance, it that sounds like a better approach.

### DIFF
--- a/fuse_core/include/fuse_core/eigen.h
+++ b/fuse_core/include/fuse_core/eigen.h
@@ -36,6 +36,9 @@
 
 #include <Eigen/Core>
 
+#include <sstream>
+#include <string>
+
 
 namespace fuse_core
 {
@@ -65,6 +68,16 @@ using Matrix9d = Eigen::Matrix<double, 9, 9, Eigen::RowMajor>;
 
 template <typename Scalar, int RowsAtCompileTime, int ColsAtCompileTime>
 using Matrix = Eigen::Matrix<Scalar, RowsAtCompileTime, ColsAtCompileTime, Eigen::RowMajor>;
+
+template <typename Derived>
+std::string to_string(const Eigen::DenseBase<Derived>& m)
+{
+  static const Eigen::IOFormat pretty(4, 0, ", ", "\n", "[", "]");
+
+  std::ostringstream oss;
+  oss << m.format(pretty) << '\n';
+  return oss.str();
+}
 
 }  // namespace fuse_core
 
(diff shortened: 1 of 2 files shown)


fuse_core/include/fuse_core/eigen.h
```diff
@@ -66,6 +68,14 @@ using Matrix9d = Eigen::Matrix<double, 9, 9, Eigen::RowMajor>;
 template <typename Scalar, int RowsAtCompileTime, int ColsAtCompileTime>
 using Matrix = Eigen::Matrix<Scalar, RowsAtCompileTime, ColsAtCompileTime, Eigen::RowMajor>;
 
+template<typename Derived>
+std::string to_string(const Eigen::DenseBase<Derived>& m)
+{
+  std::ostringstream oss;
+  oss << m;
```
svwilliams: Do you want to do any fancy Eigen formatting to make it look prettier?
E.g.
```
Eigen::IOFormat pretty(4, 0, ", ", "\n", "[", "]");
oss << m.format(pretty) << "\n"
```
Or whatever you preferred format might be?

svwilliams (approved):

fuse_core/include/fuse_core/eigen.h
```diff
@@ -36,6 +36,8 @@
 
 #include <Eigen/Core>
 
+#include <sstream>
```
svwilliams, line 39: [fuse_core:make] fuse/fuse_core/include/fuse_core/eigen.h:72:  Add #include <string> for string  [build/include_what_you_use] [4]

```suggestion
#include <sstream>
#include <string>
```

ayrton04 (approved):

fuse_core/include/fuse_core/eigen.h
efernandez: Sure thing. Done

efernandez: Sure thing. Done

fuse_core/include/fuse_core/eigen.h
efernandez: Done

PR #129, opened 2020-01-13
Publish linear acceleration

efernandez:
This publishes the linear accelerations and uses them while predicting to the current time if the
new param `predict_with_acceleration` is `true` (default value).

Example output and forward / backwards manual command send to the robot:
acceleration_Screenshot from 2020-01-13 16-07-52

### DIFF
--- a/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
+++ b/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
@@ -134,7 +134,7 @@ public:
 
   bool publish_tf { true };
   bool predict_to_current_time { false };
-  bool predict_with_acceleration { true };
+  bool predict_with_acceleration { false };
   double publish_frequency { 10.0 };
   fuse_core::Matrix8d process_noise_covariance;   //!< Process noise covariance matrix
   ros::Duration tf_cache_time { 10.0 };
--- a/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
+++ b/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
@@ -73,6 +73,7 @@ public:
   {
     nh.getParam("publish_tf", publish_tf);
     nh.getParam("predict_to_current_time", predict_to_current_time);
+    nh.getParam("predict_with_acceleration", predict_with_acceleration);
     nh.getParam("publish_frequency", publish_frequency);
 
     std::vector<double> process_noise_diagonal(8, 0.0);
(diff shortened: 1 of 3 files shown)


fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
```diff
@@ -126,12 +127,14 @@ struct Odometry2DPublisherParams : public ParameterBase
...
 
     nh.getParam("topic", topic);
+    nh.getParam("acceleration_topic", acceleration_topic);
 
     fuse_core::loadCovarianceOptionsFromROS(ros::NodeHandle(nh, "covariance_options"), covariance_options);
   }
 
   bool publish_tf { true };
   bool predict_to_current_time { false };
+  bool predict_with_acceleration { true };
```
efernandez: From my tests, when I use the acceleration it seems I get worse results in the prediction, because I have some delay.

Maybe it's better to default this to `false` :thinking:

I'm also open to changing the name of this param if you don't like it.

fuse_models/src/odometry_2d_publisher.cpp
```diff
@@ -283,13 +316,20 @@ void Odometry2DPublisher::publishTimerCallback(const ros::TimerEvent& event)
...
+
     odom_output_.header.stamp = event.current_real;
+    acceleration_output_.header.stamp = event.current_real;
 
     // Either the last covariance computation was skipped because there was no subscriber,
     // or it failed
     if (latest_covariance_stamp_ == latest_stamp_)
     {
-      Eigen::Matrix<double, 6, 6> covariance;
+      Eigen::Matrix<double, 8, 8> covariance;
```
efernandez: Not strictly from this PR, but if you want, I can change this to `fuse_core::Matrix8d covariance;`

efernandez (commented):
I left a couple of comments for things I'd like to get an answer from the reviewers.

fuse_models/src/odometry_2d_publisher.cpp
svwilliams: Update to `fuse_core::Matrix8d` if you don't mind.

svwilliams (approved):

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
svwilliams: The name seems fine. After reviewing the robot_localization package, I feel like it uses the acceleration data in its prediction. So setting the default to True seems correct to me.

ayrton04 (approved):

fuse_models/src/odometry_2d_publisher.cpp
efernandez: Changed

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
efernandez: Actually, with `predict_with_acceleration` enabled, in my setup, likely due to delays, using acceleration in prediction gives worse results.

Not sure if that's enough a good reason to default it to `false` instead of `true`. I could always set the param to `false` in my setup, but it might be the same in most cases for other people.

Personally, I'm leaning towards the most conservative option, which is setting the default to `false`.

Thoughts?

svwilliams: I don't actually care. Tom says he has no strong opinion either. So as the contributor of this new feature, I defer to you.

efernandez: Ok. I've disabled `predict_with_acceleration` by default.

PR #130, opened 2020-01-14
Scale process noise covariance

efernandez:
This scales the process noise covariance pose by the norm of the current
state velocity.

A new parameter `velocity_norm_min` is added, that prevents the process
noise scaling from setting the pose components to zero or a very small
value that could lead to NaN or a rank deficient Jacobian in the problem
solved, due to an ill-condition covariance for the process noise.

### DIFF
--- a/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
+++ b/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
@@ -91,6 +91,9 @@ public:
 
     process_noise_covariance = fuse_core::Vector8d(process_noise_diagonal.data()).asDiagonal();
 
+    nh.param("scale_process_noise", scale_process_noise, scale_process_noise);
+    nh.param("velocity_norm_min", velocity_norm_min, velocity_norm_min);
+
     double tf_cache_time_double = tf_cache_time.toSec();
     nh.getParam("tf_cache_time", tf_cache_time_double);
     tf_cache_time.fromSec(tf_cache_time_double);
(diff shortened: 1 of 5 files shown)


fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
```diff
@@ -91,6 +91,9 @@ struct Odometry2DPublisherParams : public ParameterBase
 
     process_noise_covariance = fuse_core::Vector8d(process_noise_diagonal.data()).asDiagonal();
 
+    nh.param("scale_process_noise", scale_process_noise, scale_process_noise);
```
svwilliams, line 95: This was called "dynamic_process_noise_covariance" in robot_localization. Do you like the "scale_process_noise" name better? Or would you rather stay closer to the r_l parameter names? We are far from being backwards compatible with r_l, so I'm not particularly concerned either way.
https://github.com/cra-ros-pkg/robot_localization/blob/melodic-devel/src/ros_filter.cpp#L954

svwilliams (approved):

ayrton04 (approved):

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
efernandez: I personally prefer the `scale_process_noise` param name because it's more explicit on the fact that the process noise is being scaled, while `dynamic_process_noise_covariance` seems a little more generic to me.

svwilliams: Works for me.

efernandez: Ok, leaving as is: `scale_process_noise`.

PR #131, opened 2020-01-15
Better validate partial measurement output

efernandez:
* Show source if validation fails
  * Also changes from throwing/crashing to `ROS_ERROR`.

* Add precision arg for covariance validation. This allows to:
    * Relax the default precision when validating the covariance matrix is
      symmetric.
    * Print the covariance matrix with `Eigen::FullPrecision` when the
      symmetry test fails with `isApprox`, so we can see the magnitude of
      the error.

### DIFF
--- a/fuse_models/include/fuse_models/common/sensor_proc.h
+++ b/fuse_models/include/fuse_models/common/sensor_proc.h
@@ -259,6 +259,7 @@ bool transformMessage(const tf2_ros::Buffer& tf_buffer, const T& input, T& outpu
  * @param[in] loss - The loss function for the 2D pose constraint generated
  * @param[in] target_frame - The frame ID into which the pose data will be transformed before it is used
  * @param[in] tf_buffer - The transform buffer with which we will lookup the required transform
+ * @param[in] validate - Whether to validate the measurements or not. If the validation fails no constraint is added
  * @param[out] transaction - The generated variables and constraints are added to this transaction
  * @return true if any constraints were added, false otherwise
  */
@@ -271,6 +272,7 @@ inline bool processAbsolutePoseWithCovariance(
   const std::vector<size_t>& position_indices,
   const std::vector<size_t>& orientation_indices,
   const tf2_ros::Buffer& tf_buffer,
+  const bool validate,
   fuse_core::Transaction& transaction)
 {
   if (position_indices.empty() && orientation_indices.empty())
@@ -323,14 +325,18 @@ inline bool processAbsolutePoseWithCovariance(
 
   populatePartialMeasurement(pose_mean, pose_covariance, indices, pose_mean_partial, pose_covariance_partial);
 
-  try
+  if (validate)
   {
-    validatePartialMeasurement(pose_mean_partial, pose_covariance_partial);
-  }
-  catch (const std::runtime_error& ex)
-  {
-    ROS_ERROR_STREAM_THROTTLE(10.0, "Invalid partial absolute pose measurement from '" << source
-                                                                                       << "' source: " << ex.what());
+    try
+    {
+      validatePartialMeasurement(pose_mean_partial, pose_covariance_partial);
+    }
+    catch (const std::runtime_error& ex)
+    {
+      ROS_ERROR_STREAM_THROTTLE(10.0, "Invalid partial absolute pose measurement from '" << source
+                                                                                         << "' source: " << ex.what());
+      return false;
+    }
   }
 
   // Create an absolute pose constraint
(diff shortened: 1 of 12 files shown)


fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -459,7 +471,16 @@ inline bool processDifferentialPoseWithCovariance(
...
+
+  try
+  {
+    validatePartialMeasurement(pose_relative_mean_partial, pose_relative_covariance_partial, 1e-6);
+  }
+  catch (const std::runtime_error& ex)
+  {
+    ROS_ERROR_STREAM_THROTTLE(10.0, "Invalid partial differential pose measurement from '"
+                                        << source << "' source: " << ex.what());
+  }
```
svwilliams, line 496: I feel like if the validation fails, we should not be creating a constraint from it. In the same way that tf errors prevent the creation of a constraint. Thoughts?

```suggestion
    return false;
  }
```

efernandez: The reason I still create constraints it's because I'm playing with https://github.com/clearpathrobotics/fuse/pull/7 and I still get cases where the covariance is slightly non-symmetric or non-PSD, but the constraint with those slightly bad covariances don't make the optimizer explode.

Maybe we could wait til EOW to see if I get things to actually pass this checks. I'm also planning to create an upstream PR for https://github.com/clearpathrobotics/fuse/pull/7 once I get things to work better.

ayrton04: Would an extra parameter work? Something like `permit_invalid`?

ayrton04 (approved):

fuse_models/include/fuse_models/common/sensor_proc.h
efernandez: Actually, we could have a parameter to disable the checks. Something like `disable_checks` in Ceres solver.

The checks could be enabled by default, so the user can easily detect issues with the inputs and understand what's wrong with them. This would incur a small overhead.

Once things are known to work and be correct consistently, then `disable_checks: true` could be set, so we don't incur any overhead. If you're ok with something like this, I can add it.

efernandez: @svwilliams What's your opinion on this?

fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -189,25 +189,28 @@ inline void populatePartialMeasurement(
...
+    throw std::runtime_error("Non-symmetric partial covariance matrix\n" +
+                             fuse_core::to_string(covariance_partial, Eigen::FullPrecision));
   }
 
   Eigen::SelfAdjointEigenSolver<fuse_core::MatrixXd> solver(covariance_partial);
   if (solver.eigenvalues().minCoeff() <= 0.0)
   {
-    throw std::runtime_error("Non-positive-definite partial covariance matrix " +
+    throw std::runtime_error("Non-positive-definite partial covariance matrix\n" +
                              fuse_core::to_string(covariance_partial));
```
efernandez: I'd like to change this to `fuse_core::to_string(covariance_partial, Eigen::FullPrecision)` as well.

And also print the `solver.eigenvalues()`.

Sounds good?

Mostly a heads up, so you wait for those changes before you merge this PR. :smiley:

fuse_models/include/fuse_models/common/sensor_proc.h
svwilliams: I like the "disable_checks" theory. With checks enabled, failed tests prevent constraints from being generated. And with checks disabled, the constraints are obviously generated all the time.

fuse_models/include/fuse_models/common/sensor_proc.h
svwilliams: More information on failure is better than less. Go for it.

efernandez: Added the additional information.

fuse_models/include/fuse_models/common/sensor_proc.h
efernandez: Add `disable_checks` param (defaults to `false`) to all sensor models and `validate` argument to all `sensor_proc.h` processing functions.

If `disable_checks == false`, then `validate == true` and all validation checks are evaluated. When a validation check fails, no constraint is created for the partial measurements involved.

If `disable_checks == true`, then `validate == false` and no validation check is evaluated. Therefore, constraints are created for all partial measurements.

svwilliams (approved):

PR #133, opened 2020-01-22
Test getConstraints with lvalue iterator

efernandez:
This fails to change the `iterator` from the caller frame, since
all changes happen in the local copy of the iterator into the
`getConstraints` method.

This PR is only to demonstrate `getConstraints` is broken for this use case, which is the one exercised in https://github.com/locusrobotics/fuse/blob/900680df67f9622279ad9f8ac0d588e6294411b4/fuse_constraints/src/marginalize_variables.cpp#L100-L109, where `*p_iter` is always `0` because of this.

It looks like that code has never produced the correct values for the `p` array.


efernandez:
Closing, since this is only to demonstrate the bug in `getConstraints` that is fixed in #134

PR #134, opened 2020-01-22
Fix get constraints with lvalue iterator input

efernandez:
This PR is on top of https://github.com/locusrobotics/fuse/pull/133, that extends the unit test for `getConstraints` to show a bug in the code that it's fixed here. See https://github.com/locusrobotics/fuse/pull/133 for more details.

PR #135, opened 2020-01-22
Test compute elimination order with orphan variables

efernandez:
This tests `computeEliminationOrder` when there are orphan variables in the graph and in the list of variables to marginalize.

Orphan variables are variables with no constraints.

This test shows that when this happens, CCOLAMD fails. If the number of variables is sufficiently large we even get a crash because we end up accessing this `std::vector` out of range: https://github.com/locusrobotics/fuse/blob/devel/fuse_constraints/src/marginalize_variables.cpp#L117, because `variable_order` does NOT contain any entry for the orphan variables. Remember the `operator[]` automatically creates new entries increasing the size of a vector it maintains internally.

This PR is only to show the issue.

It's on top of https://github.com/locusrobotics/fuse/pull/134 because that bug impacts the same logic, although they're two different bugs.

### DIFF
--- a/fuse_constraints/include/fuse_constraints/variable_constraints.h
+++ b/fuse_constraints/include/fuse_constraints/variable_constraints.h
@@ -98,7 +98,7 @@ public:
    * Accessing a variable id that is not part of this container results in undefined behavior
    */
   template <typename OutputIterator>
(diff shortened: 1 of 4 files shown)


efernandez:
Closing, since this PR is only to demonstrate the test fails and that https://github.com/locusrobotics/fuse/pull/136 fixes it.

PR #136, opened 2020-01-22
Fix compute elimination order with orphan variables

efernandez:
This fixes https://github.com/locusrobotics/fuse/pull/135. See https://github.com/locusrobotics/fuse/pull/135 for more details.

This PR is on top of https://github.com/locusrobotics/fuse/pull/135

### DIFF
--- a/fuse_constraints/include/fuse_constraints/variable_constraints.h
+++ b/fuse_constraints/include/fuse_constraints/variable_constraints.h
@@ -94,6 +94,11 @@ public:
   void insert(const unsigned int constraint, VariableIndexIterator first, VariableIndexIterator last);
 
   /**
+   * @brief Add a single orphan variable, i.e. a variable without constraints
+   */
+  void insert(const unsigned int variable);
+
+  /**
    * @brief Insert all of the constraints connected to the requested variable into the provided container
    *
    * Accessing a variable id that is not part of this container results in undefined behavior
--- a/fuse_constraints/src/marginalize_variables.cpp
+++ b/fuse_constraints/src/marginalize_variables.cpp
@@ -68,6 +68,8 @@ UuidOrdering computeEliminationOrder(
   // In order to compute A and p efficiently, we first construct a VariableConstraints object
   // We will construct sequential indices for the variables and constraints while we populate the VariableConstraints
   // object.
+  // For orphan variables, i.e. variables with no constraints, p[c] == p[c+1], which still satisfies CCOLAMD specs:
+  // https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/master/CCOLAMD/Source/ccolamd.c#L174
   auto variable_order = UuidOrdering();
   auto constraint_order = UuidOrdering();
   auto variable_constraints = VariableConstraints();
@@ -76,14 +78,23 @@ UuidOrdering computeEliminationOrder(
     // Get all connected constraints to this variable
     const auto constraints = graph.getConnectedConstraints(variable_uuid);
 
-    // Add each constraint to the VariableConstraints object
-    // New constraint and variable indices are automatically generated
-    for (const auto& constraint : constraints)
+    // If the variable is orphan (it has no constraints), add it to the VariableConstraints object without constraints
+    // New variable index is automatically generated
+    if (boost::empty(constraints))
     {
-      unsigned int constraint_index = constraint_order[constraint.uuid()];
-      for (const auto& constraint_variable_uuid : constraint.variables())
+      variable_constraints.insert(variable_order[variable_uuid]);
+    }
+    else
+    {
+      // Add each constraint to the VariableConstraints object
+      // New constraint and variable indices are automatically generated
+      for (const auto& constraint : constraints)
       {
-        variable_constraints.insert(constraint_index, variable_order[constraint_variable_uuid]);
+        unsigned int constraint_index = constraint_order[constraint.uuid()];
+        for (const auto& constraint_variable_uuid : constraint.variables())
+        {
+          variable_constraints.insert(constraint_index, variable_order[constraint_variable_uuid]);
+        }
       }
     }
   }
@@ -114,7 +125,7 @@ UuidOrdering computeEliminationOrder(
   for (const auto& variable_uuid : marginalized_variables)
   {
     // Reassign the marginalized variables to group0
-    variable_groups[variable_order[variable_uuid]] = 0;
+    variable_groups[variable_order.at(variable_uuid)] = 0;
   }
 
   // Create some additional CCOLAMD required structures
--- a/fuse_constraints/src/variable_constraints.cpp
+++ b/fuse_constraints/src/variable_constraints.cpp
@@ -78,6 +78,15 @@ void VariableConstraints::insert(const unsigned int constraint, std::initializer
   return insert(constraint, variable_list.begin(), variable_list.end());
 }
 
+void VariableConstraints::insert(const unsigned int variable)
+{
+  if (variable >= variable_constraints_.size())
+  {
+    // This automatically create a new variable entry with an empty ConstraintCollection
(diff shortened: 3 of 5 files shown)


svwilliams (approved):

svwilliams (approved):
Thanks for this. I am all for being robust to possible situations. I will issue a word of caution: I am unsure how Ceres-Solver handles the orphan variable situation. I know GTSAM does not support a "forest" of graphs -- multiple unconnected trees in a single container. I skimmed the online Ceres-Solver docs, but I did not see any mention of this situation specifically. The closest thing I found is:
>  In fact you can set any number of parameter blocks to be constant, and Ceres is smart enough to figure out what part of the problem you have constructed depends on the parameter blocks that are free to change and only spends time solving it. So for example if you constructed a problem with a million parameter blocks and 2 million residual blocks, but then set all but one parameter blocks to be constant and say only 10 residual blocks depend on this one non-constant parameter block. Then the computational effort Ceres spends in solving this problem will be the same if you had defined a problem with one parameter block and 10 residual blocks.

So Ceres-Solver may be "smart enough" to handle this as well.

ayrton04 (approved):

efernandez:
I've updated the PR to fix a `roslint` error.

# Orphan Variables

@svwilliams  Regarding your orphan variables comment, I was indeed going to ask that. There could be other parts of the `fuse` library that don't handle that well, or even the Ceres solver itself, although from your comment it might actually be "smart enough".

I wonder if we shouldn't allow a **forest** (of graphs) to exist. I was thinking of starting by instrumenting the code to detect when and where are orphan variables introduced into the graph, and then decide whether it's possible to reject them or not. I suspect the orphan variables should exist for a period of time because they're created by a sensor model, and they can eventually get connected to other variables in the graph by a motion model constraint. In that case, we should allow a **forest** to exist, but maybe we can discard the orphan variables for some computation steps, like the marginalization or optimization, if they don't handle them well. But maybe, after this PR, they're handled well... :thinking: What are your thoughts on this?

# Context for this PR

BTW, I forgot to give you a bit of context here. In brief, with the fixed lag smoother running for a week or so in multiple robots, we eventually got a crash in this logic. I believe it happened twice in 1-2 months.

This time I decided to find a way to reproduce the issue, so I printed the entire graph (with hundreds of variables and constraints) and the list of marginalized variables (to be marginalized, actually). There were 5 orphan variables in the list of marginalized variables, that had a total of 15.

I was getting a crash when the `variable_groups` vector was being destroyed, with `gdb` pointing to https://github.com/clearpathrobotics/fuse/blob/022ce17971c4144cf5bf08ec2812be4855dce736/fuse_constraints/src/marginalize_variables.cpp#L113 . A little weird at the beginning, but at the end it made some sense because we were writing out of range in https://github.com/clearpathrobotics/fuse/blob/022ce17971c4144cf5bf08ec2812be4855dce736/fuse_constraints/src/marginalize_variables.cpp#L117, and getting a crash with this error from `malloc`: `free(): invalid next size (fast)`, because the memory got corrupted.

I created a giant unit test with the graph and marginalized variables I got in the core dump, and it crashed with the same `malloc` error message and pointing to the same line. It was a unit test with more than a thousand lines of code, so that's the reason I simplified it to adding one orphan variable to the simple graph used in other tests. I also tried with a few more orphan variables, being 3 the magic number to get the same crash I got before. I wish it were an easier why to create such a unit test (or program) to reproduce the issue from the core dump. I had to manually C++ify the `gdb` pretty printer output, so I could run the test in C++. After that it was a matter of tracing the logic and printing values to narrow down the issue... the boring part. :smiley: 

I think a `gdb` pretty printer for the `fuse_graphs::HashGraph` would be something useful in this cases.

This should give you an idea of the number of orphan variables I have in my specific instance/configuration of the fixed lag smoother I'm running. It looks like it's quite common to have a bunch of orphan variables in the graph, and eventually it could happen that they're marginalized before they get connected to the rest of the graph. I'm not sure why this could happen, although I remember you mentioned it could be because of the separate logic that handles the time stamps to create the motion model constraints, IIRC. I'm using a `lag_duration` of 2 seconds and a `transaction_timeout` of 0.1 seconds, which I believe you told me they're the only params that could have an impact on this, right? Maybe we could play with those to stress the code, so we build more confidence on the support for orphan variables. :thinking:

svwilliams:
> I suspect the orphan variables should exist for a period of time because they're created by a sensor model, and they can eventually get connected to other variables in the graph by a motion model constraint.

Within the fixed-lag smoother/batch optimizer implementation, motion models are generated and added to the graph at the same time as new sensor transaction. This should prevent the orphaned variables from existing in the graph at any time....at least for a well configured system. As an example of a poorly configured system, we could have omitted the motion model entirely. In such a case, the system is expected to be under-constrained and should break relatively quickly.

However, I believe there is a logic error in how the fixed-lag smoother handles the lag. As a simple example, imagine a smoother with a lag of 10.0s. At some point, the smoother contains variables at t=9s, t=5s, and t=0s. All of these are connected together with motion model constraints, and all three variables are within the lag length of 10.0s.

```
x0 --*-- x5 --*-- x9
```

Then a new measurement arrives at t=12s. A motion model constraint is generated, connect t=9s to t=12s.

```
x0 --*-- x5 --*-- x9 --*-- x12
```

The lag window now covers from t=2s -> t=12s. The variable at t=0s is outside of the lag window and it therefore marginalized out. This also consumes the motion model constraint from t=0s -> t=5s.

```
     *-- x5 --*-- x9 --*-- x12
```

Now let's say a measurement arrives that has been delayed for some reason...computation delay, network delay, whatever. We have a new measurement at t=3s. The lag window spans from t=2s -> t=12s, so this new measurement is inside the window and should be included in the graph. When the motion model is generated to connect it to the graph, the motion model system knows that a constraint from t=0s -> t=5s was previously generated. This constraint must be removed, and two new constraints must be added that cover that same span: one from t=0s -> t=3s and one from t=3s -> t=5s. This is where the logic error comes in. We need to add a constraint from t=0s -> t=3s, but the variable at t=0s has already been marginalized out. We also need to remove the constraint from t=0s -> t=5s, but that constraint has already been removed from the graph as well. Because I did not consider this case when I was coding up the original motion model handling system, something gets out of sync and an orphaned variable get added..either x0 or x3, I don't remember. And even if the orphaned variable was not added, the graph would still be "wrong". The marginal on x5 would still be included in the graph. And that marginal contains information from the t=0s -> t=5s motion model constraint. Additionally, the x3 variable and the motion model from t=3s -> t=5s would be part of the graph. Thus, a portion of the model model is included twice.

```
     *-- x5 --*-- x9 --*-- x12
          |
          *
          |
         x3
```

In contrast, had the variables arrived in order, we would have this graph:

```
*-- x3 --*-- x5 --*-- x9 --*-- x12
```

The intention of the motion model system was to always generate the same graph, regardless of the other the variables were added.

I have a fix for this planned. It is somewhat more complicated than the above illustration might imply, as the cases of multiple motion model chains should be handled. Unfortunately, it may be several weeks before I get time to implement and test the fix. And that assumes the situation described above is indeed the cause of your orphaned variables.

---

Regardless of all of that, the code fix in this PR is valuable and I would like to get it merged. Any chance you can rebase this on the latest version of devel?

efernandez:
> Regardless of all of that, the code fix in this PR is valuable and I would like to get it merged. Any chance you can rebase this on the latest version of devel?

Rebased

> I have a fix for this planned. It is somewhat more complicated than the above illustration might imply, as the cases of multiple motion model chains should be handled. Unfortunately, it may be several weeks before I get time to implement and test the fix. And that assumes the situation described above is indeed the cause of your orphaned variables.

If I have some time, I'd at least try to confirm the issues matches your explanation. I'd be happy to help and contribute a fix, but from your comments it looks like it won't be trivial. And with this PR fix in place, this becomes less critical, so I'll have to focus on other stuff in the short-term. :smiley:

PR #137, opened 2020-01-28
Support dependent relative pose measurements

efernandez:
* Add `independent` param if `differential == true`, that defaults to
  `true` to keep the current behaviour:
  * If `independent == true` the relative pose measurements are assumed
    independent and the covariance of each pose is propagated and added.
  * If `independent == false` the relative pose measurements are assumed
    dependent and the covariance of each pose is propagated and the
    first pose one is substracted from the second one.
* Add `minimum_pose_relative_covariance_diagonal` param if
  `independent == true`, that is added to the resulting pose relative
  covariance in order to guarantee that it's not zero or
  ill-conditioned.

### DIFF
--- a/fuse_models/include/fuse_models/common/sensor_proc.h
+++ b/fuse_models/include/fuse_models/common/sensor_proc.h
@@ -368,12 +368,19 @@ inline bool processAbsolutePoseWithCovariance(
  * pose_relative = pose_absolute1^-1 * pose_absolute2
  *
  * Additionally, the covariance of each pose message is rotated into the robot's base frame at the time of
- * pose_absolute1. They are then added in the constraint. This assumes independence between the pose measurements.
+ * pose_absolute1. They are then added in the constraint if the pose measurements are independent.
+ * Otherwise, if the pose measurements are dependent, the covariance of pose_absolute1 is substracted from the
+ * covariance of pose_absolute2. A small minimum relative covariance is added to avoid getting a zero or
+ * ill-conditioned covariance. This could happen if both covariance matrices are the same or very similar, e.g. when
+ * pose_absolute1 == pose_absolute2, it's possible that the covariance is the same for both poses.
  *
  * @param[in] source - The name of the sensor or motion model that generated this constraint
  * @param[in] device_id - The UUID of the machine
  * @param[in] pose1 - The first (and temporally earlier) PoseWithCovarianceStamped message
  * @param[in] pose2 - The first (and temporally later) PoseWithCovarianceStamped message
+ * @param[in] independent - Whether the pose measurements are indepent or not
+ * @param[in] minimum_pose_relative_covariance - The minimum pose relative covariance that is always added to the
+ *                                               resulting pose relative covariance
  * @param[in] loss - The loss function for the 2D pose constraint generated
  * @param[in] validate - Whether to validate the measurements or not. If the validation fails no constraint is added
  * @param[out] transaction - The generated variables and constraints are added to this transaction
@@ -384,6 +391,8 @@ inline bool processDifferentialPoseWithCovariance(
   const fuse_core::UUID& device_id,
   const geometry_msgs::PoseWithCovarianceStamped& pose1,
   const geometry_msgs::PoseWithCovarianceStamped& pose2,
+  const bool independent,
+  const fuse_core::Matrix3d& minimum_pose_relative_covariance,
   const fuse_core::Loss::SharedPtr& loss,
   const std::vector<size_t>& position_indices,
   const std::vector<size_t>& orientation_indices,
@@ -427,19 +436,6 @@ inline bool processDifferentialPoseWithCovariance(
     sy * x_diff + cy * y_diff,
     (pose2_2d.rotation() - pose1_2d.rotation()).getAngle();
 
-  // Compute Jacobians so we can rotate the covariance
-  fuse_core::Matrix3d j_pose1;
-  j_pose1 <<
-    -cy,  sy,  sy * x_diff + cy * y_diff,
-    -sy, -cy, -cy * x_diff + sy * y_diff,
-      0,   0,                         -1;
-
-  fuse_core::Matrix3d j_pose2;
-  j_pose2 <<
-     cy, -sy,  0,
-     sy,  cy,  0,
-      0,   0,  1;
-
   // Create the covariance components for the constraint
   fuse_core::Matrix3d cov1;
   cov1 <<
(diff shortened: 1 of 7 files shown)


fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -427,19 +436,6 @@ inline bool processDifferentialPoseWithCovariance(
     sy * x_diff + cy * y_diff,
     (pose2_2d.rotation() - pose1_2d.rotation()).getAngle();
 
-  // Compute Jacobians so we can rotate the covariance
```
efernandez, line 430: This is simply moved inside the `if (independent)` block below.

fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -465,7 +461,197 @@ inline bool processDifferentialPoseWithCovariance(
...
+    //        (          0,           0,                                   -1)
+    //
+    //        (R1 | 0)   (cos(yaw1) -sin(yaw1) 0)
+    // J_p2 = (------) = (sin(yaw1)  cos(yaw1) 0)
+    //        ( 0 | 1)   (        0          0 1)
+    //
+    // which are the j_pose1 and j_pose2 jacobians used above for the covariance propagation expresion for independent
+    // pose measurements.
+    //
+    // This seems to be the approach adviced in https://github.com/cra-ros-pkg/robot_localization/issues/356, but after
```
efernandez, line 630: The motivation of this PR is the same as https://github.com/cra-ros-pkg/robot_localization/issues/356

fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -465,7 +461,197 @@ inline bool processDifferentialPoseWithCovariance(
...
+      -sy, -cy, -cy * x_diff + sy * y_diff,
+        0,   0,                         -1;
+
+    fuse_core::Matrix3d j_pose2;
+    j_pose2 <<
+       cy, -sy,  0,
+       sy,  cy,  0,
+        0,   0,  1;
+
+    pose_relative_covariance = j_pose1 * cov1 * j_pose1.transpose() + j_pose2 * cov2 * j_pose2.transpose();
```
efernandez, line 480: I also try this naive change for the `independent == false` case, using the jacobians from this `if (independent)` block: `pose_relative_covariance = j_pose2 * cov2 * j_pose2.transpose() - j_pose1 * cov1 * j_pose1.transpose();`

And also adding a `minimum_pose_relative_covariance` to the resulting `pose_relative_covariance` of that, to avoid zero covariance matrices when `cov1 == cov2`.

efernandez (commented):
Added some comments to help and focus the review.

fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -465,7 +461,197 @@ inline bool processDifferentialPoseWithCovariance(
...
+    //
+    // p2 = p1 * p12    [1]
+    //
+    // where p12 is the relative pose between p1 and p2, which is computed here as:
+    //
+    // p12 = p1^-1 * p2
+    //
+    // Note that the twist t12 is computed as:
+    //
+    // t12 = p12 * dt
```
svwilliams: t12 = p12 / dt ?

svwilliams (approved):
I had considered doing something like this, but I was always fearful that the covariance subtraction would be numerically unstable. It assumes that the pose covariance in the source message is computed and propagated in exactly the same way as it is being reversed here. If the pose/odometry source deviates, for example the source data is integrated at a higher rate than it is published, then the inversion scheme here will be slightly off. This always seemed problematic to me. Since the covariance deltas are likely to be small, slight errors in the inversion could easily create ill-conditioned or even invalid covariances.

Which isn't to say that we should not include this as an option, but more that I am curious how well it works in practice. In particular, I wonder what would happen if you put in some covariance validity checks (symmetric, positive-definite, condition number) before the minimum covariance parameter is added. Not that such checks need to be in the production code; I just wonder how well the covariance subtraction actually works.

fuse_models/include/fuse_models/common/sensor_proc.h
efernandez: Corrected the comment.

efernandez:
> In particular, I wonder what would happen if you put in some covariance validity checks (symmetric, positive-definite, condition number) before the minimum covariance parameter is added. Not that such checks need to be in the production code; I just wonder how well the covariance subtraction actually works.

First, so we're all in the same page, I want to provide a rough idea of what we get with the following approaches:

1. With the `independent` assumption we compute the relative pose covariance as `\Sigma = J_1 * \Sigma_1 * J_1^T + J_2 * \Sigma_2 * J_2^T`. This is basically adding the previous and current absolute pose covariances, and it produces a huge `\Sigma`. It could even be bigger than the map, and it grows unbounded if `\Sigma_1` and `\Sigma_2` (which likely depends on `\Sigma_1`) grow unbounded, as it's the case of wheel odometry, for example.
2. With the `dependent` assumption, if we simply do `\Sigma = J_1 * \Sigma_1 * J_1^T - J_2 * \Sigma_2 * J_2^T`, the resulting covariance is no longer huge, but it's often ill-conditioned (it doesn't pass the symmetry or non-PSD checks).
3. With the `dependent` assumption, and the equation I'm using in this PR, things get better if `\Sigma_2` was computed as we assume from the inverted equation used here.

So, with this PR (item 3 above), we still need a minimum covariance because:
1. In some cases `\Sigma_1 == \Sigma_2`, as in the case of a robot not moving.
2. We still get ill-conditioned covariance matrices quite often in some cases, especially if we reduce the minimum covariance added.

I've been using a minimum covariance equal to a diagonal matrix with `1e-4` in the diagonal, which is certainly large. If I try reducing its value, I start getting issues like this:
```bash
[ WARN] [/fuse_rl] [1580420633.921739552, 1565577217.774568141]: cov1 =[3518904.89393846, 2000488.40958658, 35556.4218530272][2000488.40958658,  1875557.9435916, 28200.4408226669][35556.4218530272, 28200.4408226669, 487.282000827242]
[ WARN] [/fuse_rl] [1580420633.921787147, 1565577217.774568141]: cov2 =[3518095.99970031, 1999582.80770825, 35550.8786857925][1999582.80770825, 1874630.29587236, 28192.4251616275][35550.8786857925, 28192.4251616275, 487.282131955437]
[ERROR] [/fuse_rl] [1580420634.185842287, 1565577218.015982029]: Invalid partial differential pose measurement from 'wheel_odometry' source: Non-symmetric partial covariance matrix[ 6.00000834199614e-06,  4.04517358735906e-08, -3.30848855798916e-09][ 4.06845665172446e-08,  7.25709491734567e-09,  7.37593736795773e-07][-3.30848855798916e-09,  7.37593736795773e-07,   9.8346146762207e-05]
[ WARN] [/fuse_rl] [1580420634.186023508, 1565577218.015982029]: cov1 =[3517490.89481459, 1998903.80485542, 35546.7314742358][1998903.80485542, 1873933.92577313, 28186.4064355888][35546.7314742358, 28186.4064355888, 487.282230301582]
[ WARN] [/fuse_rl] [1580420634.186149422, 1565577218.015982029]: cov2 =[3516887.34337947, 1998224.98326313,  35542.594426693][1998224.98326313, 1873236.89715842, 28180.3807294901][ 35542.594426693, 28180.3807294901, 487.282328647728]
[ERROR] [fuse_rl] [1580420634.273769155, 1565577218.125809971]: Invalid partial differential pose measurement from 'wheel_odometry' source: Non-positive-definite partial covariance matrix[ 3.99970956814397e-06, -6.99140638960684e-10,  4.97952735970431e-12][ -6.9914063896075e-10,  1.25421240412702e-09,  3.27829309820908e-07][ 4.97952735970431e-12,  3.27829309820908e-07,  6.55640980990134e-05] with eigenvalues[-3.85090820281205e-10][  3.9997096903461e-06][ 6.55657372800357e-05]
```

I don't remember exactly what value I used for the diagonal there, but it was likely `1e-6` or `1e-8`.

I could try what you're saying, ie. performing those checks before adding the minimum covariance, but the approach I finally came with in https://github.com/clearpathrobotics/fuse/pull/9 seems better:

- Only needs the minimum covariance just in case the input twist covariance is zero.
- The minimum covariance can be very small and we don't get any ill-conditioned covariance. I've even used values like `1e-18` with no checks complaining.
- It doesn't make any assumption on how the pose and its covariance is computed and propagated.

Obviously, it assumes the relative pose covariance can be retrieved from the twist covariance. This is the case in the diff drive controller in https://github.com/clearpathrobotics/ros_controllers, and I think that should generally be the case.

I see two steps forward from here (not exclusive):

1. Close this PR unmerged because it doesn't solve the problem completely. However, it make things better, or give a minimal support for `dependent`/correlated pose measurements.
2. I create an upstream PR for https://github.com/clearpathrobotics/fuse/pull/9 and we merge it. There we could discuss limitations and details.

efernandez:
I've created https://github.com/locusrobotics/fuse/pull/138 that uses the **twist covariance** for the specific case of the `fuse_models::Odometry2D` sensor model.

It's on top of this PR because I think we should still get this one in for the other sensor models it provides support for dependent pose measurements:

- `fuse_models::Imu2D`
- `fuse_models::Pose2D`

If you agree on this, this PR should be reviewed and merged before reviewing https://github.com/locusrobotics/fuse/pull/138

Otherwise, let me know what changes you'd like to see. :smiley:

svwilliams (approved):
I agree. There are times when velocity information is not available, so this mode is still useful.

@ayrton04 I think this PR is ready for your review, when you get a chance.

fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -368,12 +368,19 @@ inline bool processAbsolutePoseWithCovariance(
  * pose_relative = pose_absolute1^-1 * pose_absolute2
  *
  * Additionally, the covariance of each pose message is rotated into the robot's base frame at the time of
- * pose_absolute1. They are then added in the constraint. This assumes independence between the pose measurements.
+ * pose_absolute1. They are then added in the constraint if the pose measurements are independent.
+ * Otherwise, if the pose measurements are dependent, the covariance of pose_absolute1 is substracted from the
```
ayrton04, line 372: I've found covariance subtraction to be...challenging. Haven't read the code yet, though.

fuse_models/include/fuse_models/parameters/odometry_2d_params.h
```diff
@@ -77,11 +77,38 @@ struct Odometry2DParams : public ParameterBase
...
+          std::vector<double> minimum_pose_relative_covariance_diagonal(3, 0.0);
+          nh.param("minimum_pose_relative_covariance_diagonal", minimum_pose_relative_covariance_diagonal,
+                   minimum_pose_relative_covariance_diagonal);
+
+          if (minimum_pose_relative_covariance_diagonal.size() != 3)
+          {
+            throw std::runtime_error("Minimum pose relative covariance diagonal must be of length 3!");
+          }
+
+          if (std::any_of(minimum_pose_relative_covariance_diagonal.begin(),
```
ayrton04, line 101: Unrelated to this PR, but I'm guessing that we need to validate covariance diagonals in quite a few places. Might be an opportunity for a utility method somewhere.

ayrton04 (approved):
Neat!

fuse_models/include/fuse_models/parameters/odometry_2d_params.h
efernandez: If you don't mind, I can do that in another PR.

I also want to move the checks for **symmetry** and **PSD-ness** to a header in `fuse_core`. That way I can also unit test it, and for the case of **symmetry** I want to stop using `isApprox`, because it doesn't play well with zero (and close to zero) matrices, and for a symmetry check I think it's more intuitive to do a simple element-wise comparison against the user-given precision. I believe all this is enough to justify another PR. I promise I'll send it during this week. :pray:

ayrton04: Sorry! I should clarify: this is great as-is, and even the change I proposed was more of a musing for another time (for anyone to tackle), rather than a concrete ask.

efernandez: I've done this in https://github.com/locusrobotics/fuse/pull/139

Since the changes are on top of this PR, this should be merged before you can review https://github.com/locusrobotics/fuse/pull/139

fuse_models/include/fuse_models/common/sensor_proc.h
efernandez: I'm essentially doing the inverse of the common covariance propagation of `p2 = p1 * dp`, where `p1` and `p2` are two consecutive absolute poses and `dp` is the incremental/differential/relative pose.

That would have a covariance `\Sigma_2 = J_p1 * \Sigma_1 * J_p1^T + J_dp * \Sigma_dp * J_dp`. Here I'm simply moving `\Sigma_dp` to the left hand side. This gives better results than simply changing the `+` into `-` for the equation that was already here, for the covariance, that has the jacobians for `dp = p1^-1 * p2`, not `p2 = p1 * dp`.

Anyway, for the `fuse_models::Odometry2D` sensor model, the approach in https://github.com/locusrobotics/fuse/pull/138 is preferred. This would be an option for the `fuse_models::Imu2D` and `fuse_models::Pose2D` sensor models, because their measurements don't have a **twist** (covariance).

efernandez:
I believe all review comments here and in https://github.com/locusrobotics/fuse/pull/138 has been addressed and both PRs are ready to be merged. And I also created https://github.com/locusrobotics/fuse/pull/139 to factor out some code used in different validation checks and parameter loading.

Let me know if something is still missed. :smiley:

PR #138, opened 2020-02-03
Use twist covariance for differential dependent pose measurements

efernandez:
This PR is on top of https://github.com/locusrobotics/fuse/pull/137 and extends the support for `differential` `dependent` pose measurements for the specific case of the `fuse_modesl::Odometry2D` sensor models.

The suboptimal approach proposed in https://github.com/locusrobotics/fuse/pull/137 has not be discarded because it can still be used for other sensor models:

- `fuse_models::Imu2D`: https://github.com/locusrobotics/fuse/pull/137/files#diff-d5ec9b8db4843c97eab48d6481dedd3e
- `fuse_models::Pose2D`: https://github.com/locusrobotics/fuse/pull/137/files#diff-64dc3a4b2e0f4da6a2cbbb17d7f8562c

Here, for the `fuse_models::Odometry2D` a single commit is contributed, that does the following.

When `differential: true` and `independent: false`, the relative pose covariance should NOT be computed from the consecutive absolute pose covariance matrices because they grow unbounded, so the resulting relative pose covariance suffers from numerical issues.

Instead, we can use the twist covariance of the last pose to compute the relative pose covariance, using the time difference between the consecutive absolute poses. The twist covariance is usually equivalent to the relative pose covariance divided by time difference. Indeed, this is exactly what the `diff_drive_controller` odometry computation logic does in our fork: https://github.com/clearpathrobotics/ros_controllers/blob/indigo-devel/diff_drive_controller/src/odometry.cpp#L198

The only limitation is that we cannot throttle the input topics, because otherwise the twist covariance from the intermediate/throttled messages is missed. We'll have to throttle inside the sensor model, by integrating the intermediate messages. However, in practice, I've throttled the inputs and I haven't seen any issues.

Note that the `minimum_pose_relative_covariance` is still used, but now it can be set to values as small as `1e-18` (for a diagonal covariance matrix) without any issues (the resulting covariance is sane, it passes all the checks: symmetry, PSD-ness). It can still be used to handle the cases when the twist covariance is zero or almost zero. Also note its default value is zero, which should be fine if the producer of the input doesn't generate a zero twist covariance (even when the robot doesn't move).

### DIFF
--- a/fuse_models/include/fuse_models/common/sensor_proc.h
+++ b/fuse_models/include/fuse_models/common/sensor_proc.h
@@ -707,6 +707,188 @@ inline bool processDifferentialPoseWithCovariance(
 }
 
 /**
+ * @brief Extracts relative 2D pose data from a PoseWithCovarianceStamped and adds that data to a fuse Transaction
+ *
+ * This method computes the delta between two poses and creates the required fuse variables and constraints, and then
+ * adds them to the given \p transaction. Only 2D data is used. The pose delta is calculated as
+ *
+ * pose_relative = pose_absolute1^-1 * pose_absolute2
+ *
+ * Additionally, the twist covariance of the last message is used to compute the relative pose covariance using the time
+ * difference between the pose_absolute2 and pose_absolute1 time stamps. This assumes the pose measurements are
+ * dependent. A small minimum relative covariance is added to avoid getting a zero or ill-conditioned covariance. This
+ * could happen if the twist covariance is very small, e.g. when the twist is zero.
+ *
+ * @param[in] source - The name of the sensor or motion model that generated this constraint
+ * @param[in] device_id - The UUID of the machine
+ * @param[in] pose1 - The first (and temporally earlier) PoseWithCovarianceStamped message
+ * @param[in] pose2 - The first (and temporally later) PoseWithCovarianceStamped message
+ * @param[in] twist - The first (and temporally later) TwistWithCovarianceStamped message
+ * @param[in] minimum_pose_relative_covariance - The minimum pose relative covariance that is always added to the
+ *                                               resulting pose relative covariance
+ * @param[in] loss - The loss function for the 2D pose constraint generated
+ * @param[in] validate - Whether to validate the measurements or not. If the validation fails no constraint is added
+ * @param[out] transaction - The generated variables and constraints are added to this transaction
+ * @return true if any constraints were added, false otherwise
+ */
+inline bool processDifferentialPoseWithTwistCovariance(
+  const std::string& source,
+  const fuse_core::UUID& device_id,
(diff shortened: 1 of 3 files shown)


fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -706,6 +706,194 @@ inline bool processDifferentialPoseWithCovariance(
...
+  orientation1->yaw() = pose1_2d.yaw();
+
+  auto position2 = fuse_variables::Position2DStamped::make_shared(pose2.header.stamp, device_id);
+  auto orientation2 = fuse_variables::Orientation2DStamped::make_shared(pose2.header.stamp, device_id);
+  position2->x() = pose2_2d.x();
+  position2->y() = pose2_2d.y();
+  orientation2->yaw() = pose2_2d.yaw();
+
+  // Create the delta for the constraint
+  const double sy = ::sin(-pose1_2d.yaw());
```
svwilliams: Since you don't need `cy` or `sy` outside of the pose manipulation, this could be simplified to something like:
```
auto delta = pose1_2d.inverseTimes(pose2_2d);
fuse_core::Vector3d pose_relative_mean;
pose_relative_mean << delta.x(), delta.y(), delta.yaw();
```

fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -706,6 +706,194 @@ inline bool processDifferentialPoseWithCovariance(
...
+  {
+    ROS_ERROR_STREAM_THROTTLE(10.0, "Very small time difference " << dt << "s from '" << source << "' source.");
+    return false;
+  }
+
+  fuse_core::Matrix3d j_twist;
+  j_twist.setIdentity();
+  j_twist *= dt;
+
+  pose_relative_covariance = j_twist * cov * j_twist.transpose() + minimum_pose_relative_covariance;
```
svwilliams, line 836: Any idea how this interacts with the official version of `ros_controllers`? The odometry calculation there seems very different than the ClearPath fork.
https://github.com/ros-controls/ros_controllers/blob/melodic-devel/diff_drive_controller/src/odometry.cpp

svwilliams (approved):

fuse_models/include/fuse_models/common/sensor_proc.h
efernandez: IIRC Upstream the covariance is just constant, so this will work perfectly, but you'll still get a constant relative pose covariance.

fuse_models/include/fuse_models/common/sensor_proc.h
efernandez: Done, I like it.

ayrton04 (approved):

efernandez:
Just rebased this PR, right after #137 got merged.

PR #139, opened 2020-02-06
Cleanup validation checks

efernandez:
This PR is on top of https://github.com/locusrobotics/fuse/pull/137 and https://github.com/locusrobotics/fuse/pull/138 because it changes code contributed in those.

This provides the following:

1. Add` getCovarianceMatrixDiagonalParam` helper function, that allows to load a covariance matrix from the parameter server, provided in a list with the diagonal values.
2. Add unit test for `util.h`. This only tests the `getCovarianceMatrixDiagonalParam` helper function for now.
3. Add `isSymmetric` and `isPSD` helper functions
4. Add unit test for `eigen.h`. Only tests `isSymmetric` and `isPSD`.

### DIFF
--- a/fuse_core/include/fuse_core/eigen.h
+++ b/fuse_core/include/fuse_core/eigen.h
@@ -70,6 +70,17 @@ using Matrix9d = Eigen::Matrix<double, 9, 9, Eigen::RowMajor>;
 template <typename Scalar, int RowsAtCompileTime, int ColsAtCompileTime>
 using Matrix = Eigen::Matrix<Scalar, RowsAtCompileTime, ColsAtCompileTime, Eigen::RowMajor>;
 
+/**
+ * @brief Serialize a matrix into an std::string using this format:
+ *
+ * [1, 2, 3]
+ * [4, 5, 6]
+ * [7, 8, 9]
+ *
+ * @param[in] m - The matrix to serialize into an std::string.
+ * @param[in] precision - The precision to print the matrix elements with.
+ * @return An std::string with the matrix serialized into it.
+ */
 template <typename Derived>
 std::string to_string(const Eigen::DenseBase<Derived>& m, const int precision = 4)
 {
--- a/fuse_core/include/fuse_core/eigen.h
+++ b/fuse_core/include/fuse_core/eigen.h
@@ -80,6 +80,14 @@ std::string to_string(const Eigen::DenseBase<Derived>& m, const int precision = 
   return oss.str();
 }
 
+/**
+ * @brief Check if a matrix is symmetric.
+ *
+ * @param[in] m - Square matrix to check symmetry on
+ * @param[in] precision - Precision used to compared the matrix m with its transpose, which is the property used to
+ *                        check for symmetry.
+ * @return True if the matrix m is symmetric; False, otherwise.
+ */
 template <typename Derived>
 bool isSymmetric(const Eigen::DenseBase<Derived>& m,
                  const typename Eigen::DenseBase<Derived>::RealScalar precision =
@@ -88,6 +96,12 @@ bool isSymmetric(const Eigen::DenseBase<Derived>& m,
   return m.isApprox(m.transpose(), precision);
 }
 
+/**
+ * @brief Check if a matrix is Positive Definite (PD), i.e. all eigenvalues are `> 0.0`.
+ *
+ * @param[in] m - Square matrix to check PD-ness on.
+ * @return True if the matrix m is PD; False, otherwise.
+ */
 template <typename Derived>
 bool isPSD(const Eigen::DenseBase<Derived>& m)
 {
--- a/fuse_core/include/fuse_core/eigen.h
+++ b/fuse_core/include/fuse_core/eigen.h
@@ -35,6 +35,7 @@
 #define FUSE_CORE_EIGEN_H
 
 #include <Eigen/Core>
+#include <Eigen/Eigenvalues>
 
 #include <sstream>
 #include <string>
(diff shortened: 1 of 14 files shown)


fuse_core/include/fuse_core/eigen.h
```diff
@@ -79,6 +80,21 @@ std::string to_string(const Eigen::DenseBase<Derived>& m, const int precision =
...
+template <typename Derived>
+bool isSymmetric(const Eigen::DenseBase<Derived>& m,
+                 const typename Eigen::DenseBase<Derived>::RealScalar precision =
+                     Eigen::NumTraits<typename Eigen::DenseBase<Derived>::Scalar>::dummy_precision())
+{
+  return m.isApprox(m.transpose(), precision);
+}
+
+template <typename Derived>
+bool isPSD(const Eigen::DenseBase<Derived>& m)
```
svwilliams: Just to be pedantic, doesn't this check for positive-definite (> 0), not positive-semidefinite (>= 0)?

fuse_core/include/fuse_core/eigen.h
```diff
@@ -79,6 +80,21 @@ std::string to_string(const Eigen::DenseBase<Derived>& m, const int precision =
   return oss.str();
 }
 
+template <typename Derived>
+bool isSymmetric(const Eigen::DenseBase<Derived>& m,
+                 const typename Eigen::DenseBase<Derived>::RealScalar precision =
+                     Eigen::NumTraits<typename Eigen::DenseBase<Derived>::Scalar>::dummy_precision())
+{
+  return m.isApprox(m.transpose(), precision);
```
efernandez: I didn't want to change behaviour here, so I still use `isApprox`, but according to https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#ae8443357b808cd393be1b51974213f9c and my experience, when the matrix `m` is zero (or very close to zero), the `precision` doesn't "work" as we'd expect.

Maybe we should use `isMuchSmallerThan` or a `cwise` comparison for the off-diagonal elements here. :thinking:

fuse_core/include/fuse_core/eigen.h
```diff
@@ -79,6 +80,21 @@ std::string to_string(const Eigen::DenseBase<Derived>& m, const int precision =
...
+}
+
+template <typename Derived>
+bool isPSD(const Eigen::DenseBase<Derived>& m)
+{
+  Eigen::SelfAdjointEigenSolver<Derived> solver(m);
+  return solver.eigenvalues().minCoeff() > 0.0;
```
efernandez, line 128: With this we lose the ability to print the `eigenvalues()` when this check fails, as you can see in the code that uses `isPSD` in the rest of the changes in this PR.

If you think that information is important in case of error, we'd have to rethink this check. I think readability and encapsulation is more important in this case, since someone can also copy the matrix `m` into some software and compute the eigenvalues there or other stuff.

fuse_core/include/fuse_core/util.h
```diff
@@ -178,6 +181,45 @@ T getPositiveParam(const ros::NodeHandle& node_handle, const std::string& parame
...
+ * @tparam Size - An int size that specifies the expected size of the covariance matrix (rows and columns)
+ *
+ * @param[in] node_handle - The node handle used to load the parameter
+ * @param[in] parameter_name - The parameter name to load
+ * @param[in] default_value - A default value to use for all the diagonal elements if the provided parameter name does
+ *                            not exist
+ * @return The loaded (or default) covariance matrix, generated from the diagonal vector
+ */
+template <int Size, typename Scalar = double>
+fuse_core::Matrix<Scalar, Size, Size> getCovarianceMatrixDiagonalParam(const ros::NodeHandle& node_handle,
```
efernandez: I'm open to suggestions to change this function name. Maybe `getCovarianceDiagonalParam` is enough, so at least it's a little smaller.

I'm also not sure if this is the best helper function we could provide for this, or whether it should be divided in some additional parts.

efernandez (commented):
Added some comments for some changes I'd like to get some feedback.

svwilliams (approved):
Looks great. Thank you.

fuse_core/include/fuse_core/eigen.h
svwilliams: Something like `((m - m.transpose()).cwiseAbs().maxCoeff() < precision)`? That works for me. That's what I'm doing in the `eigen_gtest.h` file.
https://github.com/locusrobotics/fuse/blob/devel/fuse_core/include/fuse_core/eigen_gtest.h#L99

fuse_core/include/fuse_core/util.h
svwilliams: Either way. I assume if you've noticed anything about my coding style, it's that I don't mind long names.

fuse_core/include/fuse_core/eigen.h
svwilliams: I do like having this wrapped in a function. Having a simple function call makes it far more likely that I will test things that should be tested. And printing a message that says the matrix is not positive definite seems like sufficient information to track down the problem.

fuse_core/include/fuse_core/eigen.h
efernandez: You're right. I'm going to change it to `isPositiveDefinite`.

Maybe I should also add a version for `isPSD` (maybe renamed to `isPositiveSemiDefinite`). It could also be the same function with a flag :thinking:

fuse_core/include/fuse_core/util.h
efernandez: Renamed to `getCovarianceDiagonalParm`, since covariance implies matrix.

fuse_core/include/fuse_core/eigen.h
svwilliams: Adding a flag with a default parameter is an easy change, so you might as well. That said, I'm not sure where in this system a positive-semi-definite matrix is actually acceptable. Maybe there is one somewhere I'm not thinking of.

fuse_core/include/fuse_core/eigen.h
efernandez: Changed to the expression you've suggested.

efernandez:
Rebased and updated with review feedback.

fuse_core/include/fuse_core/eigen.h
efernandez: Ok. Leaving as is for now. If the need comes, then we could add it.

svwilliams (approved):

ayrton04 (approved):

PR #140, opened 2020-02-10
Throttle optimization duration exceeded warning

efernandez:
When the optimization exceeds the configured duration, it's likely to happen quite often, so I think it's better to throttle the `ROS_WARN` message that reports that.

In a future PR we could also try to make the optimization duration available in the diagnostics.

### DIFF
--- a/fuse_optimizers/src/fixed_lag_smoother.cpp
+++ b/fuse_optimizers/src/fixed_lag_smoother.cpp
@@ -218,8 +218,8 @@ void FixedLagSmoother::optimizationLoop()
       auto optimization_complete = ros::Time::now();
       if (optimization_complete > optimization_deadline)
       {
-        ROS_WARN_STREAM("Optimization exceeded the configured duration by " <<
-                        (optimization_complete - optimization_deadline) << "s");
+        ROS_WARN_STREAM_THROTTLE(10.0, "Optimization exceeded the configured duration by "
+                                           << (optimization_complete - optimization_deadline) << "s");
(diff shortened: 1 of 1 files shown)


fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -218,8 +218,8 @@ void FixedLagSmoother::optimizationLoop()
       auto optimization_complete = ros::Time::now();
       if (optimization_complete > optimization_deadline)
       {
-        ROS_WARN_STREAM("Optimization exceeded the configured duration by " <<
-                        (optimization_complete - optimization_deadline) << "s");
+        ROS_WARN_STREAM_THROTTLE(10.0, "Optimization exceeded the configured duration by "
+                                           << (optimization_complete - optimization_deadline) << "s");
```
svwilliams, line 222: Spacing looks weird on github

svwilliams (approved):

fuse_optimizers/src/fixed_lag_smoother.cpp
efernandez: Well, that's actually the spacing I get for those two lines with my `clang-format` configuration. I have a `vim` plugin that uses it, so I can run `:Autoformat` on specific lines.

I can change it if you want.

Although it'd be great if we can have a `.clang-format` file in the `fuse` packages. :smiley:

ayrton04 (approved):

PR #141, opened 2020-02-13
Support ScaledLoss

efernandez:
This adds supports to `ceres::ScaledLoss`: https://github.com/ceres-solver/ceres-solver/blob/032d5844c2db412e25f780c6acf95ecc7a0b7975/include/ceres/loss_function.h#L311-L349

This allows to scale the output of a loss function by a scalar factor `a`.

This is useful if we want to scale the loss function influence of a particular source more than others, or even to workaround the incorrect implementation of the `TukeyLoss`: https://github.com/ceres-solver/ceres-solver/blob/032d5844c2db412e25f780c6acf95ecc7a0b7975/internal/ceres/loss_function.cc#L118-L132, that should be multiplied by `2` because the cost multiplies the output of the loss function `Evaluate` method by `0.5`, which is Ceres solver convention.

PR #142, opened 2020-02-13
New loss functions

efernandez:
This adds the following loss functions:
* **DCS** (Dynamic Covariance Scaling): see http://www2.informatik.uni-freiburg.de/~spinello/agarwalICRA13.pdf (p. 3) and http://www2.informatik.uni-freiburg.de/~agarwal/resources/agarwal-thesis.pdf (p.89)
* **Fair**, **Geman-McClure** and **Welsch**: see http://www.audentia-gestion.fr/research.microsoft/ZhangIVC-97-01.pdf (p. 24)
* **Pseudo-Huber**: see https://en.wikipedia.org/wiki/Huber_loss#Pseudo-Huber_loss_function

### DIFF
--- a/fuse_loss/include/fuse_loss/fair_loss.h
+++ b/fuse_loss/include/fuse_loss/fair_loss.h
@@ -1,0 +1,150 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2020, Clearpath Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_LOSS_FAIR_LOSS_H
+#define FUSE_LOSS_FAIR_LOSS_H
+
+#include <fuse_core/loss.h>
+
+#include <boost/serialization/access.hpp>
+#include <boost/serialization/base_object.hpp>
+#include <boost/serialization/export.hpp>
+
+#include <ostream>
+#include <string>
+
+
+namespace fuse_loss
+{
+
+/**
+ * @brief The Fair loss function.
+ *
+ * This class encapsulates the ceres::FairLoss class, adding the ability to serialize it and load it
+ * dynamically.
+ *
+ * The Fair loss is not provided by the Ceres solver, so it is implemented here, based on table #1 from:
+ * http://www2.informatik.uni-freiburg.de/~spinello/agarwalICRA13.pdf (p. 3)
+ *
+ * See the Ceres documentation for more details. http://ceres-solver.org/nnls_modeling.html#lossfunction
+ */
+class FairLoss : public fuse_core::Loss
+{
+public:
+  FUSE_LOSS_DEFINITIONS(FairLoss);
+
+  /**
+   * @brief Constructor
+   *
+   * @param[in] a FairLoss parameter 'a'
+   */
+  explicit FairLoss(const double a = 1.0);
+
+  /**
+   * @brief Destructor
+   */
+  ~FairLoss() override = default;
+
+  /**
+   * @brief Perform any required post-construction initialization, such as reading from the parameter server.
+   *
+   * This will be called on each plugin after construction.
+   *
+   * @param[in] name A unique name to initialize this plugin instance, such as from the parameter server.
+   */
+  void initialize(const std::string& name) override;
+
+  /**
+   * @brief Print a human-readable description of the loss function to the provided stream.
+   *
+   * @param[out] stream The stream to write to. Defaults to stdout.
+   */
+  void print(std::ostream& stream = std::cout) const override;
+
+  /**
+   * @brief Return a raw pointer to a ceres::LossFunction that implements the loss function
+   *
+   * The Ceres interface requires a raw pointer. Ceres will take ownership of the pointer and promises to properly
+   * delete the loss function when it is done. Additionally, Fuse promises that the Loss object will outlive any
+   * generated loss functions (i.e. the Ceres objects will be destroyed before the Loss Function objects). This
+   * guarantee may allow optimizations for the creation of the loss function objects.
+   *
+   * @return A base pointer to an instance of a derived ceres::LossFunction.
+   */
+  ceres::LossFunction* lossFunction() const override;
+
+  /**
+   * @brief Parameter 'a' accessor.
+   *
+   * @return Parameter 'a'.
+   */
+  double a() const
+  {
+    return a_;
+  }
+
+  /**
+   * @brief Parameter 'a' mutator.
+   *
+   * @param[in] a Parameter 'a'.
+   */
+  void a(const double a)
+  {
+    a_ = a;
+  }
+
+private:
+  double a_{ 1.0 };  //<! FairLoss parameter 'a'
+
+  // Allow Boost Serialization access to private methods
+  friend class boost::serialization::access;
+
+  /**
+   * @brief The Boost Serialize method that serializes all of the data members in to/out of the archive
+   *
+   * @param[in/out] archive - The archive object that holds the serialized class members
+   * @param[in] version - The version of the archive being read/written. Generally unused.
+   */
+  template<class Archive>
+  void serialize(Archive& archive, const unsigned int /* version */)
+  {
+    archive & boost::serialization::base_object<fuse_core::Loss>(*this);
+    archive & a_;
+  }
+};
+
+}  // namespace fuse_loss
+
+BOOST_CLASS_EXPORT_KEY(fuse_loss::FairLoss);
+
+#endif  // FUSE_LOSS_FAIR_LOSS_H
--- a/fuse_loss/include/fuse_loss/loss_function.h
+++ b/fuse_loss/include/fuse_loss/loss_function.h
@@ -1,0 +1,284 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2020, Clearpath Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_LOSS_LOSS_FUNCTION_H
+#define FUSE_LOSS_LOSS_FUNCTION_H
+
+#include <ceres/loss_function.h>
+
+// This provides additional loss functions that are not available in:
+// https://github.com/ceres-solver/ceres-solver/blob/master/include/ceres/loss_function.h
+//
+// Most of them are available in other libraries like:
+// * g2o: https://github.com/RainerKuemmerle/g2o/blob/master/g2o/core/robust_kernel_impl.h
+// * GTSAM: https://github.com/borglab/gtsam/blob/develop/gtsam/linear/LossFunctions.h
+//          (formerly in https://github.com/borglab/gtsam/blob/develop/gtsam/linear/NoiseModel.h)
+//
+// Remember that the loss functions in Ceres solver implement a slight variation of the \rho(r) function, defined for
+// multiple M-estimators in table #1 from http://www.audentia-gestion.fr/research.microsoft/ZhangIVC-97-01.pdf (p. 24).
+//
+// In Ceres the loss functions evaluate:
+//
+//   rho(s) = 2 * \rho(sqrt(s))
+//
+// where s = r^2, being r the residual.
+//
+// This is because the output of rho(s) is mutiplied by 0.5 when the residual block cost is computed, either with or
+// without loss:
+// * cost w/o loss: https://github.com/ceres-solver/ceres-solver/blob/master/internal/ceres/residual_block.cc#L159
+// * cost w/  loss: https://github.com/ceres-solver/ceres-solver/blob/master/internal/ceres/residual_block.cc#L165
+//
+// Note that according to this, it looks like the Tukey loss function is incorrectly implemented in Ceres because it
+// must be multiplied by 2, so instead of dividing by 6 it should divide by 3. See:
+//
+//   https://github.com/ceres-solver/ceres-solver/blob/master/include/ceres/loss_function.h#L281-L282
+//
+// There is an easy workaround for this: combine TukeyLoss with ScaledLoss, using a scaled factor of 2.
+//
+// There is also a PR with a patch already sent to Ceres:
+//
+//   https://ceres-solver-review.googlesource.com/c/ceres-solver/+/16700
+namespace ceres
+{
+
+// Dynamic Covariance Scaling (DCS), equivalent to Switchable Constraints and similar to the Geman-McClure with the
+// tuning constant 'a'.
+//
+// The term is computed as:
+//
+//   rho(s) = a * (3 * s  - a) / (a + s)    for s >  a
+//   rho(s) = s                             for s <= a
+//
+// which gives the weight function:
+//
+//   rho'(s) = min { 1, (2 * a / (a + s))^2 }
+//
+// as described in Eq. 5.19 and 5.20 in:
+//
+//   http://www2.informatik.uni-freiburg.de/~agarwal/resources/agarwal-thesis.pdf (p. 89)
+//
+// that is equal to the square of the scaling factor in Eq. 15 in the original paper:
+//
+//   http://www2.informatik.uni-freiburg.de/~spinello/agarwalICRA13.pdf (p.3)
+//
+// which is also reproduced in Eq. 5.15 in:
+//
+//   http://www2.informatik.uni-freiburg.de/~agarwal/resources/agarwal-thesis.pdf (pp. 85-88)
+//
+// The M-estimator rho(s) equivalent to DCS is obtained from Eq. 5.18 and 5.19 by integration, giving the expression in
+// Eq. 5.20. This way we obtain a valid robust kernel, i.e. one that has positive definite weight function rho'(s) >= 0,
+// that can be used in an Iteratively Reweighted Least Squares (IRLS) problem. For more details, see:
+//
+//   http://www2.informatik.uni-freiburg.de/~agarwal/resources/agarwal-thesis.pdf (p. 89)
+//
+// The relation with the Geman-McClure loss function is explained in:
+//
+//   http://www2.informatik.uni-freiburg.de/~agarwal/resources/agarwal-thesis.pdf (pp. 90-91)
+//
+// The DCS rho(s) equation is equivalent to a scaled and translated generalized Geman-McClure for s > a (outlier region)
+// with a_GemanMcClure = sqrt(a_DCS), as shown in Eq. 5.28. With the rho(s) = 2 * \rho(sqrt(s)) notation, where s = r^2
+// (r = x in the thesis), we have:
+//
+//   rho_DCS(s) = 4 * rho_GemanMcClure(s) - a
+//
+// The implementation in GTSAM has the same weight function rho'(s) we use here:
+//
+//   https://github.com/borglab/gtsam/blob/09b0f03542bfbec5cca62645a60c5d1d4f8f/gtsam/linear/LossFunctions.cpp#L348-L357
+//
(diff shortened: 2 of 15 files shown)


fuse_loss/include/fuse_loss/loss_function.h
```diff
@@ -0,0 +1,191 @@
...
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_LOSS_LOSS_FUNCTION_H
+#define FUSE_LOSS_LOSS_FUNCTION_H
+
+#include <ceres/loss_function.h>
+
+// This provides additional loss functions that are not available in:
```
efernandez, line 39: This comment is important because it explain some details that are very relevant when implemented loss functions compatible with Ceres solver conventions.

It'd be great if you could confirm my conclusions are correct.

fuse_loss/include/fuse_loss/loss_function.h
```diff
@@ -0,0 +1,191 @@
...
+//
+//   rho(s) = s * 2 * a / (a + s)    for s <  a
+//   rho(s) = s                      for s >= a
+//
+// See http://www2.informatik.uni-freiburg.de/~spinello/agarwalICRA13.pdf (p. 3), where 's' is multiplied by the factor:
+//
+//   min(1, 2 * a / (a + s))
+//
+// The g2o implementation seems to use the squared of that factor:
+// https://github.com/RainerKuemmerle/g2o/blob/master/g2o/core/robust_kernel_impl.cpp#L167
```
efernandez: I believe I'm doing the right thing, and either g2o implementation is:
1. Wrong: it doesn't match the explanation on the paper BUT the resulting loss function is good, but different
2. Wrong: it doesn't match the explanation on the paper AND the resulting loss function is bad
3. Right: it simply requires the scale factor squared because of the way the result is used in the rest of the system, e.g. used to multiplied the information matrix, instead of the residuals.

Not sure what's the right answer here or how to validate which one is the correct implementation. I tried plotting the loss function using the scale factor as is and squared, and it looks to me that it shouldn't be squared. I don't have any reference plot to compare with, but with the scale factor as is, it looks closer to **Geman-McClure**.

efernandez (commented):

ayrton04 (approved):

efernandez:
Rebased `--skip` to resolve the conflicts now that #141 is in.

fuse_loss/include/fuse_loss/loss_function.h
```diff
@@ -0,0 +1,191 @@
...
+// * cost w/  loss: https://github.com/ceres-solver/ceres-solver/blob/master/internal/ceres/residual_block.cc#L165
+//
+// Note that according to this, it looks like the following loss functions are incorrectly implemented in Ceres:
+// * Tukey: It must be multiplied by 2, so instead of dividing by 6 it should divide by 3. See:
+//
+//            https://github.com/ceres-solver/ceres-solver/blob/master/include/ceres/loss_function.h#L281-L282
+//
+//          There is an easy workaround for this: combine TukeyLoss with ScaledLoss, using a scaled factor of 2.
+//
+// * Huber: It is actually a Pseudo-Huber loss function, which is smoother than the original Huber one in table #1 from
```
svwilliams: I'm not sure I follow/agree.
From the paper on pg 24, the Huber loss function is defined as:
\rho(x) = x^2/2, if |x| < k
\rho(x) = k * (|x| - k/2), if |x| > k

We need to reformulate that function for Ceres, as you suggest.
RHO = 2 * \rho(sqrt(s)), a = k
which results in:
RHO = s, if sqrt(s) < a
RHO = 2 * a * sqrt(s) - a^2, if sqrt(s) > a

That seems to be what is implemented by Ceres:
r = sqrt(s)
b = a * a
RHO = 2 * a * r - b
https://github.com/ceres-solver/ceres-solver/blob/master/internal/ceres/loss_function.cc#L53

That also seems to follow the standard Huber definition in the wikipedia article
image
and not the pseudo-Huber function.

What is your issue with the Ceres implementation?

fuse_loss/include/fuse_loss/loss_function.h
```diff
@@ -0,0 +1,191 @@
...
+//
+// where s = r^2, being r the residual.
+//
+// This is because the output of rho(s) is mutiplied by 0.5 when the residual block cost is computed, either with or
+// without loss:
+// * cost w/o loss: https://github.com/ceres-solver/ceres-solver/blob/master/internal/ceres/residual_block.cc#L159
+// * cost w/  loss: https://github.com/ceres-solver/ceres-solver/blob/master/internal/ceres/residual_block.cc#L165
+//
+// Note that according to this, it looks like the following loss functions are incorrectly implemented in Ceres:
+// * Tukey: It must be multiplied by 2, so instead of dividing by 6 it should divide by 3. See:
```
svwilliams: I agree. It does seem to be off by a factor of two. Have you tried submitting a patch to Ceres?

fuse_loss/include/fuse_loss/loss_function.h
```diff
@@ -0,0 +1,191 @@
...
+//
+// according to table #1 from http://www.audentia-gestion.fr/research.microsoft/ZhangIVC-97-01.pdf (p. 24).
+//
+// Remember that in Ceres the implementation of rho(s) must be multiplied by 2 because the cost is set as:
+//
+//   cost = 0.5 * rho(s)
+//
+// Here we also consider a tuning constant 'a', so we actually have:
+//
+//   rho(s) = s * a / (a + s)
```
svwilliams: The scale parameter, `a`, generally scales the unsquared residual `x`, not the squared residual `s`. For example, the Fair, Cauchy, Welsh, and Tukey loss functions all use an (x/a) term.

In such a case, the Geman-McClure \rho function would be:
\rho(x) = 0.5 * a^2 * (x/a)^2 / (1 + (x/a)^2)

After converting to Ceres format, we would get:
RHO = a^2 * s / (a^2 + s)

This is the exact same form as what you have written, except that the scale factor is squared. Obviously this is just a difference is the definition of the scale parameter. But for consistency with the other loss functions, do you think it would make more sense to define the scale as (x/a) instead of (s/a)?

fuse_loss/include/fuse_loss/loss_function.h
```diff
@@ -0,0 +1,191 @@
...
+  void Evaluate(double, double*) const override;
+
+private:
+  const double a_;
+};
+
+// Welsch, similar to Tukey loss, it tries to reduce the effect of large errors, but it does not suppress outliers as
+// Tukey might do.
+//
+// The terms i computed as:
```
svwilliams:
```suggestion
// The term is computed as:
```

fuse_loss/include/fuse_loss/welsch_loss.h
```diff
@@ -0,0 +1,150 @@
...
+
+#include <ostream>
+#include <string>
+
+
+namespace fuse_loss
+{
+
+/**
+ * @brief The Fai loss function.
```
svwilliams:
```suggestion
 * @brief The Welsch loss function.
```

fuse_loss/include/fuse_loss/fair_loss.h
```diff
@@ -0,0 +1,150 @@
...
+
+#include <ostream>
+#include <string>
+
+
+namespace fuse_loss
+{
+
+/**
+ * @brief The Fai loss function.
```
svwilliams:
```suggestion
 * @brief The Fair loss function.
```

fuse_loss/include/fuse_loss/loss_function.h
svwilliams: In the paper, the scale factor s_ij is multiplied with the error term, before being squared.
For example, see Eq (6):
... || s_ij * (f(x_i, x_j) - z_ij) ||^2 ...

That would imply to me that the scale factor should be squared before multiplying with `s`, the Ceres squared residual  term.

Looking further on in Eq (6), you see they reduce the equation to: s_kl^2 * X^2
Again, this points towards the scale factor squared times the residual squared being the intent of the paper.

And justbefore the start of Section 4-A on pg4:
> the residuals can be scaled with s_ij or the information matrix be scaled with s_ij^2 depending on implementation

And since this paper was written by Wolfram Burgard, g2o was also written by Wolfram Burgard, and the paper says this scale factor was tested with g2o, I would suspect that g2o contains the reference implementation.

fuse_loss/src/loss_function.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <cmath>
+#include <limits>
+
+
+namespace ceres
+{
+
+void DCSLoss::Evaluate(double s, double rho[3]) const
+{
+  if (s > a_)
```
svwilliams, line 45: Per the previous comment, I feel like this should follow the g2o implementation. With the exception that g2o is a bit sloppy with the derivatives.
```
const double scale = (2.0 * a_) / (a_ + s);
if (scale >= 1.0)
{
  rho[0] = s;
  rho[1] = 1.0;
  rho[2] = 0;
}
else
{
  rho[0] = scale * scale * s;
  rho[1] = ???;
  rho[2] = ???;
}
```

fuse_loss/include/fuse_loss/loss_function.h
efernandez: Yes, but no answer yet: https://ceres-solver-review.googlesource.com/c/ceres-solver/+/16700

I actually created https://ceres-solver-review.googlesource.com/c/ceres-solver/+/16681 first, but I didn't know how to change the description in Gerrit :disappointed:

efernandez: I've mentioned the PR in the comment.

fuse_loss/include/fuse_loss/loss_function.h
efernandez: Thanks for taking the time to look at the paper. I think you're right, `s_kl` in eq. (6) represents the switchable constraint for the edge `kl`. Since that point in the paper, they simply call it `s`, and compute a first intermediate value in eq. (8) and substitute it in `h()`, which is defined as part of eq. (6). I haven't verified the derivative of `\hat{h}`, but I'm going to assume it's correct for now.

So I agree, from the paper, that we should be scaling by the squared, since in Ceres we're multiplying the squared residual.

However, when I plot this approach (let's call it g2o) vs. the one I implemented initially, I get different functions. That's expected, but I'm not sure if it's correct that the weight function is negative for the outlier region. If you think that's ok, then I'd make the change. It's just that the weight function jumps to 0, then goes up (in absolute value) and it finally converges to zero, which is the only part that makes real sense to me.

The following table compares the two implementations. My main concern can be observed in the DCS g2o absolute weight plot, which is what I tried to described in the previous paragraph. What do you think?

| Function | DCS | DCS g2o |
| ------------ | ------ | ------------ |
| `\rho` | rho_dcs | rho_dcs_g2o |
| Influence (`\phi`) | influence_dcs | influence_dcs_g2o |
| Weight | weight_dcs | weight_dcs_g2o |
| Absolute Weight | absolute_weight_dcs | absolute_weight_dcs_g2o |

To put the DCS g2o approach into context, the following table has it together with the other loss functions:

| Function | Plot |
|----| ----|
|`\rho`| rho |
|Influence (`\phi`) | influence|
| Weight | weight|
|Absolute Weight | absolute_weight |

fuse_loss/src/loss_function.cpp
efernandez: Updated in a separate commit.

Note that `s > a_` is equivalent to `scale < 1.0`, for the outlier region. That way I only have to compute `scale` for the outlier region.

I've also computed the derivatives and they pass the finite differences test.

efernandez: I also update the doxygen and plugin documentation accordingly.

fuse_loss/include/fuse_loss/loss_function.h
efernandez: My fault, I didn't reformulate if for Ceres, so I thought it was a different implementation. Now I see it's accurate.

I removed the comment about the Huber loss.

fuse_loss/include/fuse_loss/loss_function.h
efernandez: I think it makes more sensor to define the scale as `x/a`.

I've updated the **GemanMcClure** loss accordingly.

fuse_loss/include/fuse_loss/loss_function.h
efernandez: @svwilliams I've addressed all your comments, including this one. At this moment, the PR uses the squared of the scale factor suggested in the paper, as **g2o** does.

However, I left this conversation unresolved, so you can see how the plots look like (the weight and absolute weight in particular) and comment. If you think this is the right thing, you can merge this PR, since it already has that approach implemented. Otherwise, I can revert the commit back.

Another review pass would be required though, to confirm all comments have been updated and they're consistent. I did a passed myself and I believe they're, but I could have missed some.

fuse_loss/include/fuse_loss/loss_function.h
efernandez: And **g2o** implementation of **Huber** is indeed like Ceres one: https://github.com/RainerKuemmerle/g2o/blob/fcba4eaca6f20d9a5792404cc8ef303aeb8ba5d2/g2o/core/robust_kernel_impl.cpp#L65-L78

It also implements a **PseudoHuber** loss:
https://github.com/RainerKuemmerle/g2o/blob/fcba4eaca6f20d9a5792404cc8ef303aeb8ba5d2/g2o/core/robust_kernel_impl.cpp#L80-L89

Maybe I could contribute it as well, now that I'm touching this. :thinking: 

Acoording to https://github.com/RainerKuemmerle/g2o/blob/master/g2o/core/robust_kernel_impl.h#L87, it must be based on https://en.wikipedia.org/wiki/Huber_loss#Pseudo-Huber_loss_function

fuse_loss/include/fuse_loss/loss_function.h
efernandez: I just gave the DCS **g2o** implementation a try and I got a crashed because the derivative must be positive:

``` bash
WARNING: Logging before InitGoogleLogging() is written to STDERR
F0224 18:28:51.008424  9942 corrector.cc:93] Check failed: rho[1] > 0.0 (-7.18086e-06 vs. 0) 
```

This error message is likely part of an `assert` because right after that the node crashed.

So far I was simply running the unit test.

`rho[1]` is negative by a small margin though, so I'm not completely sure the implementation is wrong, or not acceptable according to some Ceres invariants that loss functions should satisfy, or maybe it's simply a numeric error. @svwilliams Thoughts? :thinking: 

See https://github.com/ceres-solver/ceres-solver/blob/13c7a22cee789fa03a72520b1cb861cd8e9a5a54/internal/ceres/corrector.cc#L87-L93 for the reason behind the check that failed.

fuse_loss/include/fuse_loss/loss_function.h
efernandez: I just added the **PseudoHuber** loss, for the sake of completeness.

fuse_loss/include/fuse_loss/loss_function.h
efernandez: I've done all the math again to confirm the derivatives I've implemented are correct.

In **g2o** the derivatives are wrong: https://github.com/RainerKuemmerle/g2o/blob/fcba4eaca6f20d9a5792404cc8ef303aeb8ba5d2/g2o/core/robust_kernel_impl.cpp#L162-L169

In order for `rho[1]` to be `scale * scale`, `rho[0]` should be `2 * scale * s`. And the second derivative isn't `0`, but `rho[2] = - 2 * rho[1] / (phi + e2)`, where `phi + e2` would be `a + s` in my DCS implementation.

So **g2o** is basically using the squared of the scale factor in the paper for `rho[0]`, but it computes the first derivative as if `rho[0]` had been computed with the scale factor (not squared, and multiplied by `2`), and the second derivative is just set to `0`. :unamused:

efernandez: For the current implementation in this PR, using the squared scale factor as in **g2o**, but the correct derivatives, I get:

- `rho[2] > 0` for `s > a/2`
- `rho[1] < 0` for `s > a`

When this happens, the corrector fails.

Since the outlier region is indeed defined for `s > a`, that means the corrector will always fail for any outlier, so somehow the DCS loss function with the squared scale factor is invalid. Or I'm missing something else.

svwilliams: I read the g2o implementation as pretending the scale term was not a function of s.
`rho[0] = scale * scale * s;`
If you pretend `scale` is a constant, then d/ds( scale * scale * s) = scale * scale
`rho[1] = scale * scale;`
And then the derivative of a constant is zero.
`rho[2] = 0;`

Not that I think this is correct, because `scale` is a function of s.

I don't really have any suggestions right now. I'll poke at the derivatives a bit later, but I'm leaning towards "the DCS formulation is not a valid loss function" being the answer.

efernandez: > I'm leaning towards "the DCS formulation is not a valid loss function" being the answer

 Me too. It's a shame in the paper they didn't plot the `rho`-function or the weight (the `rho` first derivative), so at least I get a better idea of what they actually used. I'd assume it was the **g2o** implementation as you mentioned before.

Yep, it looks like they didn't apply the chain rule for the `scale` derivative, so they treat it as a constant in **g2o**.

At this point, since this PR also contributes other loss functions, I wonder what would be an acceptable path forward to get this PR in:

- [x] Leave the DCS with the squared scale factor, but hack a zero 2nd derivative `rho[2] = 0;`, and `DISABLED_` the unit test for the DCS loss function.
- [ ] Revert the DCS to the implementation that uses the non-squared scale factor, and add a comment explaining why we don't use the squared scale factor, although according to the paper and **g2o** implementation we should be using the squared scale factor.
- [ ] Remove the DCS loss function and leave a comment in the plugins failing explaining why the DCS is not provided. This would be a shame though, because it's available in **g2o** and also **GTSAM**: https://github.com/borglab/gtsam/blob/09b0f03542bfbec5cca62645a60c5d1d4f8fc48c/gtsam/linear/LossFunctions.cpp#L348-L357

That being said, I also wonder if the DCS loss function is not a valid loss function / M-estimator, or it's a type not supported by Ceres corrector. I wish I had a reference book that enunciates the properties an M-estimator should satisfy. I guess I could look at the techniques mentioned in the corrector code, since they should clarify why `rho[1]` must be positive when `rho[2]` is positive. But it's indeed better to have loss functions where `rho[2]` is negative, because that runs a faster path in the corrector, from the little I read in the Ceres implementation file.

svwilliams: I'm curious what happens if you use the g2o derivatives for both the first and second derivative. Obviously it won't pass the finite difference unit tests, but I'm curious if Ceres will freak out. Having the derivatives point in an inaccurate direction may not result in anything good.

If some variation of that doesn't work, I say comment it out.

Either way, you should add a big warning to the DCS header.

efernandez: I think it'd be worth taking **GTSAM** implementation as a reference as well: https://github.com/borglab/gtsam/blob/develop/gtsam/linear/LossFunctions.cpp

I think the `residual` method should be equivalent to `rho[0]` and the `weight` method should be equivalent to `rho[1]`.

IIRC I contributed DCS to **GTSAM** a while ago, but it looks like the noise models have been refactored and changed significantly since them. Indeed, I don't remember there was a separate `weight` and `residual` method, and the `residual` method implementation doesn't look familiar to me, so it was contributed by someone else.

Indeed, this was my contribution: https://github.com/borglab/gtsam/commit/b9c63ef5df83767b055b2fb4165100dffc139dd8

By then, I only had to implement the `weight` method, which I believe I did correctly. :thinking: 

The noise models got separated into a loss function file about 3 months ago, making it harder to blame who contributed the `residuals` method implementation: https://github.com/borglab/gtsam/commit/73542d980ab6cef78a2866b404ac01d05d6178a5

The `residual` method was added in https://github.com/borglab/gtsam/commit/431c37090a1e133ee400e5316cfd206cf792d82f by https://github.com/varunagrawal

efernandez: **GTSAM** in https://github.com/borglab/gtsam/blob/develop/gtsam/linear/LossFunctions.cpp implements the `\rho(x)` (`residual` method) and `w(x)` (`weight` method) equations from table (1) in http://www.audentia-gestion.fr/research.microsoft/ZhangIVC-97-01.pdf (p. 24).

We have that:

- `\rho(x)` is the `residual` method
- `\phi(x) = d \rho(x) / dx`
- `w(x) = \phi(x) / x` is the `weight` method

In **Ceres** we should have the following equivalences in `Evaluate(s)`, where `s = x^2` (`s` being the squared residual, and `x` the residual):

- `rho(s)[0] = 2 * residual(sqrt(s))`, so `residual(x) = 0.5 * rho(x^2)[0]`
- `rho(s)[1] = weight(sqrt(s))` because the 1st derivative of `residual(x)` is `0.5 * rho(x^2)[1] * 2 * x = rho(x^2)[1] * x = \phi(x)`. So `rho(x^2)[1] = \phi(x) / x = w(x)`

A good example is the **Cauchy** loss function:

- Ceres: https://github.com/ceres-solver/ceres-solver/blob/13c7a22cee789fa03a72520b1cb861cd8e9a5a54/internal/ceres/loss_function.cc#L77-L78
- GTSAM: https://github.com/borglab/gtsam/blob/09b0f03542bfbec5cca62645a60c5d1d4f8fc48c/gtsam/linear/LossFunctions.cpp#L211-L218

If we take the DCS implementation in GTSAM: https://github.com/borglab/gtsam/blob/09b0f03542bfbec5cca62645a60c5d1d4f8fc48c/gtsam/linear/LossFunctions.cpp#L348-L367, we have:

- `residual(x) = (a^2 * x^2 + a * x^4) / (a + x^2)^2` that can be further simplified into `a * x^2 * (a + x^2)  / (a + x^2)^2 = a * x^2 / (a + x^2)`
- `weight(x) = (2 * a / (a + x^2))^2`

I've tried to confirm `weight(x) = d residual(x) / dx * x`, but it's not exactly the same, but almost. I get `weight(x) = 2 * (a / (a + x^2))^2`.

In order to get `weight(x) = (2 * a / (a + x^2))^2`, we simply need `residual(x) = 2 * a * x^2 / (a + x^2)`.

Therefore, in **Ceres** we'd have:

- `rho(s)[0] = 2 * residual(sqrt(s)) = 4 * a * s / (a + s)`
- `rho(s)[1] = weight(sqrt(s)) = (2 * a / (a + s))^2`

In other words, the `scale` factor for `rho(s)[0]` would be `4 * a / (a + s)`.

This would be very similar to the **Geman-McClure** loss function: https://github.com/borglab/gtsam/blob/09b0f03542bfbec5cca62645a60c5d1d4f8fc48c/gtsam/linear/LossFunctions.cpp#L314-L325. This is what **GTSAM** has for it:

- `residual(x) = 0.5 * (a^2 * e^2) / (a^2 + e^2)`
- `weight(x) = a^4/(a^2 + e^2)^2`

That in **Ceres** would be:

- `rho(s)[0] = (a^2 * s) / (a^2 + s)`
- `rho(s)[1] = a^4/(a^2 + s)^2`

This is certainly different that **DCS**, but my implementation of **Geman-McClure** in this PR doesn't match that.

It looks like I could do the following:
1. Update the **DCS** implementation to the one described before (similar to **GTSAM**, expect for an additional factor of `2`)
2. Change the **Geman-McClure** implementation to match the one describe before. Note that **GTSAM** implementation actually matches the equations from table (1) in http://www.audentia-gestion.fr/research.microsoft/ZhangIVC-97-01.pdf (p. 24) if we replace `1` with `a^2` in the `w(x)` function. If we replace `x` with `x/a` in `\rho(x)` we don't get the `a^2` though.

Even if **DCS** and **Geman-McClure** are similar, at least with **DCS** there's an inlier and outlier region, so it still makes sense to have both. If I do what I've mentioned before, I should be able to get something that works, since it's very similar to the original implementation I sent to this PR. Does this make sense and work for you, @svwilliams ?

svwilliams: That works for me.

efernandez: I found http://www2.informatik.uni-freiburg.de/~agarwal/resources/agarwal-thesis.pdf and I think now everything makes sense.

I've updated the DCS `rho[0]` equation accordingly and the comments in header, implementation and plugin files. You can see and review the details there.

I also rebased the commits.

This is how the loss functions look like now, including DCS with the last implementation (sorry it's in yellow):

|Function|Plot|
|-|-|
|`\rho(r)`|rho|
|Influence: `\rho'(r) * r`| influence|
|Weight: `\rho'(r)`|weight|
|`\rho''(r)`|second_derivative|

fuse_loss/include/fuse_loss/loss_function.h
```diff
@@ -0,0 +1,284 @@
...
+// that can be used in an Iteratively Reweighted Least Squares (IRLS) problem. For more details, see:
+//
+//   http://www2.informatik.uni-freiburg.de/~agarwal/resources/agarwal-thesis.pdf (p. 89)
+//
+// The relation with the Geman-McClure loss function is explained in:
+//
+//   http://www2.informatik.uni-freiburg.de/~agarwal/resources/agarwal-thesis.pdf (pp. 90-91)
+//
+// The DCS rho(s) equation is equivalent to a scaled and translated generalized Geman-McClure for s > a (outlier region)
+// with a_GemanMcClure = sqrt(a_DCS), as shown in Eq. 5.28. With the rho(s) = 2 * rho(sqrt(s)) notation, where s = r^2
```
svwilliams:
```suggestion
// with a_GemanMcClure = sqrt(a_DCS), as shown in Eq. 5.28. With the rho(s) = 2 * \rho(sqrt(s)) notation, where s = r^2
```
It seems like you were distinguishing the Ceres rho(s) function from the standard rho(r) function with a \. I assumed that was intentional, to make things slightly less confusing.

fuse_loss/include/fuse_loss/loss_function.h
```diff
@@ -0,0 +1,284 @@
...
+//
+// where s = r^2, being r the residual.
+//
+// This is because the output of rho(s) is mutiplied by 0.5 when the residual block cost is computed, either with or
+// without loss:
+// * cost w/o loss: https://github.com/ceres-solver/ceres-solver/blob/master/internal/ceres/residual_block.cc#L159
+// * cost w/  loss: https://github.com/ceres-solver/ceres-solver/blob/master/internal/ceres/residual_block.cc#L165
+//
+// Note that according to this, it looks like the Tukey loss function is incorrectly implemented in Ceres because it
+// must be multiplied by 2, so instead of dividing by 6 it should divide by 3. See:
```
svwilliams, line 62: One final question:
Do you want to make a fuse version of the Tukey loss function with the proper scale? Basically a copy&paste from the Ceres implementation, but then divide by the correct value. We could add a TODO note to remove the fuse version if/when the Ceres PR gets merged. That would avoid having to use the Scale+Tukey combo.

fuse_loss/include/fuse_loss/loss_function.h
efernandez: Yes. TBH I've been a little lazy at naming them properly, so I'm certainly abusing notation by just distinguishing them with the `\` prefix.

Anyway, changed.

fuse_loss/include/fuse_loss/loss_function.h
efernandez: I'll prefer to wait a little more to get an answer from https://ceres-solver-review.googlesource.com/c/ceres-solver/+/16700 before I do that, so I'd prefer this is merged as is now. I'm happy with the `Scale(2) + Tukey` combo workaround for the time being.

Basically, I want to get the confirmation that their implementation must be fixed.

After that I'd send another PR here anyway, because the fixed version would live only in the latest Ceres version. So I'd send a PR that conditionally implements the Tukey loss function here, or uses Ceres one, based on the Ceres version, as I already did for other stuff.

svwilliams (approved):

PR #143, opened 2020-02-13
Plot loss

efernandez:
This adds a test that plots the loss functions `\rho`, influence (`\phi`) and weight functions, as explained in 

The test doesn't really perform any check, but it simply generates the plots, which are useful to understand how the loss functions look like. The fact that the plots are generated (saved as SVG files) is considered as the test not failing, in the sense that it didn't crashed. I couldn't think of a better way to get this in, but I'm open to suggestions.

The plots look like this:

| Function | Plot |
| ------------ | ------ |
| `\rho` | rho |
| influence (`\phi`)      | influence |
| weight     | weight |

As mentioned in https://github.com/locusrobotics/fuse/pull/141, you can see the `TukeyLoss` weight function doesn't reach `1` when the residual `r == 0`. In fact, it reaches `0.5`. This is because I believe it has a bug and it should be multiplied by `2`.

There's a cmake option called `BUILD_WITH_INTERACTIVE_TESTS` that defaults to `OFF`. In this cases it only saves the plots as SVG. When it's turned `ON`, the plots are also shown in Qt windows, so you can zoom and pan to inspect them.

This PR is on top of https://github.com/locusrobotics/fuse/pull/142

### DIFF
--- a/fuse_loss/CMakeLists.txt
+++ b/fuse_loss/CMakeLists.txt
@@ -89,6 +89,15 @@ install(
 if(CATKIN_ENABLE_TESTING)
   find_package(roslint REQUIRED)
 
+  find_package(Qt5 REQUIRED COMPONENTS
+    Core
+    Widgets)
+
+  # Find Qwt using FindQwt.cmake copied from:
+  # https://gitlab.kitware.com/cmake/community/-/wikis/contrib/modules/FindQwt
+  list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+  find_package(Qwt REQUIRED)
+
   # Lint tests
   set(ROSLINT_CPP_OPTS "--filter=-build/c++11,-runtime/references")
   roslint_cpp()
@@ -141,6 +150,41 @@ if(CATKIN_ENABLE_TESTING)
       CXX_STANDARD 14
       CXX_STANDARD_REQUIRED YES
   )
+
+# Qwt Loss Plot Tests
+  catkin_add_gtest(test_qwt_loss_plot
+    test/test_qwt_loss_plot.cpp
+  )
+  add_dependencies(test_qwt_loss_plot
+    ${catkin_EXPORTED_TARGETS}
+  )
+  target_include_directories(test_qwt_loss_plot
+    PRIVATE
+      include
+      ${catkin_INCLUDE_DIRS}
+      ${CERES_INCLUDE_DIRS}
+      ${Qt5Core_INCLUDE_DIRS}
+      ${Qt5Widgets_INCLUDE_DIRS}
+      ${QWT_INCLUDE_DIRS}
+  )
+  target_link_libraries(test_qwt_loss_plot
+    ${PROJECT_NAME}
+    ${catkin_LIBRARIES}
+    ${CERES_LIBRARIES}
+    ${Qt5Core_LIBRARIES}
+    ${Qt5Widgets_LIBRARIES}
+    ${QWT_LIBRARIES}
+  )
+  set_target_properties(test_qwt_loss_plot
+    PROPERTIES
+      CXX_STANDARD 14
+      CXX_STANDARD_REQUIRED YES
+  )
+
+  option(BUILD_WITH_INTERACTIVE_TESTS "Build with interactive tests" OFF)
+  if(BUILD_WITH_INTERACTIVE_TESTS )
+    target_compile_definitions(test_qwt_loss_plot PUBLIC INTERACTIVE_TESTS)
+  endif()
 
   # Scaled Loss Tests
   catkin_add_gtest(test_scaled_loss
--- a/fuse_loss/CMakeLists.txt
+++ b/fuse_loss/CMakeLists.txt
@@ -151,40 +151,43 @@ if(CATKIN_ENABLE_TESTING)
       CXX_STANDARD_REQUIRED YES
   )
 
-# Qwt Loss Plot Tests
-  catkin_add_gtest(test_qwt_loss_plot
-    test/test_qwt_loss_plot.cpp
-  )
-  add_dependencies(test_qwt_loss_plot
-    ${catkin_EXPORTED_TARGETS}
-  )
-  target_include_directories(test_qwt_loss_plot
-    PRIVATE
-      include
-      ${catkin_INCLUDE_DIRS}
-      ${CERES_INCLUDE_DIRS}
-      ${Qt5Core_INCLUDE_DIRS}
-      ${Qt5Widgets_INCLUDE_DIRS}
-      ${QWT_INCLUDE_DIRS}
-  )
-  target_link_libraries(test_qwt_loss_plot
-    ${PROJECT_NAME}
-    ${catkin_LIBRARIES}
-    ${CERES_LIBRARIES}
-    ${Qt5Core_LIBRARIES}
-    ${Qt5Widgets_LIBRARIES}
-    ${QWT_LIBRARIES}
-  )
-  set_target_properties(test_qwt_loss_plot
-    PROPERTIES
-      CXX_STANDARD 14
-      CXX_STANDARD_REQUIRED YES
-  )
+  # Qwt Loss Plot Tests
+  option(BUILD_WITH_PLOT_TESTS "Build with plot tests. These test might fail to run in headless systems." OFF)
+  if(BUILD_WITH_PLOT_TESTS)
+    catkin_add_gtest(test_qwt_loss_plot
+      test/test_qwt_loss_plot.cpp
+    )
+    add_dependencies(test_qwt_loss_plot
+      ${catkin_EXPORTED_TARGETS}
+    )
+    target_include_directories(test_qwt_loss_plot
+      PRIVATE
+        include
+        ${catkin_INCLUDE_DIRS}
+        ${CERES_INCLUDE_DIRS}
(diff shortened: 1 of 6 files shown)


efernandez:
Could someone confirm the PR checks failed because of the Qt5 and Qwt dependencies? Or maybe it's something else. :thinking: 

I thought there are standard `rosdep` entries for both of them: https://github.com/ros/rosdistro/blob/f9999563f4ac8ebe54245ac7301bba0a3af00a00/rosdep/base.yaml#L4914 and https://github.com/ros/rosdistro/blob/f9999563f4ac8ebe54245ac7301bba0a3af00a00/rosdep/base.yaml#L3343, respectively.

Actually, I just saw I depended in `libqwt` instead of `libqwt-qt5-dev`. Let's see if it passes now.

ayrton04:
This PR appears to be including [changes](https://github.com/locusrobotics/fuse/pull/143/files#diff-b1a6f3be767663f123ceb6305799d147R34) made in [other PRs](https://github.com/locusrobotics/fuse/pull/141/files#diff-b1a6f3be767663f123ceb6305799d147R34).

efernandez:
> This PR appears to be including changes made in other PRs.

@ayrton04 Sorry, I forgot to mention this is on top of https://github.com/locusrobotics/fuse/pull/141 and https://github.com/locusrobotics/fuse/pull/142. Those must be approved and merged before this one can be reviewed.

ayrton04:
OK, I'll wait to review. Thanks.

fuse_loss/fuse_plugins.xml
```diff
@@ -8,6 +8,37 @@
 
   In the notation used here, `rho(s)` takes the squared residuals `s` as an input, not the residuals `r`, as in Ceres
   solver.
+
```
efernandez, line 11: I've improved the documentation here.

fuse_loss/include/fuse_loss/qwt_loss_plot.h
```diff
@@ -0,0 +1,297 @@
...
+  {
+    if (size == 0)
+    {
+      return;
+    }
+
+    colormap_.reserve(size);
+
+    double hue = 0.0;
+    const double hue_increment = 1.0 / size;
```
efernandez, line 73: I changed this from the original code I used. Before I was dividing by `size - 1` to get the last color in the colormap with `hue == 1.0`, but that make the first and last color the same because the hue space is circular/periodic. The plots in the description has that original colormap, so the first and last series in the legend has the same color. That's fixed now with this code.

efernandez (commented):
Rebased and added to comments regarding two additional changes wrt the original code submitted.

fuse_loss/CMakeLists.txt
```diff
@@ -33,12 +33,18 @@ add_compile_options(-Wall -Werror)
 add_library(${PROJECT_NAME}
   src/arctan_loss.cpp
   src/cauchy_loss.cpp
+  src/dcs_loss.cpp
+  src/fair_loss.cpp
+  src/geman_mcclure_loss.cpp
```
ayrton04: Does this mean we need to rebase again?

efernandez: Yes, this is on top of https://github.com/locusrobotics/fuse/pull/142, which is still in review.

svwilliams:
Build Failure:
```
[2020-02-26T10:41:57.542Z] 4: [==========] Running 1 test from 1 test case.
[2020-02-26T10:41:57.542Z] 4: [----------] Global test environment set-up.
[2020-02-26T10:41:57.542Z] 4: [----------] 1 test from QwtLossPlotTest
[2020-02-26T10:41:57.542Z] 4: [ RUN      ] QwtLossPlotTest.PlotLossQt
[2020-02-26T10:41:57.542Z] 4: QXcbConnection: Could not connect to display 
[2020-02-26T10:41:57.542Z] 4: Aborted (core dumped)
```

Something about running on a headless cloud instance maybe? I know next to nothing about Qt.

efernandez:
> Something about running on a headless cloud instance maybe? I know next to nothing about Qt.

Makes sense, even if by default the test only saves the plots as SVG but doesn't create any GUI widget. I guess it still requires some Qt stuff for rendering.

I'll try to test it in a headless VM to see if I can reproduce the issue and hopefully fix it.

However, I'm afraid I won't have much time to spend into this in the near future.

If you don't want to have this PR open for too long, one option would be to mark the test as `DISABLED_` for now, so at least it can be run offline if someone wants to compare the loss functions, which IMHO is a nice thing to have. :smiley:

efernandez:
Apart from the checks failing, likely because of the headless machine where they run, this PR can now be reviewed. I just rebased it.

svwilliams:
If you just want to disable the unit test for the plot function, that's fine with me. I don't view the plot function as "live production code".

fuse_loss/include/fuse_loss/qwt_loss_plot.h
```diff
@@ -0,0 +1,333 @@
...
+};
+
+class LossEvaluator
+{
+public:
+  explicit LossEvaluator(const std::vector<double>& residuals) : residuals_(residuals)
+  {
+  }
+
+  std::vector<double> rho(const ceres::LossFunction* loss_function) const
```
efernandez, line 104: I was lazy here and didn't re-use the code of these methods. In principle I should be able to only pass the lambda that is used in `std::transform`. I'm just not sure what would be the argument type for that though.

fuse_loss/include/fuse_loss/qwt_loss_plot.h
```diff
@@ -0,0 +1,333 @@
...
+    grid_.attach(&plot_);
+
+    // Setup panner modifiers:
+    panner_.setMouseButton(Qt::LeftButton, Qt::ControlModifier);
+
+    // Add a legend:
+    plot_.insertLegend(&legend_);
+  }
+
+  void plotRho(const std::shared_ptr<fuse_core::Loss>& loss)
```
efernandez: I also was lazy here. These methods are almost the same, only changing the method called from the `LossEvaluator` class above.

Again, I'm not completely sure of the best way to organize this, and since it's a unit test, I didn't care much at the beginning. But maybe now that there are 4 methods, it's worth trying to factor things out. :thinking:

fuse_loss/test/test_qwt_loss_plot.cpp
```diff
@@ -0,0 +1,206 @@
...
+    for (size_t i = 0; i < samples; ++i)
+    {
+      residuals.push_back(x_min + i * step);
+    }
+  }
+
+  std::vector<double> residuals;
+};
+
+TEST_F(QwtLossPlotTest, PlotLossQt)
```
svwilliams, line 78: Given that there are no actual tests in this unit test, would you want to convert this to just a plain executable?

fuse_loss/include/fuse_loss/qwt_loss_plot.h
svwilliams: It should be pretty trivial to factor out a `QwtPlotCurve* createCurve(name, values)` method. And maybe a `std::string getName(type)` so you don't have the `11` and `-15` in multiple places. But however you want to skin that cat is fine.

fuse_loss/test/test_qwt_loss_plot.cpp
efernandez: One reason I decided to create a test instead of an executable is that I can use `<test_depend>`s, so I don't bring Qt and Qwt dependencies in for the regular package installation.

Another option could be to consolidate this test with the other test for loss functions. That way the test will check the `rho` values are correct for all the residuals evaluated to generate the plots, which is likely a better and more exhaustive test, and still quick. The only reason I didn't do that was because I expected some complexity in the logic that generates the plots, but I can try simplifying it as suggested in the other review comments. But there's still the problem with the headless machine failing to run this test, so that should be resolved before doing something like that.

Thoughts?

svwilliams: The dependency concern is valid. There is always gnuplot as an option. I believe it will run in headless mode. And it looks like there is a rosdep rule for it already. https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L1059

Or you can mark it as deprecated and whoever wants to run it will just need to modify that line. That works for me as well.

efernandez: Not sure about **gnuplot**, but with **GT/Qwt** I can have the interactive mode, that is useful to zoom and pan, so I can inspect the plots better if I enable the interactive mode with the `BUILD_WITH_INTERACTIVE_TESTS` option.

I've just added another option to enable the plotting tests, that default to `OFF`, so the checks should pass now: `BUILD_WITH_PLOT_TESTS`.

This way we can easily enable the test from the command line doing `cmake . -DBUILD_WITH_PLOT_TEST=ON` from the `build` directory.
Or `cmake . -DBUILD_WITH_PLOT_TESTS=ON -DBUILD_WITH_INTERACTIVE_TESTS=ON` if we also want to interact with the plots using the **Qwt** plots.

fuse_loss/include/fuse_loss/qwt_loss_plot.h
efernandez: Done.

Although I guess I could factor things out a bit more.

svwilliams (approved):
Works for me.

ayrton04 (approved):

PR #144, opened 2020-02-14
Visualize loss

efernandez:
This PR shows the impact of a loss function into the constraint cost in the rviz display plugin for the graph.

There are two commits because the first one implements a different visualization approach I explain later. The second commit implements the final approach, that is explained below.

The error line brightness is scaled by the quotient of the constraint cost with and without loss, as it's explained in the code). This way, a dark error line means the cost is reduced significantly by the loss function, which is the expected behaviour when we get outliers, as shown in the screenshot below:

74250011-4e6c4c80-4cea-11ea-87e1-14f6d849bef1

The alternative visualization approach consisted on splitting the error line in two, using the aforementioned quotient to scaled the length of the line that belong to the loss and the rest. In this case, we'd get something like this:

74046263-86178380-49ce-11ea-976e-3e23478ea74b

I used a different width in order to distinguish one segment from the other. They're highlighted in the next image:

74214809-36bba680-4c9f-11ea-8158-6a4c9d417861

I'm open to suggestion and comments regarding the two approaches and whether you think the one I'm going with at the end is better or worse than the previous one (last one explained here). Or maybe I should combine them, but I'm quite happy with the end result for now.

### DIFF
--- a/fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp
+++ b/fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp
@@ -43,6 +43,7 @@
 #include <rviz/ogre_helpers/axes.h>
 #include <rviz/ogre_helpers/billboard_line.h>
 #include <rviz/ogre_helpers/movable_text.h>
+#include <tf2/utils.h>
 #include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 
 #include <OgreQuaternion.h>
@@ -72,6 +73,12 @@ RelativePose2DStampedConstraintVisual::RelativePose2DStampedConstraintVisual(
   error_line_->setMaxPointsPerLine(2);
   error_line_->setNumLines(1);
 
+  // Create constraint loss error line:
+  loss_error_line_node_ = root_node_->createChildSceneNode();
+  loss_error_line_ = std::make_shared<BillboardLine>(scene_manager_, loss_error_line_node_);
+  loss_error_line_->setMaxPointsPerLine(2);
+  loss_error_line_->setNumLines(1);
+
   // Create constraint relative pose axes:
   relative_pose_axes_node_ = root_node_->createChildSceneNode();
   relative_pose_axes_ = std::make_shared<rviz::Axes>(scene_manager_, relative_pose_axes_node_, 10.0, 1.0);
@@ -97,6 +104,7 @@ RelativePose2DStampedConstraintVisual::~RelativePose2DStampedConstraintVisual()
   delete text_;
   scene_manager_->destroySceneNode(relative_pose_line_node_->getName());
   scene_manager_->destroySceneNode(error_line_node_->getName());
+  scene_manager_->destroySceneNode(loss_error_line_node_->getName());
   scene_manager_->destroySceneNode(relative_pose_axes_node_->getName());
   scene_manager_->destroySceneNode(text_node_->getName());
   scene_manager_->destroySceneNode(root_node_->getName());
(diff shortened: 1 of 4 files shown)


ayrton04 (approved):

fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp
```diff
@@ -137,6 +140,61 @@ void RelativePose2DStampedConstraintVisual::setConstraint(
...
+    // Interpolate between the constraint's absolute position and its second variable position by the quotient between
+    // the cost with and without loss:
+    //
+    //              loss_cost      0.5 * rho[0]         rho[0]
+    // loss_scale = --------- = ------------------ = ------------
+    //                cost      0.5 * squared_norm   squared_norm
+    //
+    // Remember that in principle `rho[0] <= squared_norm`, with `rho[0] == squared_norm` for the inlier region, and
+    // `rho[0] < squared_norm` for the outlier region:
+    loss_scale_ = squared_norm == 0.0 ? 0.0 : rho[0] / squared_norm;
```
svwilliams: I'm trying to decide what the scale factor should be in the case where `squared_norm == 0`. If `rho[0]` is also zero, then it seems like the loss function did not modify the cost, and the scale should be 1.0. If `rho[0]` is not equal to zero, then something bad happened? Not sure what the color should be in that case.

svwilliams (approved):
A tiny quibble about an edge case.
As far as the broader question of how the loss function should be visualized? I'm really not sure. Maybe a "rainbow" mode, so inlier costs are green and move to red as the loss function takes affect? But this seems reasonable enough to me.

fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp
efernandez: Good catch. The scale should be `1.0` when the `squared_norm == 0`.

If `rho[0] != 0` when `squared_norm == 0`, something bad happened, but maybe we can rely on the unit tests in `fuse_loss` to confirm that `rho[0] == 0` when `squared_norm == 0`? :thinking: 

Indeed, if `squared_norm == 0` but `rho[0] != 0`, that violates the fact that `rho[0] <= squared_norm`, which I think should always be satisfied.

There are also other case that aren't considered, like negatives values in `rho[0]`, which will also be bad. Maybe for all this cases I could add a messages like this: `ROS_WARN_STREAM_THROTTLE(10.0, "Detected invalid loss cost " << rho[0]);` :thinking:

svwilliams: This is visualization code. I'm not particularly concerned about issuing warnings for invalid loss function implementations here. If we were going to do that, I feel like that should get handled in the loss function base class, or Ceres, or something that gets executed all the time.

I'll leave it to you to decide if/how you want to visualize these cases. Ignore them, set the color to black, whatever you want.

efernandez: I've set the scale to `1.0` if `sqared_norm == 0` and warn if `rho[0] > squared_norm`, and in that case clamp `rho[0]` to `squared_norm`.

That should be fine for now. I can revisit later if I find a better way. :smiley: 

I believe Ceres doesn't check any of that for efficiency reasons, or maybe because in some cases it could still be allowed that `rho[0] > squared_norm` :thinking:

fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp
```diff
@@ -43,15 +43,29 @@
...
 #include <rviz/ogre_helpers/movable_text.h>
+#include <tf2/utils.h>
 #include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 
 #include <OgreQuaternion.h>
 #include <OgreSceneManager.h>
 #include <OgreSceneNode.h>
 
+#include <algorithm>
+
```
svwilliams, line 55: /home/ec2-user/workspace/test_rosdistro_fuse_PR-144/package/fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp:64:  Add #include <string> for string  [build/include_what_you_use] [4]

```suggestion
#include <string>

```

efernandez: Fixed

PR #147, opened 2020-02-19
Use transaction stamp in SerializedPublisher

efernandez:
By using the transaction stamp instead of `ros::Time::now()` it's
possible to replay things with the same transaction and compare the
original and new generated graphs.


svwilliams (approved):
I despise `ros::Time::now()`. It causes problems constantly.

ayrton04 (approved):

PR #149, opened 2020-02-21
Support YAML struct for models and publishers

efernandez:
This allows to compound multiple YAML files that provide additional
models or publishers. This cannot be done with a list/array, because the
previous values get overwritten/lost.

There's an example in the new rostest provided with this PR.


efernandez:
The checks failed because it looks like there's some stuff from https://github.com/locusrobotics/fuse/pull/141 in this PR.

Once https://github.com/locusrobotics/fuse/pull/141 is merged, I'll rebase this one on top of `devel`.

efernandez:
Rebased and roslinted. All checks should pass now.

fuse_optimizers/src/optimizer.cpp
```diff
@@ -100,36 +100,64 @@ void Optimizer::loadMotionModels()
     return;
   }
   // Validate the parameter server values
-  XmlRpc::XmlRpcValue motion_model_list;
```
efernandez, line 103: The diff here is a pain, but I didn't change much. The diff ignoring white space is a little bit better, but still not great because if shows the diff between the previous code for the `TypeArray` vs the new code for the `TypeStruct`.

In brief, I did this for each loading method:
1. Rename the variable where we get the pam to remove the `_list` suffix, because now it could either be a list (`TypeArray`) or a map/dict (`TypeStruct`).
2. Remove the `if` that was checking for `!= TypeArray` and moved the exception thrown into the new `else` block.
3. Created an `if` block for `== TypeArray` with the original code, that only has the variable w/o the `_list` suffix renamed.
4. Created an `else if` block for `== TypeStruct` with the equivalent code for the struct, which is similar to the `TypeArray` one, but with subtle differences.

ayrton04 (approved):

fuse_optimizers/src/optimizer.cpp
```diff
@@ -100,36 +100,64 @@ void Optimizer::loadMotionModels()
...
+      // Get the setting we need from the parameter server
+      std::string motion_model_name = static_cast<std::string>(motion_models[motion_model_index]["name"]);
+      std::string motion_model_type = static_cast<std::string>(motion_models[motion_model_index]["type"]);
+      // Create a motion model object using pluginlib. This will throw if the plugin name is not found.
+      auto motion_model = motion_model_loader_.createUniqueInstance(motion_model_type);
+      // Initialize the publisher
+      motion_model->initialize(motion_model_name);
+      // Store the publisher in a member variable for use later
+      motion_models_.emplace(motion_model_name, std::move(motion_model));
+    }
```
svwilliams: Can we separate out the parameter parsing from the plugin initialization?
Something like:
```
private_node_handle_.getParam("motion_models", motion_models);
struct PluginConfig
{
  std::string name;
  std::string type;
};
std::vector<PluginConfig> motion_model_configs;
if (motion_models.getType() == XmlRpc::XmlRpcValue::TypeArray)
{
if ( (motion_models[motion_model_index].getType() != XmlRpc::XmlRpcValue::TypeStruct)
    || (!motion_models[motion_model_index].hasMember("name"))
    || (!motion_models[motion_model_index].hasMember("type")))
  {
    throw std::invalid_argument("The 'motion_models' parameter should be a list of the form: "
                                "-{name: string, type: string}");
  }
  motion_model_configs.push_back(
    {static_cast<std::string>(motion_models[motion_model_index]["name"]),
     static_cast<std::string>(motion_models[motion_model_index]["type"])});
}
else if (motion_models.getType() == XmlRpc::XmlRpcValue::TypeStruct)
{
  if ( (motion_model_entry.second.getType() != XmlRpc::XmlRpcValue::TypeStruct)
    || (!motion_model_entry.second.hasMember("type")))
  {
    throw std::invalid_argument("The 'motion_models' parameter should be a struct of the form: "
                                "{string: {type: string}}");
  }
  motion_model_configs.push_back(
    {static_cast<std::string>(motion_model_entry.first),
     static_cast<std::string>(motion_model_entry.second["type"])});
}
else
{
  throw std::invalid_argument("The 'motion_models' parameter should be a list of the form: "
                              "-{name: string, type: string} or a struct of the form: {string: {type: string}}");
}
for (const auto& config : motion_model_configs)
{
  auto motion_model = motion_model_loader_.createUniqueInstance(config.type);
  // Initialize the motion model
  motion_model->initialize(config.name);
  // Store the motion model in a member variable for use later
  motion_models_.emplace(motion_model_name, std::move(motion_model));
}
```
That removes the repeated code from the two parameter server layout options. It's not that big of a deal for the motion models, but the sensor models are bit more involved.

Either that, or drop support for the list version. I'm fine with that as well.

fuse_optimizers/test/common_robot_config.yaml
```diff
@@ -0,0 +1,36 @@
...
+  laser_localization:
+    type: fuse_models::Odometry2D
+    motion_models: ['unicycle_2d']
+
+publishers:
+  odometry_publisher:
+    type: fuse_models::Odometry2DPublisher
+
+
+unicycle_2d:
```
svwilliams: I'm curious what you think. Given the optimizer layout above, we could move all of the model-specific configuration into a single place. Like so:
```
motion_models:
  unicycle_2d:
    type: fuse_models::Unicycle2D
    process_noise_diagonal: [0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1]

sensor_models:
  unicycle_2d_ignition:
    type: fuse_models::Unicycle2DIgnition
    motion_models: ['unicycle_2d']
    initial_sigma: [0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1]
    publish_on_startup: false
  wheel_odometry:
    type: fuse_models::Odometry2D
    motion_models: ['unicycle_2d']
    topic: odom
    differential: true
    twist_target_frame: base_link
  laser_localization:
    type: fuse_models::Odometry2D
    motion_models: ['unicycle_2d']
    topic: pose
    pose_target_frame: map
    twist_target_frame: base_link

publishers:
  odometry_publisher:
    type: fuse_models::Odometry2DPublisher
```

It would require a bit of extra work, as the motion model/sensor/publisher interface would need to change to accept a node handle in the configuration namespace. And that doesn't need to happen in this PR. I'm just curious if you like that better? Or do you want to keep the motion model configuration separate from the optimizer configuration (as is done now)?

svwilliams (changes requested):
There's some code duplication between the `TypeArray` handing and the `TypeStruct` handling that I'd like to eliminate.

fuse_optimizers/test/common_robot_config.yaml
efernandez: Uhm, interesting. The config suggestion here would be more compact, but on the other side, I like that with the current configuration we have something like a top view / summary / interface, and the details are **enforced** to be somewhere else.

Certainly, with your suggestion we can still have the model specific parameters later in the file, or in another file, if we want to provide a summary-like config w/o the details (model specific parameters). It's simply not enforced. I believe this is the only different from the config point of view.

Another point would be that with your suggested approach here, things get better organized in the param server, since we have all the config under less namespaces, e.g. `sensor_models` will have all the sensor models params, instead of having each in its own, "top" level namespace.

I certainly don't have a strong opinion. Probably your suggested config format is superior because we can do everything we can with the current one and more, so it's more flexible.

fuse_optimizers/src/optimizer.cpp
efernandez: Done.

I prefer to keep backwards compatibility with the **list/array** format, at least for now. But it's probably a good idea to deprecate it, since the **struct** format looks superior.

I also added a test for the **array** configuration format, and organized the tests better.

svwilliams (approved):

fuse_optimizers/test/common_robot_config.yaml
svwilliams: Alright. I'll look into what needs to happen to support that. Thanks.

efernandez: Another thing that is still hard to extend, is the `ignition_sensors` parameter, which is a list/array.

It'd be great if we could remove that parameter completely, and add something to the `SensorModel` plugins, so they have a method that indicates whether they're ignition sensor models or not. It sounds like a compile time thing that could be done with traits, but I believe it'd have to be done with a runtime virtual method because we're using plugins.

Thoughts?

svwilliams: The parameter could definitely be moved from a separate list to part of the sensor config. Something like:
```
sensor_models:
  unicycle_2d_ignition:
    type: fuse_models::Unicycle2DIgnition
    ignition: true
    motion_models: ['unicycle_2d']
```
And maybe we could think of a better name than "ignition" at some point.

But I view "ignition-ness" more as a situation-specific property and less as a property of the sensor itself. Perhaps someone whats to use a laserscan-to-map match to initialize the state. Or maybe a GPS reading. At least that was the original intention of the system. At this point I'm not sure if reality matches that intention or not. But I will think about it.

efernandez: Makes sense.

I was mostly thinking of the ability of a SensorModel to work as an ignition sensor model being something in the class.

But you're also right in that the user could decide whether to use such a sensor model as a regular one, i.e. not being used for "ignition", or not.

It's probably a good idea to have both things, so if someone sets `ignition: true` on a sensor that doesn't have igniton capabilities, an exception is thrown.

PR #154, opened 2020-03-11
Fix Unicycle2DIgnition set_pose

efernandez:
Sometimes the pose set with the `set_pose` service/topic using the `Unicycle2DIgnition` sensor model is ignored and the pose is set at a wrong, different location, typically at the origin of the map. This happens when the other sensor models include the `Odometry2D` using the `differential` mode.

This PR fixes several bugs that all contributed to this issue. In particular, it provides the following fixes:
1. Initialize `StateHistoryElement::velocity_yaw` which was used uninitialized: https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/include/fuse_models/unicycle_2d.h#L102
    - This impacted prediction in https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L177 because `base_state` is default constructed when the `beginning_stamp` isn't  in the state history: https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L139-L150
    - This produced incorrect predictions starting from the first state in the history.
    - And it also propagated into other states because the resulting predicted state is stored in the history: https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L234-L235
    - The history and predicted states are used to set the initial value of the variables in the motion model constraint: https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L206-L221 https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L246-L258, and these variables overwrite the ones from other constraints: https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L83 https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_core/src/timestamp_manager.cpp#L163, so the optimization ends up using a bad initialization and cannot converge to the correct solution.

2. Process ignition transactions individually. This is important because the state history is only updated on graph updates: https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L95
    - If the ignition transactions that contains the `set_pose` is processed together with some relative constraints from `differential` mode `Odometry2D` measurements (or equivalent), the state history cannot be updated.
    - Therefore, the state defaults to `0`/identity: https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L139-L150 https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L177 That state is used for the previous pose variable in the relative constraints when they're connected to the variable of the `set_pose` absolute constraint.
    - The solution consists on processing the ignition transaction individually. This way the state history is updated to the `set_pose` value when it receives the new updated graph in: https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L95

3. Call motion model generator with last stamp in order to update its state history.
    - When the history is empty or the `last_stamp` is newer than any stamp in the history, the motion model history is updated to include it in: https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_core/src/timestamp_manager.cpp#L154-L159
    - However, the `Unicycle2D` motion model state history is not updated.
    - As a consequence, the `set_pose` doesn't get any entry into the state history and it cannot be updated when a new graph is received in: https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L95
    - We can workaround this issue by calling the motion model `generator_` with the `last_stamp` as beginning and ending stamp. This will add a single entry because the predicted state will be the same as the beginning one: https://github.com/locusrobotics/fuse/blob/a1281b17a3c3b4c2b034a637ac1066615f167d7a/fuse_models/src/unicycle_2d.cpp#L234-L235 We don't need the constraint and variables generated, so we discard them. All we're looking for is an entry in the state history, so it can be updated when the graph is received.

### DIFF
--- a/fuse_core/include/fuse_core/transaction.h
+++ b/fuse_core/include/fuse_core/transaction.h
@@ -150,6 +150,13 @@ public:
   const_uuid_range removedVariables() const { return removed_variables_; }
 
   /**
+   * @brief Check if the transaction is empty, i.e. it has no added or removed constraints or variables
+   *
+   * @return  True if the transaction is empty, false otherwise
+   */
+  bool empty() const;
+
+  /**
    * @brief Add a timestamp to the "involved stamps" collection
    *
    * Duplicate timestamps will be ignored, so adding a stamp multiple times will have no effect.
--- a/fuse_core/include/fuse_core/transaction.h
+++ b/fuse_core/include/fuse_core/transaction.h
@@ -150,7 +150,8 @@ public:
   const_uuid_range removedVariables() const { return removed_variables_; }
 
   /**
-   * @brief Check if the transaction is empty, i.e. it has no added or removed constraints or variables
+   * @brief Check if the transaction is empty, i.e. it has no added or removed constraints or variables, and no involved
+   * stamps
    *
    * @return  True if the transaction is empty, false otherwise
    */
--- a/fuse_core/src/timestamp_manager.cpp
+++ b/fuse_core/src/timestamp_manager.cpp
@@ -156,6 +156,11 @@ void TimestampManager::query(
     // Insert the last timestamp into the motion model history, but with no constraints. The last entry in the motion
     // model history will always contain no constraints.
     motion_model_history_.emplace(last_stamp, MotionModelSegment());
+
+    // Call the motion model generator so it inserts the last timestamp into its state history.
+    std::vector<Constraint::SharedPtr> constraints;
+    std::vector<Variable::SharedPtr> variables;
+    generator_(last_stamp, last_stamp, constraints, variables);
   }
   // Purge any old entries from the motion model history
   purgeHistory();
--- a/fuse_core/src/transaction.cpp
+++ b/fuse_core/src/transaction.cpp
@@ -39,6 +39,7 @@
 #include <ros/time.h>
 
 #include <boost/iterator/transform_iterator.hpp>
+#include <boost/range/empty.hpp>
 
 #include <algorithm>
 #include <ostream>
@@ -125,6 +126,12 @@ Transaction::const_variable_range Transaction::addedVariables() const
   return const_variable_range(
     boost::make_transform_iterator(added_variables_.cbegin(), to_variable_ref),
     boost::make_transform_iterator(added_variables_.cend(), to_variable_ref));
+}
+
+bool Transaction::empty() const
+{
+  return boost::empty(added_variables_) && boost::empty(removed_variables_) && boost::empty(added_constraints_) &&
+         boost::empty(removed_constraints_);
 }
 
 void Transaction::addVariable(Variable::SharedPtr variable, bool overwrite)
--- a/fuse_core/src/transaction.cpp
+++ b/fuse_core/src/transaction.cpp
@@ -130,8 +130,8 @@ Transaction::const_variable_range Transaction::addedVariables() const
 
 bool Transaction::empty() const
 {
-  return boost::empty(added_variables_) && boost::empty(removed_variables_) && boost::empty(added_constraints_) &&
-         boost::empty(removed_constraints_);
+  return boost::empty(added_variables_) && boost::empty(removed_variables_) &&
+         boost::empty(added_constraints_) && boost::empty(removed_constraints_) && involved_stamps_.empty();
 }
 
 void Transaction::addVariable(Variable::SharedPtr variable, bool overwrite)
--- a/fuse_core/test/test_timestamp_manager.cpp
+++ b/fuse_core/test/test_timestamp_manager.cpp
@@ -542,7 +542,7 @@ TEST_F(TimestampManagerTestFixture, MultiSegment)
   ASSERT_EQ(0ul, generated_time_spans.size());
 }
 
-TEST_F(TimestampManagerTestFixture, MultiSegmentBeforBeginning)
+TEST_F(TimestampManagerTestFixture, MultiSegmentBeforeBeginning)
 {
   // Test:
   // Existing: |------111111112222222233333333-------> t
--- a/fuse_core/test/test_timestamp_manager.cpp
+++ b/fuse_core/test/test_timestamp_manager.cpp
@@ -112,9 +112,12 @@ TEST_F(TimestampManagerTestFixture, Empty)
   EXPECT_EQ(ros::Time(20, 0), *stamp_range_iter);
 
   // Verify the expected queries were performed
-  ASSERT_EQ(1ul, generated_time_spans.size());
+  // And additional time span is generated for the last stamp as beginning and ending stamp
+  ASSERT_EQ(2ul, generated_time_spans.size());
   EXPECT_EQ(ros::Time(10, 0), generated_time_spans[0].first);
   EXPECT_EQ(ros::Time(20, 0), generated_time_spans[0].second);
+  EXPECT_EQ(ros::Time(20, 0), generated_time_spans[1].first);
+  EXPECT_EQ(generated_time_spans[1].first, generated_time_spans[1].second);
 }
 
 TEST_F(TimestampManagerTestFixture, Exceptions)
@@ -420,9 +423,12 @@ TEST_F(TimestampManagerTestFixture, AfterEndAligned)
   EXPECT_EQ(ros::Time(45, 0), *stamp_range_iter);
 
   // Verify the expected new motion model segments were generated
-  ASSERT_EQ(1ul, generated_time_spans.size());
+  // And additional time span is generated for the last stamp as beginning and ending stamp
+  ASSERT_EQ(2ul, generated_time_spans.size());
   EXPECT_EQ(ros::Time(40, 0), generated_time_spans[0].first);
   EXPECT_EQ(ros::Time(45, 0), generated_time_spans[0].second);
+  EXPECT_EQ(ros::Time(45, 0), generated_time_spans[1].first);
+  EXPECT_EQ(generated_time_spans[1].first, generated_time_spans[1].second);
 }
 
 TEST_F(TimestampManagerTestFixture, AfterEndUnaligned)
@@ -455,11 +461,14 @@ TEST_F(TimestampManagerTestFixture, AfterEndUnaligned)
   EXPECT_EQ(ros::Time(45, 0), *stamp_range_iter);
 
   // Verify the expected new motion model segments were generated
-  ASSERT_EQ(2ul, generated_time_spans.size());
+  // And additional time span is generated for the last stamp as beginning and ending stamp
+  ASSERT_EQ(3ul, generated_time_spans.size());
   EXPECT_EQ(ros::Time(40, 0), generated_time_spans[0].first);
   EXPECT_EQ(ros::Time(42, 0), generated_time_spans[0].second);
   EXPECT_EQ(ros::Time(42, 0), generated_time_spans[1].first);
   EXPECT_EQ(ros::Time(45, 0), generated_time_spans[1].second);
+  EXPECT_EQ(ros::Time(45, 0), generated_time_spans[2].first);
+  EXPECT_EQ(generated_time_spans[2].first, generated_time_spans[2].second);
 }
 
 TEST_F(TimestampManagerTestFixture, AfterEndOverlap)
@@ -492,13 +501,16 @@ TEST_F(TimestampManagerTestFixture, AfterEndOverlap)
   EXPECT_EQ(ros::Time(45, 0), *stamp_range_iter);
 
   // Verify the expected new motion model segments were generated
-  ASSERT_EQ(3ul, generated_time_spans.size());
+  // And additional time span is generated for the last stamp as beginning and ending stamp
+  ASSERT_EQ(4ul, generated_time_spans.size());
   EXPECT_EQ(ros::Time(30, 0), generated_time_spans[0].first);
   EXPECT_EQ(ros::Time(35, 0), generated_time_spans[0].second);
   EXPECT_EQ(ros::Time(35, 0), generated_time_spans[1].first);
   EXPECT_EQ(ros::Time(40, 0), generated_time_spans[1].second);
   EXPECT_EQ(ros::Time(40, 0), generated_time_spans[2].first);
   EXPECT_EQ(ros::Time(45, 0), generated_time_spans[2].second);
+  EXPECT_EQ(ros::Time(45, 0), generated_time_spans[3].first);
+  EXPECT_EQ(generated_time_spans[3].first, generated_time_spans[3].second);
 }
 
 TEST_F(TimestampManagerTestFixture, MultiSegment)
@@ -591,9 +603,12 @@ TEST_F(TimestampManagerTestFixture, MultiSegmentPastEnd)
   EXPECT_EQ(ros::Time(45, 0), *stamp_range_iter);
 
   // Verify the expected new motion model segments were generated
-  ASSERT_EQ(1ul, generated_time_spans.size());
+  // And additional time span is generated for the last stamp as beginning and ending stamp
+  ASSERT_EQ(2ul, generated_time_spans.size());
   EXPECT_EQ(ros::Time(40, 0), generated_time_spans[0].first);
   EXPECT_EQ(ros::Time(45, 0), generated_time_spans[0].second);
+  EXPECT_EQ(ros::Time(45, 0), generated_time_spans[1].first);
+  EXPECT_EQ(generated_time_spans[1].first, generated_time_spans[1].second);
 }
 
 TEST_F(TimestampManagerTestFixture, MultiSegmentPastBothEnds)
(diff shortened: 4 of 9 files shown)


efernandez:
@svwilliams I was thinking of contributing a unit test or rostest for this, but TBH I'm not sure how to create one for this. To start with, I'm not sure where it should go, or whether it should be a unit test or a rostest. If you give me some guidance, I might be able to contribute one, although I'm quite busy these days.

FYI @ayrton04 

BTW For some reason I still can't add any reviewers to the PRs I send here.

fuse_core/test/test_timestamp_manager.cpp
```diff
@@ -112,9 +112,12 @@ TEST_F(TimestampManagerTestFixture, Empty)
   EXPECT_EQ(ros::Time(20, 0), *stamp_range_iter);
 
   // Verify the expected queries were performed
-  ASSERT_EQ(1ul, generated_time_spans.size());
```
efernandez, line 115: I have to update the tests here because now `TimestampManager::query` calls the `generator_` for the `last_stamp`, which adds a `(last_stamp, last_stamp)` time span add the end.

Note that in the actual generator in `Unicycle2D` the state history is a map with the time stamp as a key, so only a single entry is added per stamp.

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -255,12 +256,58 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction)
...
+      erase(pending_transactions_, transaction_rbegin);
+      return;
+    }
+    else
+    {
+      // The motion model processing failed. When this happens to an ignition sensor transaction there is no point on
+      // trying again next time, so we ignore this transaction.
+      ROS_ERROR_STREAM("The queued ignition transaction with timestamp " << element.stamp() << " from sensor " <<
+                       element.sensor_name << " could not be processed. Ignoring this transaction.");
+      erase(pending_transactions_, transaction_rbegin);
```
svwilliams: Since the ignition transaction was never applied, I feel like we should also set `started_` to false. That way we don't start using transactions on an improperly initialized system.

There is a nasty edge-case in here as well, when the pending transactions include a second ignition transaction. To be very thorough we would need to crawl through any remaining transactions and check if it is an ignition source.
* If so, then purge the transactions before it and leave with `started_ = true` and `ignited_ = true`. This would trigger an ignition attempt during the next optimization cycle.
* If not, then leave the queue as-is, and exit with `started_ = false` and `ignited_ = false`. The next received ignition transaction will then trigger an ignition attempt.

svwilliams (commented):
I have created a unit test to check for this improperly initialized motion model bug. I'll get that cleaned up and submitted tomorrow. I have already tested that it fails with the current code, and passes with this fix.

fuse_optimizers/src/fixed_lag_smoother.cpp
efernandez: First of all, when I wrote this I obviously thought of the edge cases, but I didn't test most of them. :sweat_smile: 

> Since the ignition transaction was never applied, I feel like we should also set `started_` to false. That way we don't start using transactions on an improperly initialized system.

The reason I didn't do that is because `processQueue` is called from the optimization loop, and it looks to me it's too late to stop processing things.

Here it looks like we should do:
``` c++
started_ = false;
pending_transactions_.clear();
```

With this we'll end up with an empty `new_transaction` in the optimization loop: https://github.com/locusrobotics/fuse/blob/devel/fuse_optimizers/src/fixed_lag_smoother.cpp#L198

Maybe there we could make `processQueue` return a `bool` to indicate if there's something to process, or we could just check if `new_transaction` has any constraints or is empty. We could also check the `started_` flag. It looks like any of those options would work.
My point here is that I think we also need to add some changes to the optimization loop, to abort processing the queue. :thinking: 

> There is a nasty edge-case in here as well, when the pending transactions include a second ignition transaction. To be very thorough we would need to crawl through any remaining transactions and check if it is an ignition source.

I thought about this, and I came to the **wrong** conclusion that a second ignition transaction wouldn't produce an ignition cycle. The first one would set `started_ = true`, so the second one wouldn't be able to enter https://github.com/locusrobotics/fuse/blob/devel/fuse_optimizers/src/fixed_lag_smoother.cpp#L367. I was thinking of the case where an ignition sensor model is used both to generate ignition and non-ignition transactions.

But I think you're right. If the first ignition fails, then we need to purge all transactions starting from the first ignition till the next ignition, if any. If there's a second one, then we can either process it in this cycle, or leave for the next one. I believe it's possible to process it in this cycle, but it's likely cleaner to simply purge things in this cycle and process in the next one, so we only have to do this (instead of the code snippet above in this comment):

``` c++
// Purge all transaction after the ignition one that just failed. But if we find another ignition
// transaction, we schedule it to be processed in the next optimization cycle
const auto pending_ignition_transaction_iter =
    std::find_if(pending_transactions_.rbegin(), pending_transactions_.rend(),
        [this](const auto& element)
        {
          return std::binary_search(params_.ignition_sensors.begin(),
                                    params_.ignition_sensors.end(),
                                    element.sensor_name);
        });

if (pending_ignition_transaction_iter == pending_transactions_.rend())
{
   pending_transactions_.clear();
   started_ = false;
}
else
{
  pending_transactions_.erase(pending_ignition_transaction_iter.base(), pending_transactions_.rbegin().base());
  ignited_ = true;
  return;
}
```

I did some basic validation of this code here: https://wandbox.org/permlink/PGaY4yBQ3V5GjQNz

And, as I mentioned before, we probably need to do changes in the optimization loop, so it properly handles **empty** transactions, if it currently doesn't.

Is this more or less what you're thinking about?

I can update the PR with this changes, but since you already have a unit test, I guess you can do that yourself, or I wait for you to push the unit test to this or another PR. Thanks for adding a unit test, btw! :smiley:

svwilliams: Yeah, that's the kind of thing I was thinking about.

I'm not sure you need to clear all the pending transactions in the first `if`. When a new ignition transaction is received, the pending transactions will get cleared as needed. I would think removing the failed ignition transaction would be sufficient.

And then handling the empty transaction case, which I can do in a separate PR if you can add the above block to this PR.

efernandez: > I'm not sure you need to clear all the pending transactions in the first if. When a new ignition transaction is received, the pending transactions will get cleared as needed. I would think removing the failed ignition transaction would be sufficient.

Sure, instead of clearing the pending transactions, I could just `return;`

Just saw your message. Give me a minute to update this PR. Thanks.

efernandez: Updated.

However, bear in mind I haven't tested the edges cases this logic handles.

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -196,6 +197,12 @@ void FixedLagSmoother::optimizationLoop()
       //         queue and obtaining the lock for the graph. But we have now obtained two different locks. If we are
       //         not extremely careful, we could get a deadlock.
       processQueue(*new_transaction);
+      // Skip this optimization cycle if the transaction is empty because something failed while processing the pending
```
efernandez, line 200: @svwilliams I've added this as a tentative way to abort running the optimization cycle if the transaction is empty.

It looks like most of the methods handle empty transactions gracefully, except for the graph `update` and `optimize` methods, just looking at the code though.

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -255,12 +262,84 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction)
...
+        erase(pending_transactions_, transaction_rbegin);
+      }
+      else
+      {
+        // The motion model processing failed. When this happens to an ignition sensor transaction there is no point on
+        // trying again next time, so we ignore this transaction.
+        ROS_ERROR_STREAM("The queued ignition transaction with timestamp " << element.stamp() << " from sensor " <<
+                         element.sensor_name << " could not be processed. Ignoring this ignition transaction.");
+
+        // Purge all transactions after the ignition one that just failed. But if we find another ignition transaction,
```
svwilliams: Don't we need to erase the failed transaction first? Before the search for the next ignition transaction?
```
erase(pending_transactions_, transaction_rbegin);
```

fuse_optimizers/src/fixed_lag_smoother.cpp
svwilliams: This works as well as anything.

I would think the `graph.update()` call would be fine (the for-loops should work just fine with empty ranges), but I'm unclear what the `ceres::Solve()` call would do if the Problem object is empty. Best to avoid the issue.

svwilliams (commented):
One final question

fuse_optimizers/src/fixed_lag_smoother.cpp
efernandez: Oh, yes. I forgot that. :man_facepalming:

efernandez: Fixed.

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -255,12 +262,86 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction)
...
+          started_ = false;
+        }
+        else
+        {
+          // Erase all transactions before the other ignition transaction pending. This other ignition transaction will
+          // be processed in the next optimization cycle.
+          pending_transactions_.erase(pending_ignition_transaction_iter.base(), pending_transactions_.rbegin().base());
+          ignited_ = true;
+        }
+        erase(pending_transactions_, transaction_rbegin);
```
svwilliams: And this one should go away?
I should have mentioned previously you were deleting the failed transaction after the search instead of before the search.

efernandez: Fixed. I didn't look that far down. At the end I didn't forgot it, I just left it in the wrong spot somehow. :laughing:

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -255,12 +262,85 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction)
...
+        else
+        {
+          // Erase all transactions before the other ignition transaction pending. This other ignition transaction will
+          // be processed in the next optimization cycle.
+          pending_transactions_.erase(pending_ignition_transaction_iter.base(), pending_transactions_.rbegin().base());
+          ignited_ = true;
+        }
+      }
+
+      return;
```
efernandez, line 342: Maybe this deserves a comment. :thinking:

fuse_core/src/timestamp_manager.cpp
```diff
@@ -156,6 +156,11 @@ void TimestampManager::query(
     // Insert the last timestamp into the motion model history, but with no constraints. The last entry in the motion
     // model history will always contain no constraints.
     motion_model_history_.emplace(last_stamp, MotionModelSegment());
+
+    // Call the motion model generator so it inserts the last timestamp into its state history.
```
efernandez, line 160: I wonder if something cleaner could be done, but I guess for that we'll have to make this class know about the motion model. :thinking:

fuse_core/include/fuse_core/transaction.h
```diff
@@ -149,6 +149,13 @@ class Transaction
    */
   const_uuid_range removedVariables() const { return removed_variables_; }
 
+  /**
+   * @brief Check if the transaction is empty, i.e. it has no added or removed constraints or variables
+   *
+   * @return  True if the transaction is empty, false otherwise
+   */
+  bool empty() const;
```
efernandez, line 158: This probably deserves a unit test :smiley:

efernandez (commented):
@svwilliams I addressed all your comments, but I also added three other comments to things that you might want to review again.

fuse_core/src/transaction.cpp
```diff
@@ -127,6 +128,12 @@ Transaction::const_variable_range Transaction::addedVariables() const
     boost::make_transform_iterator(added_variables_.cend(), to_variable_ref));
 }
 
+bool Transaction::empty() const
+{
+  return boost::empty(added_variables_) && boost::empty(removed_variables_) && boost::empty(added_constraints_) &&
+         boost::empty(removed_constraints_);
```
svwilliams: I'm unsure if this check should include the `involved_stamps_` collection or not.

svwilliams (approved):

fuse_core/src/transaction.cpp
efernandez: I believe the `involved_stamps_` should always be empty when the others are, but I don't mind adding `&& involved_stamps_.empty()`. Indeed, since `addInvolvedStamp` is used for that, it's probably better to check that.

Either way, it doesn't impact this PR.

efernandez: Added the `involved_stamps_`  check to the `empty()` method, and also added a unit test for `empty()`.

fuse_core/include/fuse_core/transaction.h
efernandez: Unit test added

fuse_core/src/timestamp_manager.cpp
efernandez: Here I just want to hear your thoughts, but this is certainly something for a another PR, or we just go with this for now.

I addressed all other conversations and the PR is ready to merge.

fuse_optimizers/src/fixed_lag_smoother.cpp
efernandez: Comment added

ayrton04 (approved):

fuse_core/src/timestamp_manager.cpp
svwilliams: Admittedly, I'm not in love with calling the motion model generator with these degenerate time ranges. Particularly since this is only an issue for initializing the first state, but the solution adds an additional call to every transaction.

One option I thought about was modifying the motion model implementation here:
https://github.com/locusrobotics/fuse/blob/devel/fuse_models/src/unicycle_2d.cpp#L106

At this point, we have access to the input transaction with all of the timestamps and added variables. This would allow us to steal the initial values of the variables added by the ignition sensor, instead of using this call with a degenerate time range to insert the initial state. The implications of that get tangled. In theory, that would mean the motion model could initialize itself properly from the very beginning, and does not need a round trip to the optimizer and back to work correctly. Thus, the requirement to process the ignition constraint in its own transaction falls away. But if we relax that requirement, we are relying on the incoming transaction to have valid initial variable values. Right now, when transactions are combined, the variable values from the first transaction are retained. So as long as the first transaction is the ignition transaction, then that should work. But it felt a bit brittle. It also means we are moving coding complexity out of the Optimizer class and into the MotionModel class. And I wasn't sure that was a good trade.

Regardless, I would want to test that option more before I decided it was a good idea. And it could be that processing the ignition sensor in a separate transaction along with modifications to the motion model would be a good combination.

This current PR fixes the issue now, so I'm all for getting it merged. We can always change things up later.

efernandez: I'm 99% this is not only an issue for initializing the first state, if you mean the very first entry in the history. With relative constraints, the second variable in the constraint is usually (if not always) entering here, and therefore it was also impacted. I might be wrong though, because I was printing things here and getting a ton of spam. A unit test that exercises that case would help. Not sure if you already cover that case in your unit test in https://github.com/locusrobotics/fuse/pull/158.

Anyway, I think we can merge this an continue the discussion there.

svwilliams: At least with my understanding of the problem, I only see two conditions that cause problems:
1. If the state history is empty, then the lookup into the state history for the first involved timestamp fails. When that fails, it uses zeros as the initial state.
2. If a constraint arrives out-of-order and/or references a timestamp before the beginning of the state history, then the lookup into the state history for that timestamp will also fail and result in a zero state estimate for that state.

Generally speaking, relative pose constraints should work just fine after the state history has been properly initialized. The motion model will look up a state at or before each involved timestamp, and use the unicycle model to predict the value at those timestamps. However, if the relative pose constraints regularly include states older than the lag, then you enter condition (2) above.

Regardless, I think this PR should be merged as-is, and we will pick up this conversation over on the unit test PR #158.

efernandez: :+1:

PR #156, opened 2020-03-11
Print state history

efernandez:
This adds support to print the state history from the `Unicycle2D` motion model. Not sure if this is worth a PR, but it was useful while debugging things for https://github.com/locusrobotics/fuse/pull/154.

I'm happy to close, merge as is, or extend it to cover all the attributes in the `Unicycle2D` class.

TBH, printing the entire `state_history_` map is quite verbose, but that's better than nothing when trying to debug things. I'm all ears for suggestions on how to better debug the logic that handles the state and motion model history in `Unicycle2D` and the `TimeManager`. :smiley: 

Note this only contributes the printing support. Nothing is really printing anywhere. That part was added locally while debugging.


efernandez:

svwilliams (approved):

svwilliams:
I'll go through #154 in detail later today. Maybe I'll have some better insights into how to debug things then. But I have nothing against adding a print function regardless.

ayrton04 (approved):

PR #158, opened 2020-03-17
Added unit test to illustrate variable initialization bug

svwilliams:
There is an issue with how variables are initialized in the motion model. This PR creates a unit test that adds an initial constraint to the graph and grabs to the optimized output. Because the test has the optimizer limited to 0 iterations, the "optimized" value is actually just the initial value provided by the motion model.

This issue is fixed by #154.

### DIFF
--- a/fuse_optimizers/test/test_fixed_lag_ignition.cpp
+++ b/fuse_optimizers/test/test_fixed_lag_ignition.cpp
@@ -102,7 +102,7 @@ TEST(FixedLagIgnition, SetInitialState)
 
   // Wait for the optimizer to publish the first pose
   auto odom_msg = ros::topic::waitForMessage<nav_msgs::Odometry>("/odom", ros::Duration(1.0));
-  ASSERT_TRUE(odom_msg);
+  ASSERT_TRUE(static_cast<bool>(odom_msg));
 
   // The optimizer is configured for 0 iterations, so it should return the initial variable values
   // If we did our job correctly, the initial variable values should be the same as the service call state, give or
--- a/fuse_optimizers/test/test_fixed_lag_ignition.cpp
+++ b/fuse_optimizers/test/test_fixed_lag_ignition.cpp
@@ -67,9 +67,14 @@ TEST(FixedLagIgnition, SetInitialState)
 
   // The 'set_pose' service call triggers all of the sensors to resubscribe to their topics.
   // I need to wait for those subscribers to be ready before sending them sensor data.
-  // I'd love a ros::topic::waitForSubscriber() method, but that doesn't exist.
-  // I'm not sure what else to do, other than "wait a bit".
-  ros::Duration(0.5).sleep();
+  ros::Time timeout = ros::Time::now() + ros::Duration(1.0);
+  while ((pose_publisher.getNumSubscribers() == 0u) &&
+         (ros::Time::now() < timeout))
+  {
+    std::cout << "waiting..." << std::endl;
+    ros::Duration(0.1).sleep();
+  }
+  ASSERT_GT(pose_publisher.getNumSubscribers(), 0u);
 
   // Publish a relative pose
   auto pose_msg1 = geometry_msgs::PoseWithCovarianceStamped();
--- a/fuse_optimizers/test/test_fixed_lag_ignition.cpp
+++ b/fuse_optimizers/test/test_fixed_lag_ignition.cpp
@@ -68,13 +68,13 @@ TEST(FixedLagIgnition, SetInitialState)
   // The 'set_pose' service call triggers all of the sensors to resubscribe to their topics.
   // I need to wait for those subscribers to be ready before sending them sensor data.
   ros::Time timeout = ros::Time::now() + ros::Duration(1.0);
-  while ((pose_publisher.getNumSubscribers() == 0u) &&
+  while ((pose_publisher.getNumSubscribers() < 1u) &&
          (ros::Time::now() < timeout))
   {
     std::cout << "waiting..." << std::endl;
     ros::Duration(0.1).sleep();
   }
-  ASSERT_GT(pose_publisher.getNumSubscribers(), 0u);
+  ASSERT_GE(pose_publisher.getNumSubscribers(), 1u);
 
   // Publish a relative pose
   auto pose_msg1 = geometry_msgs::PoseWithCovarianceStamped();
--- a/fuse_optimizers/test/test_fixed_lag_ignition.cpp
+++ b/fuse_optimizers/test/test_fixed_lag_ignition.cpp
@@ -45,7 +45,6 @@ TEST(FixedLagIgnition, SetInitialState)
   auto pose_publisher = node_handle.advertise<geometry_msgs::PoseWithCovarianceStamped>("/pose", 1);
 
   // Wait for the optimizer to be ready
-  ASSERT_TRUE(ros::service::waitForService("/fixed_lag/reset", ros::Duration(1.0)));
   ASSERT_TRUE(ros::service::waitForService("/fixed_lag/set_pose", ros::Duration(1.0)));
 
   // Set the initial pose to something near zero
@@ -64,9 +63,12 @@ TEST(FixedLagIgnition, SetInitialState)
   req.pose.pose.covariance[35] = 1.0;
   fuse_models::SetPose::Response res;
   ros::service::call("/fixed_lag/set_pose", req, res);
+  ASSERT_TRUE(res.success);
 
-  // I need to wait for the subscriber to be ready. I don't know of a better way.
-  // There is no ros::topic::waitForSubscriber() method.
+  // The 'set_pose' service call triggers all of the sensors to resubscribe to their topics.
+  // I need to wait for those subscribers to be ready before sending them sensor data.
+  // I'd love a ros::topic::waitForSubscriber() method, but that doesn't exist.
+  // I'm not sure what else to do, other than "wait a bit".
   ros::Duration(0.5).sleep();
 
   // Publish a relative pose
@@ -101,7 +103,7 @@ TEST(FixedLagIgnition, SetInitialState)
   pose_publisher.publish(pose_msg2);
 
   // Wait for the optimizer to publish the first pose
-  auto odom_msg = ros::topic::waitForMessage<nav_msgs::Odometry>("/odom", ros::Duration(1.0));
+  auto odom_msg = ros::topic::waitForMessage<nav_msgs::Odometry>("/odom", ros::Duration(1.5));
   ASSERT_TRUE(static_cast<bool>(odom_msg));
 
   // The optimizer is configured for 0 iterations, so it should return the initial variable values
--- a/fuse_optimizers/test/test_fixed_lag_ignition.cpp
+++ b/fuse_optimizers/test/test_fixed_lag_ignition.cpp
@@ -1,0 +1,124 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2020, Locus Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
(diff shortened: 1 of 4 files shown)


fuse_optimizers/test/test_fixed_lag_ignition.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+
+TEST(FixedLagIgnition, SetInitialState)
+{
+  auto node_handle = ros::NodeHandle();
+  auto pose_publisher = node_handle.advertise<geometry_msgs::PoseWithCovarianceStamped>("/pose", 1);
+
+  // Wait for the optimizer to be ready
+  ASSERT_TRUE(ros::service::waitForService("/fixed_lag/reset", ros::Duration(1.0)));
+  ASSERT_TRUE(ros::service::waitForService("/fixed_lag/set_pose", ros::Duration(1.0)));
```
needphpsql: Is 1 second enough for timeout?

fuse_optimizers/test/test_fixed_lag_ignition.cpp
```diff
@@ -0,0 +1,124 @@
...
+  req.pose.pose.pose.orientation.w = 1.0;
+  req.pose.pose.covariance[0] = 1.0;
+  req.pose.pose.covariance[7] = 1.0;
+  req.pose.pose.covariance[35] = 1.0;
+  fuse_models::SetPose::Response res;
+  ros::service::call("/fixed_lag/set_pose", req, res);
+
+  // I need to wait for the subscriber to be ready. I don't know of a better way.
+  // There is no ros::topic::waitForSubscriber() method.
+  ros::Duration(0.5).sleep();
```
needphpsql: There is a return value to handle the service call. 

> See also: ros::service::call() API docs, ros::NodeHandle::serviceClient() API docs, ros::ServiceClient API docs, ros::service namespace API docs

There are two ways of calling a service in roscpp, the "handle" way and the "bare" way.

Bare

    For convenience, the ros::service namespace provides a call method, which does not require creation of a NodeHandle:
    Toggle line numbers

       1 //remember that ros::init(...) must have been called
       2 my_package::Foo foo;
       3 ...
       4 if (ros::service::call("my_service_name", foo))
       5 {
       6   ...
       7 }

    The ros::service namespace also provides some convenience functions such as exists() and waitForService(). See the ros::service namespace API docs for more information. 

Handle

    The handle way works more similar to how the rest of roscpp works, in that you are returned a ros::ServiceClient which is then used to call the service:

    ros::ServiceClient client = nh.serviceClient<my_package::Foo>("my_service_name");
    my_package::Foo foo;
    ...
    if (client.call(foo))
    {
      ...
    }

    ros::ServiceClient also has a number of other useful methods suchs as exists() and waitForExistence(). See the ros::ServiceClient API docs for more information.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
```diff
@@ -0,0 +1,124 @@
...
+  pose_msg2.pose.pose.orientation.y = 0.0;
+  pose_msg2.pose.pose.orientation.z = 0.5000;
+  pose_msg2.pose.pose.orientation.w = 0.8660;
+  pose_msg2.pose.covariance[0] = 1.0;
+  pose_msg2.pose.covariance[7] = 1.0;
+  pose_msg2.pose.covariance[35] = 1.0;
+  pose_publisher.publish(pose_msg2);
+
+  // Wait for the optimizer to publish the first pose
+  auto odom_msg = ros::topic::waitForMessage<nav_msgs::Odometry>("/odom", ros::Duration(1.0));
```
needphpsql: Same question for 1 second timeout, is it enough or should wait forever?

needphpsql (commented):
add some comments.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
svwilliams: The 1 sec timeout has been enough on my machine. We will see if it is sufficient for the CI machine. If CI fails at that point, I can bump it up a bit. But we do want to the unit test to fail eventually, so an unbounded wait would not be a great idea. (There is a fail-safe timeout for unit tests somewhere, but I believe it is large).

fuse_optimizers/test/test_fixed_lag_ignition.cpp
svwilliams: Right. So the issue is that inside of the service call, a topic subscriber is created. And, for whatever reason, the subscriber does not immediately start receiving messages. It takes an annoyingly long time. I originally set the wait time to 0.1s and that did not work. It was not until the 0.25-0.3s range that the subscriber started receiving messages. What I'd like is a "wait for subscriber" call, similar to the "wait for service" call. But that doesn't exist. There is probably a more clever way to setup this test, but I'm not sure what it is.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
svwilliams: The fixed-lag smoother is configured to update every 1.0s seconds. So maybe adding a bit more time would be safer. But if I don't receive a message, that should be considered a failure case.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
needphpsql: Is it possible to force the service call doesn't return until the subscriber receives some message or timeout inside the service all? In this way, there is no external wait and sleep required.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
svwilliams: Ideally ROS would not return from the `node_handle.subscribe()` call until the subscriber was ready. But there's nothing I can do about that. 🙂

There is a complex set of interactions between the "set_pose" service call and the initialization of the subscriber. That makes waiting for a message awkward. Even if I could do as you suggest, it would be self defeating. Waiting to publish a message until a message has been received is somewhat self-defeating.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
svwilliams: But I can use `while (publisher.getNumSubscribers() == 0)`

fuse_optimizers/test/test_fixed_lag_ignition.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  // The 'set_pose' service call triggers all of the sensors to resubscribe to their topics.
+  // I need to wait for those subscribers to be ready before sending them sensor data.
+  ros::Time timeout = ros::Time::now() + ros::Duration(1.0);
+  while ((pose_publisher.getNumSubscribers() == 0u) &&
+         (ros::Time::now() < timeout))
+  {
+    std::cout << "waiting..." << std::endl;
+    ros::Duration(0.1).sleep();
+  }
```
needphpsql: Should be (pose_publisher.getNumSubscribers() < N_sensors)?

needphpsql (approved):

ayrton04 (approved):

fuse_optimizers/test/test_fixed_lag_ignition.cpp
```diff
@@ -0,0 +1,131 @@
...
+  req.pose.pose.covariance[0] = 1.0;
+  req.pose.pose.covariance[7] = 1.0;
+  req.pose.pose.covariance[35] = 1.0;
+  fuse_models::SetPose::Response res;
+  ros::service::call("/fixed_lag/set_pose", req, res);
+  ASSERT_TRUE(res.success);
+
+  // The 'set_pose' service call triggers all of the sensors to resubscribe to their topics.
+  // I need to wait for those subscribers to be ready before sending them sensor data.
+  ros::Time timeout = ros::Time::now() + ros::Duration(1.0);
```
efernandez: You probably want to call `ros::Time::isValid()` before calling `ros::Time::now()`, or maybe even before the `waitForService` above, or even before any test starts.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
```diff
@@ -0,0 +1,131 @@
...
+  req.pose.pose.pose.orientation.z = 0.0;
+  req.pose.pose.pose.orientation.w = 1.0;
+  req.pose.pose.covariance[0] = 1.0;
+  req.pose.pose.covariance[7] = 1.0;
+  req.pose.pose.covariance[35] = 1.0;
+  fuse_models::SetPose::Response res;
+  ros::service::call("/fixed_lag/set_pose", req, res);
+  ASSERT_TRUE(res.success);
+
+  // The 'set_pose' service call triggers all of the sensors to resubscribe to their topics.
```
efernandez, line 72: What if you make the publisher latch?: `auto pose_publisher = node_handle.advertise<geometry_msgs::PoseWithCovarianceStamped>("/pose", 1, true);`

fuse_optimizers/test/test_fixed_lag_ignition.cpp
```diff
@@ -0,0 +1,131 @@
...
+  fuse_models::SetPose::Request req;
+  req.pose.header.frame_id = "map";
+  req.pose.header.stamp = ros::Time(1, 0);
+  req.pose.pose.pose.position.x = 100.1;
+  req.pose.pose.pose.position.y = 100.2;
+  req.pose.pose.pose.position.z = 0.0;
+  req.pose.pose.pose.orientation.x = 0.0;
+  req.pose.pose.pose.orientation.y = 0.0;
+  req.pose.pose.pose.orientation.z = 0.0;
+  req.pose.pose.pose.orientation.w = 1.0;
```
efernandez: Maybe also try a non-trivial orientation and check it's the one used.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include <gtest/gtest.h>
+
+
+TEST(FixedLagIgnition, SetInitialState)
+{
+  auto node_handle = ros::NodeHandle();
+  auto pose_publisher = node_handle.advertise<geometry_msgs::PoseWithCovarianceStamped>("/pose", 1);
+
+  // Wait for the optimizer to be ready
+  ASSERT_TRUE(ros::service::waitForService("/fixed_lag/set_pose", ros::Duration(1.0)));
```
efernandez, line 51: I think you could also use the `set_pose` topic and create a latch publisher here to publish the initial pose to set.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
svwilliams: Added a `ros::Time::waitForValid()` assert.
Switching to WallTime for the unit test timeout. That's probably a better fit anyway.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
svwilliams: To be explicit, I changed the topic name to "relative_pose". This is the measurement source for a relative pose sensor. As such, I need to publish two pose messages so the sensor can compute the pose change. Consequently latching the topic does not help; I need the sensor to receive both messages. The "wait for subscribers" method seems to handle the issue well enough.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
svwilliams: Changed the initial orientation and added checks on `z` and `w`.

svwilliams:
Referencing a conversation from #154

> At least with my understanding of the problem, I only see two conditions that cause problems:
>
> 1. If the state history is empty, then the lookup into the state history for the first involved timestamp fails. When that fails, it uses zeros as the initial state.
> 2. If a constraint arrives out-of-order and/or references a timestamp before the beginning of the state history, then the lookup into the state history for that timestamp will also fail and result in a zero state estimate for that state.
>
> Generally speaking, relative pose constraints should work just fine after the state history has been properly initialized. The motion model will look up a state at or before each involved timestamp, and use the unicycle model to predict the value at those timestamps. However, if the relative pose constraints regularly include states older than the lag, then you enter condition (2) above.

This unit test initializes the state far from zero so we can easily detect proper initialization of the first state. The test then adds a relative pose constraint _after_ the initialization time. This is what I would consider "normal" operation. It tests "condition 1" from the above quote, but it does not test "condition 2".

Note that this test was failing before PR #154 was merged.

I can try adding a second test the creates a relative pose that spans a time before the system was "ignited"...but I'm not exactly sure what will happen. I suppose what should happen is the measurement should get purged as it involves times before the beginning of the lag window.

efernandez:
> I can try adding a second test the creates a relative pose that spans a time before the system was "ignited"...but I'm not exactly sure what will happen. I suppose what should happen is the measurement should get purged as it involves times before the beginning of the lag window.

I can try to reproduce the issue I experienced and report back here, so it's clear what was failing with my configuration, other than the "condition 1" already covered by the unit test provided here. I'm not completely sure it was "condition 2", but I'll have to check again.

I'm a bit busy atm, but I'll try to give you an answer here by next week. I'm not saying you wait for that before merging this PR though. It can be merged as is now, but I'd still use it to continue this conversation. :smiley:

efernandez (approved):

PR #159, opened 2020-03-18
Patch Tukey loss for Ceres < 2.0.0

efernandez:
https://ceres-solver-review.googlesource.com/c/ceres-solver/+/16700 has been merged (see https://github.com/ceres-solver/ceres-solver/commit/6da364713f5b78ddf15b0e0ad92c76362c7c7683), so now the `TukeyLoss` won't need any workaround once Ceres 2.0.0 is release. You could also try building it from source and installing it in your system, as I've done.

This PR adds a workaround for `TukeyLoss` if Ceres < 2.0.0, updates the comments and unit tests.

### DIFF
--- a/fuse_loss/src/tukey_loss.cpp
+++ b/fuse_loss/src/tukey_loss.cpp
@@ -33,7 +33,7 @@
  */
 #include <fuse_loss/tukey_loss.h>
 
-#include <fuse_core/ceres_options.h>
+#include <fuse_core/ceres_macros.h>
 
 #include <pluginlib/class_list_macros.h>
 #include <ros/node_handle.h>
--- a/fuse_loss/src/tukey_loss.cpp
+++ b/fuse_loss/src/tukey_loss.cpp
@@ -33,6 +33,8 @@
  */
 #include <fuse_loss/tukey_loss.h>
 
+#include <fuse_core/ceres_options.h>
+
 #include <pluginlib/class_list_macros.h>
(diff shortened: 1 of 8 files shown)


efernandez:

svwilliams (approved):

fuse_loss/src/tukey_loss.cpp
```diff
@@ -33,6 +33,8 @@
  */
 #include <fuse_loss/tukey_loss.h>
 
+#include <fuse_core/ceres_options.h>
```
efernandez: This headers has the `CERES_VERSION_AT_LEAST` macro, but also a bunch of Ceres `Options` classes that we don't need here.

@svwilliams I wonder if I should create a new header with the macros only, something like: `fuse_core/ceres_macros.h` :thinking:

svwilliams: It might be nice to separate out the macros. As it is, you are pulling in some ROS dependencies that are not technically required. That said, I'm not too concerned about it right now. This is a ROS package that uses Ceres; both those dependencies will be available.

efernandez: Just added a new `fuse_core/ceres_macros.h` header. Not a big deal, and it reduces "surprise" when someone sees the header that is including here, because we aren't using any Ceres `Options` here.

ayrton04 (approved):

PR #160, opened 2020-03-19
Fix constness to match usage

svwilliams:
From a usage standpoint, the 'element' variable is getting modified and should not be const. The const was not causing compilation issues before because of some pointer indirection.

### DIFF
--- a/fuse_optimizers/src/fixed_lag_smoother.cpp
+++ b/fuse_optimizers/src/fixed_lag_smoother.cpp
@@ -287,7 +287,7 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction)
     ignited_ = false;
 
     const auto transaction_rbegin = pending_transactions_.rbegin();
-    const auto& element = *transaction_rbegin;
+    auto& element = *transaction_rbegin;
     if (!std::binary_search(params_.ignition_sensors.begin(), params_.ignition_sensors.end(), element.sensor_name))
     {
       // We just started, but the oldest transaction is not from an ignition sensor. We will still process the


fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -287,7 +287,7 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction)
     ignited_ = false;
 
     const auto transaction_rbegin = pending_transactions_.rbegin();
-    const auto& element = *transaction_rbegin;
+    auto& element = *transaction_rbegin;
```
efernandez, line 290: Just for clarifying: it shouldn't be `const` because the `element.transaction` is (conditionally) modified below in `applyMotionModels(element.sensor_name, *element.transaction)`, right?

efernandez (approved):

fuse_optimizers/src/fixed_lag_smoother.cpp
svwilliams: Right. The `element` variable holds a transaction pointer, so we can modify the contents of the transaction without modifying the transaction pointer, so C++ is fine with the `const`. I just feel it goes against the spirit of const-ness.

efernandez: Actually, I believe `const auto&` would be `const Transaction*`, so it's perfectly fine to modify the `element` transaction attributes. That wouldn't be the case if it were a `Transaction const*` or `const Transaction const*`, i.e. a pointer (either `const` or not) to `const` data.

So I believe this was already correct before, because the `const Transaction* element` pointer doesn't change, only its data does.

Am I missing anything?

svwilliams: Not missing anything, no. It was technically correct, but, in my opinion, not philosophically correct. The `pending_transactions_` container owns the transaction at this point. The fact that it holds a pointer instead of an object is, at best, an performance optimization detail. Since the transaction is being modified through `element`, denoting `element` as non-const helps indicate the contents of the container are going to change. Whereas having `element` be `const` hides the fact that you are modifying the contents of the container.

Perhaps what it means is the container definition should change to hold objects instead of pointers. There are a handful of places I feel I need to change shared_ptrs to unique_ptrs or objects to get the ownership semantics correct, but haven't found the time to do that refactoring yet.

efernandez: I see. Since it wasn't broken and this is equivalent and, as you say, without the `const` qualifier it doesn't lead the code reader to think the `element` transaction attributes don't change, this looks good. :+1:

ayrton04 (approved):

needphpsql (commented):
looks fine to me

PR #161, opened 2020-03-19
Keep constraint properties sorted by source

efernandez:
This keeps the constraint properties sorted by source regardless of the order in which they're discovered and added to the parent property.

This implementation relies on the alphabetical order of the constraint source `std::string`, which is used as the key in an `std::map`.

The sorted properties would look as follows:
Screenshot from 2020-03-19 20-03-14

### DIFF
--- a/fuse_viz/include/fuse_viz/serialized_graph_display.h
+++ b/fuse_viz/include/fuse_viz/serialized_graph_display.h
@@ -43,6 +43,7 @@
 #include <rviz/message_filter_display.h>
 #endif  // Q_MOC_RUN
 
+#include <map>
 #include <memory>
 #include <string>
 #include <unordered_map>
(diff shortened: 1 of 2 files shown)


efernandez:

fuse_viz/include/fuse_viz/serialized_graph_display.h
```diff
@@ -46,6 +46,7 @@
 #include <memory>
 #include <string>
 #include <unordered_map>
+#include <map>
```
svwilliams: I don't suppose you can move this up above `<memory>`? I try (unsuccessfully) to keep these alphabetized.

svwilliams (approved):

fuse_viz/include/fuse_viz/serialized_graph_display.h
efernandez: Done.

BTW I actually use a `vim` plugin that automatically `clang-format`s every C++ file on exit, using the first `.clang_format` file in the path up from the cwd. However, I always have to exit `vim` with `:noautocmd x` so it doesn't run because many things in `fuse` don't follow the style rules in that `.clang_format` file I have.

It'd be great if we could have such a `.clang_format` file in `fuse`. I'm happy to contribute that if you want, and add some `// clang-format off` - `// clang-format on` blocks where needed, since there are some pieces of code that are better formatted as they already are. :smiley: 

I wonder what are your thoughts about this?

ayrton04 (approved):

PR #162, opened 2020-03-27
Throttle

efernandez:
# Contribution

This adds a `ThrottledCallback` class and uses it in all sensor models, so they can throttle the input messages at the user given `throttle_period`, which is a new ROS param added to all of them. This param defaults to `0.0`, so no throttling happens by default, which is the previous behaviour.

# Motivation

This makes things cleaner from outside, because even though we could use `rosrun topic_tools throttle messages`, that requires changes in the `launch` files and it doesn't scale well if we have to maintain multiple robots with different sensor suites and topic names. With this change we only need to set the `throttled_period` param in a `yaml` file, and we can even pass its value from a `launch` file with  `subst_value="true"` if we want to.

### DIFF
--- a/fuse_core/include/fuse_core/parameter.h
+++ b/fuse_core/include/fuse_core/parameter.h
@@ -66,7 +66,7 @@ void getParamRequired(const ros::NodeHandle& nh, const std::string& key, T& valu
 }
 
 /**
- * @brief Helper function that loads strictly positive integral or floating point values from the parameter server
+ * @brief Helper function that loads positive integral or floating point values from the parameter server
  *
  * @param[in] node_handle - The node handle used to load the parameter
  * @param[in] parameter_name - The parameter name to load
--- a/fuse_models/include/fuse_models/common/throttled_callback.h
+++ b/fuse_models/include/fuse_models/common/throttled_callback.h
@@ -67,7 +67,8 @@ public:
    *                            nullptr
    * @param[in] throttle_period The throttling period duration in seconds. Defaults to 0.0, i.e. no throttling
    */
-  ThrottledCallback(Callback keep_callback = nullptr, Callback drop_callback = nullptr,
+  ThrottledCallback(Callback&& keep_callback = nullptr,  // NOLINT(whitespace/operators)
+                    Callback&& drop_callback = nullptr,  // NOLINT(whitespace/operators)
                     const ros::Duration& throttle_period = ros::Duration(0.0))
     : keep_callback_(keep_callback), drop_callback_(drop_callback), throttle_period_(throttle_period)
   {
--- a/fuse_models/include/fuse_models/common/throttled_callback.h
+++ b/fuse_models/include/fuse_models/common/throttled_callback.h
@@ -135,7 +135,11 @@ public:
     const auto now = ros::Time::now();
     if (!last_called_time_.isValid() || now - last_called_time_ > throttle_period_)
     {
-      keep_callback_(message);
+      if (keep_callback_)
+      {
+        keep_callback_(message);
+      }
+
       last_called_time_ = now;
     }
     else if (drop_callback_)
--- a/fuse_models/include/fuse_models/common/throttled_callback.h
+++ b/fuse_models/include/fuse_models/common/throttled_callback.h
@@ -66,11 +66,15 @@ public:
    * @param[in] drop_callback   The callback to call when the message is dropped because of the throttling. Defaults to
    *                            nullptr
    * @param[in] throttle_period The throttling period duration in seconds. Defaults to 0.0, i.e. no throttling
+   * @param[in] use_wall_time   Whether to use ros::WallTime or not. Defaults to false
    */
   ThrottledCallback(Callback&& keep_callback = nullptr,  // NOLINT(whitespace/operators)
                     Callback&& drop_callback = nullptr,  // NOLINT(whitespace/operators)
-                    const ros::Duration& throttle_period = ros::Duration(0.0))
-    : keep_callback_(keep_callback), drop_callback_(drop_callback), throttle_period_(throttle_period)
+                    const ros::Duration& throttle_period = ros::Duration(0.0), const bool use_wall_time = false)
+    : keep_callback_(keep_callback)
+    , drop_callback_(drop_callback)
+    , throttle_period_(throttle_period)
+    , use_wall_time_(use_wall_time)
   {
   }
 
@@ -85,6 +89,16 @@ public:
   }
 
   /**
+   * @brief Use wall time flag getter
+   *
+   * @return True if using ros::WallTime, false otherwise
+   */
+  bool getUseWallTime() const
+  {
+    return use_wall_time_;
+  }
+
+  /**
    * @brief Throttle period setter
    *
    * @param[in] throttle_period The new throttle period duration in seconds to use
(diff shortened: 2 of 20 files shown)


efernandez:

fuse_models/include/fuse_models/common/throttled_callback.h
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * @brief Message callback that throttles the calls to the keep callback provided. When the message is dropped because
+   * of throttling, the drop callback is called instead.
+   *
+   * @param[in] message The input message
+   */
+  void callback(const MessageConstPtr& message)
+  {
+    const auto now = ros::Time::now();
+    if (!last_called_time_.isValid() || now - last_called_time_ > throttle_period_)
```
efernandez: In case it's not clear, the `!last_called_timer_.isValid()` check is done in order to always published the first message received, even if `now` is very close to `0`, which could happen in simulation or in bag recorded from simulation.

I also wonder if I should provide a ROS param `use_wall_time` to use the `ros::WallTime::now()` instead of the `ros::Time::now()`, similar to `topic_tools throttle`:
https://github.com/ros/ros_comm/blob/d18e0dd7a7ffe5b76d5d31a9e1ba7f150e486f73/tools/topic_tools/src/throttle.cpp#L124-L138

svwilliams: You can add a walltime option if you want; it definitely doesn't hurt anything. I'm just not sure what the use-case is. If I'm running a sim at 0.5x or 5x realtime, I still want the sensor data to behave the same.

fuse_models/include/fuse_models/common/throttled_callback.h
```diff
@@ -0,0 +1,159 @@
...
+    {
+      keep_callback_(message);
```
svwilliams: The `keep_callback_`  could be null.

```suggestion
      if (keep_callback_) keep_callback_(message);
```

fuse_models/src/acceleration_2d.cpp
```diff
@@ -51,7 +51,8 @@ namespace fuse_models
 Acceleration2D::Acceleration2D() :
   fuse_core::AsyncSensorModel(1),
   device_id_(fuse_core::uuid::NIL),
-  tf_listener_(tf_buffer_)
+  tf_listener_(tf_buffer_),
+  throttled_callback_(std::move(std::bind(&Acceleration2D::process, this, std::placeholders::_1)))
```
svwilliams: Just curious, is the `std::move` needed here? I thought since `std::bind` is returning a temporary anyway, it was intrinsically moveable.

fuse_models/src/acceleration_2d.cpp
```diff
@@ -73,8 +76,8 @@ void Acceleration2D::onStart()
 {
   if (!params_.indices.empty())
   {
-    subscriber_ =
-      node_handle_.subscribe(ros::names::resolve(params_.topic), params_.queue_size, &Acceleration2D::process, this);
+    subscriber_ = node_handle_.subscribe(ros::names::resolve(params_.topic), params_.queue_size,
```
svwilliams, line 80: Interesting. I wonder if a `ThrottledSubscriber` would result in more idiomatic usage?
Something like:
```
using Callback = std::function<void(const MessageConstPtr&)>;
ThrottledSubscriber(
  const ros::NodeHandle& node_handle,
  const std::string& topic_name,
  uint32_t queue_size,
  const ros::Duration& throttle_period,
  Callback keep_callback,
  Callback drop_callback = nullptr,
  const ros::TransportHints& transport_hints = ros::TransportHints());
```
And then when you subscribe, it looks "more normal":
```
subscriber_ = ThrottledSubscriber(node_handle_, ros::names::resolve(params_.topic), params_.queue_size, params_.throttle_period, boost::bind(&Acceleration2D::process, this, _1));
```

Just a thought. The throttled callback works just fine as well.

svwilliams (approved):

fuse_models/include/fuse_models/common/throttled_callback.h
efernandez: Done

fuse_models/include/fuse_models/parameters/acceleration_2d_params.h
```diff
@@ -68,6 +68,9 @@ struct Acceleration2DParams : public ParameterBase
 
       nh.getParam("disable_checks", disable_checks);
       nh.getParam("queue_size", queue_size);
+
+      throttle_period.fromSec(fuse_core::getPositiveParam(nh, "throttle_period", throttle_period.toSec(), false));
```
efernandez: @svwilliams I just fixed this because I was using `getPositiveParam` incorrectly. I thought it was "returning" by argument. :man_facepalming: 

Any comments regarding this way to call it (maybe it's cleaner in two lines with a local variable)? and whether the `getPositiveParam` API is confusing being the only one that returns the param instead of changing the input/output argument (as `getParam` and `getParamRequired`)?

svwilliams: No particular reason, really. I'm sure I just wasn't thinking about the standard ROS function signatures at the time. I'll gladly merge a PR that changes the signature, if you get to it before I do.

efernandez: I've changed `getPositiveParam` API in https://github.com/locusrobotics/fuse/pull/169

I'll update this PR once that's in.

fuse_models/src/acceleration_2d.cpp
efernandez: The main reason I took the throttle callback path was that the ROS subscriber API supports multiple different input types, so I was afraid that would be a limitation. Besides that, a throttle callback is a more generic component that could also be used without ROS subscribers.

fuse_models/include/fuse_models/common/throttled_callback.h
efernandez: I just added a `use_wall_time_` flag and `throttle_use_wall_time` (not sure about the name :thinking:) ROS param. Both default to `false`.

I think when we play things back from a bag file, it could be useful to use the `ros::WallTime`, so everything gets processed if we either play at a low rate, or we pause playing. Indeed, when we pause playing something special happens: if the callback queue has pending callbacks, they keep getting processed, even if the `ros::Time` is not growing. Indeed, this is a corner case not handled well with the current implementation, because the `now - last_called_time_` difference would be `0.0`, and those callbacks get dropped, even if `throttle_period_ == 0.0`. :disappointed: 

I could either handle `throttle_period_ == 0.0` differently, or allow `throttle_period_ < 0.0`, and use `-1.0` as a default. But actually, `0.0` should be handled properly, so it looks like the right answer here would be to handle that value differently, as a particular special case, where all callbacks are kept regardless of `now` and `last_called_time_`.

efernandez: I've updated this PR with this. I've also rebased it on top of #169 , which now must be review and merged before continue reviewing this one.

fuse_models/src/acceleration_2d.cpp
efernandez: Removed `std::move`

fuse_core/include/fuse_core/parameter.h
```diff
@@ -66,7 +66,7 @@ void getParamRequired(const ros::NodeHandle& nh, const std::string& key, T& valu
 }
 
 /**
- * @brief Helper function that loads strictly positive integral or floating point values from the parameter server
```
efernandez, line 69: I forgot to update the doxygen in the previous PR.

I was trying to support `getPositiveParam` for `ros::Duration`, but it requires too much boilerplate, so I've given up. :smile:

svwilliams (approved):

ayrton04 (approved):

PR #163, opened 2020-03-28
Replace ignition_sensors list param with ignition field

efernandez:
This PR removes the `ignition_sensors` list param and replaces it with an `ignition` ROS param field for the `fuse_models::Unicycle2DIgnition` sensor model. It also adds a `virtual bool ignition() const` method fo the base `SensorModel` class, so the `fuse_optimizer` nodes can rely on that to implement the ignition logic.

Some minor cleanup changes are also included in this PR.

### DIFF
--- a/fuse_optimizers/src/optimizer.cpp
+++ b/fuse_optimizers/src/optimizer.cpp
@@ -243,6 +243,9 @@ void Optimizer::loadSensorModels()
 
   for (const auto& config : sensor_model_configs)
   {
+    // Check whether this is an ignition sensor model or not
+    const bool ignition = config.config.hasMember("ignition") ? static_cast<bool>(config.config["ignition"]) : false;
+
     // Create a sensor object using pluginlib. This will throw if the plugin name is not found.
     auto sensor_model = sensor_model_loader_.createUniqueInstance(config.type);
     // Initialize the sensor
@@ -250,7 +253,8 @@ void Optimizer::loadSensorModels()
       config.name,
       std::bind(&Optimizer::injectCallback, this, config.name, std::placeholders::_1));
     // Store the sensor in a member variable for use later
-    sensor_models_.emplace(config.name, std::move(sensor_model));
+    sensor_models_.emplace(config.name,
+                           SensorModelInfo{ std::move(sensor_model), ignition });  // NOLINT(whitespace/braces)
 
     // Parse out the list of associated motion models, if any
     if ( (config.config.hasMember("motion_models"))
@@ -378,7 +382,7 @@ void Optimizer::notify(
   {
     try
     {
-      name__sensor_model.second->graphCallback(graph);
+      name__sensor_model.second.model->graphCallback(graph);
     }
     catch (const std::exception& e)
     {
@@ -441,7 +445,7 @@ void Optimizer::startPlugins()
   }
   for (const auto& name_plugin : sensor_models_)
   {
-    name_plugin.second->start();
+    name_plugin.second.model->start();
   }
   for (const auto& name_plugin : publishers_)
   {
@@ -459,7 +463,7 @@ void Optimizer::stopPlugins()
   }
   for (const auto& name_plugin : sensor_models_)
   {
-    name_plugin.second->stop();
+    name_plugin.second.model->stop();
   }
   for (const auto& name_plugin : motion_models_)
   {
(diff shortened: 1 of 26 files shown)


efernandez:

svwilliams:
I'd like to propose an alternative approach and see how you feel about it.

First, I really like how your proposed solution cleans up a bunch of the code, converting annoying searches of collections of strings into a direct property lookup.

However, I feel like the decision to use a specific sensor model as an ignition source is really a system-level design question, and not something that is an intrinsic property of the sensor itself. But I do acknowledge that is not a completely true statement, as the existence of the `Unicycle2DIgnition` sensor model proves. If all sensors were designed with an `ignition` property controlled by the parameter server, I would feel better about it. Or at least all sensors derived from some reasonable base class.

But as an alternative, what if we make the `ignition` property a boolean parameter that is part of the optimizer config, in the same way that each sensor specifies the set of motion models to use:
```
sensor_models:
  sensor1:
    type: my_package::Sensor1Model
    motion_models: ['unicycle']
    ignition: true
  sensor2:
    type: my_package::Sensor2Model
    motion_models: ['unicycle']
    ignition: false
```
And we could make the `ignition` parameter optional and default to false.

Then, over in the Optimizer class, we could make a struct to hold the sensor model and the ignition parameter together:
https://github.com/locusrobotics/fuse/blob/devel/fuse_optimizers/include/fuse_optimizers/optimizer.h#L123
```
struct SensorModelInfo
{
  SensorModelUniquePtr model;
  bool ignition;
}
using SensorModels = std::unordered_map<std::string, SensorModelInfo>;
```
(only all the names should be far more clever).

That should allow the same type of direct lookup inside of the optimizer code:
```
if (sensor_models_.at(sensor_name).ignition)
```
without all of that annoying searching through a sorted list of names.

And if we were really ambitious, we could include the motion models in that structure as well, eliminating the need for the `associated_motion_models_` variable.
https://github.com/locusrobotics/fuse/blob/devel/fuse_optimizers/include/fuse_optimizers/optimizer.h#L129

Would that solution fit your needs? Or is there a reason you think ignition-ness should be defined in the sensor properties and not in the optimizer/system properties?

efernandez:
> Would that solution fit your needs? Or is there a reason you think ignition-ness should be defined in the sensor properties and not in the optimizer/system properties?

Yes, I think that would also work for me. I can try to implement that and give it a try over the weekend.

Indeed, your proposal has the benefit that existing ignition sensor models don't have to change, i.e. they don't have to implement the `ignition()` method.

efernandez:
@svwilliams Updated with your proposed approach.

fuse_optimizers/src/optimizer.cpp
```diff
@@ -243,14 +243,18 @@ void Optimizer::loadSensorModels()
...
     // Create a sensor object using pluginlib. This will throw if the plugin name is not found.
     auto sensor_model = sensor_model_loader_.createUniqueInstance(config.type);
     // Initialize the sensor
     sensor_model->initialize(
       config.name,
       std::bind(&Optimizer::injectCallback, this, config.name, std::placeholders::_1));
     // Store the sensor in a member variable for use later
-    sensor_models_.emplace(config.name, std::move(sensor_model));
+    sensor_models_.emplace(config.name,
+                           SensorModelInfo{ std::move(sensor_model), ignition });  // NOLINT(whitespace/braces)
```
efernandez, line 257: I couldn't use `{ std::move(sensor_model), ignition }`, or `{{ std::move(sensor_model), ignition }}`.

So I had to add a constructor to `SensorModelInfo`.

I also had to take the `sensor_model` by copy, but I'm using `std::move` here and in the attribute initialization. Not sure if all the `std::move`s are required, or whether I could take the argument with move semantics using `&&`. :thinking:

svwilliams: Best I can tell, passing unique_ptrs by value is the "correct" method. And then enough `std::move` calls are sprinkled around to make it work. I always come back to the [this article](https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/) to figure out how I'm supposed to do things with smart pointers.

svwilliams (approved):

fuse_optimizers/src/optimizer.cpp
efernandez: Good. Then, according to that article, in this case this class constructor implements a "sink", so the implementation looks correct. Thanks.

svwilliams:
Can you get this rebased on devel. I'm guessing something in a recent merge created a conflict.

efernandez:
Rebased and fixed conflicts.

ayrton04 (approved):

svwilliams (approved):

PR #164, opened 2020-03-29
Add missed fuse_models dependencies

efernandez:
This fixes the following compilation error:

`workspace/src/fuse_optimizers/test/test_fixed_lag_ignition.cpp:34:33: fatal error: fuse_models/SetPose.h: No such file or directory`

The `fuse_models/SetPose.h` is used in https://github.com/locusrobotics/fuse/blob/devel/fuse_optimizers/test/test_fixed_lag_ignition.cpp#L34

### DIFF
--- a/fuse_optimizers/CMakeLists.txt
+++ b/fuse_optimizers/CMakeLists.txt
@@ -129,6 +129,7 @@ install(
 if(CATKIN_ENABLE_TESTING)
   find_package(roslint REQUIRED)
   find_package(rostest REQUIRED)
+  find_package(fuse_models REQUIRED)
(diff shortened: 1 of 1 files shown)


efernandez:
@svwilliams It looks like you forgot this in https://github.com/locusrobotics/fuse/pull/158. The dependency was already in the `package.xml` file, but only as an `exec_depend`. Now is also a `build_depend`, so the additional `CMakeLists.txt` changes in this PR are required. :smiley: 

@ayrton04 for review

svwilliams (approved):
Thanks

ayrton04 (approved):

PR #165, opened 2020-03-30
Add latch param to serialized publisher

efernandez:
Add `latch` ROS param to the `fuse_publishers::SerializedPublisher` (defaults to `true`) to allow making the `graph` and `transaction` output topics latch.

This is important if we want to record the `graph` topic and play back only the first message to initialize/ignite the optimizer.


efernandez:

svwilliams (approved):

ayrton04 (approved):

PR #166, opened 2020-04-01
Don't require frame if empty indices

efernandez:
This simply makes the:
* `twist_target_frame` not required if there's no linear or angular velocity indices set
* `pose_target_frame` not required if there's no position or orientation indices set

And old comment is also updated.

### DIFF
--- a/fuse_models/include/fuse_models/common/sensor_proc.h
+++ b/fuse_models/include/fuse_models/common/sensor_proc.h
@@ -859,7 +859,7 @@ inline bool processDifferentialPoseWithTwistCovariance(
     }
   }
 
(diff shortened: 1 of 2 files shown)


efernandez:

This is certainly minor, but if you have an `Odometry2D` sensor model with a `differential: true` config and you want to temporarily disable the pose indices, then you'll have to add the `pose_target_frame`, even if not position or orientation indices are used. Obviously, the linear or angular velocity indices are set for the twist absolute constraints.

svwilliams (approved):
I just ran into that very thing a day or two ago. Thanks.

ayrton04 (approved):

PR #167, opened 2020-04-03
Fix thread issue with UUID generation

svwilliams:
The original Boost random UUID generator is thread-safe, but it was very slow. Some time ago, I replaced the generator code with equivalent logic using the STL random library to improve performance. But had the downside of not being thread-safe. This manifests itself occasionally with different constraints having the same UUID.

This PR includes:
* A mutex lock on the random number generator to handle calls from multiple threads correctly.
* A slight refactor of how the random bits are copied into the UUID bytes in an attempt to minimize the lock time.

### DIFF
--- a/fuse_core/src/uuid.cpp
+++ b/fuse_core/src/uuid.cpp
@@ -40,9 +40,9 @@
 
 #include <algorithm>
 #include <array>
+#include <mutex>
 #include <random>
 #include <string>
-
 
 
 namespace fuse_core
@@ -88,21 +88,14 @@ UUID generate()
 {
   static std::random_device rd;
   static std::mt19937 generator(rd());
-  static std::uniform_int_distribution<uint32_t> distibution;
+  static std::uniform_int_distribution<uint64_t> distibution;
+  static std::mutex distribution_mutex;
 
   UUID u;
-
-  int i = 0;
-  auto random_value = distibution(generator);
-  for (UUID::iterator it = u.begin(); it != u.end(); ++it, ++i)
   {
-    if (i == sizeof(random_value))
-    {
-      random_value = generator();
-      i = 0;
-    }
-    // static_cast gets rid of warnings of converting unsigned long to boost::uint8_t
-    *it = static_cast<UUID::value_type>((random_value >> (i*8)) & 0xFF);
+    std::lock_guard<std::mutex> lock(distribution_mutex);
+    *reinterpret_cast<uint64_t*>(u.data) = generator();
+    *reinterpret_cast<uint64_t*>(u.data + 4) = generator();
   }
 
   // set variant
--- a/fuse_core/src/uuid.cpp
+++ b/fuse_core/src/uuid.cpp
@@ -91,12 +91,17 @@ UUID generate()
   static std::uniform_int_distribution<uint64_t> distibution;
   static std::mutex distribution_mutex;
 
-  UUID u;
+  uint64_t random1;
+  uint64_t random2;
   {
     std::lock_guard<std::mutex> lock(distribution_mutex);
-    *reinterpret_cast<uint64_t*>(u.data) = generator();
-    *reinterpret_cast<uint64_t*>(u.data + 4) = generator();
+    random1 = generator();
+    random2 = generator();
   }
+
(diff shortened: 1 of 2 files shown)


fuse_core/src/uuid.cpp
```diff
@@ -88,21 +88,14 @@ UUID generate()
...
   {
-    if (i == sizeof(random_value))
-    {
-      random_value = generator();
-      i = 0;
-    }
-    // static_cast gets rid of warnings of converting unsigned long to boost::uint8_t
-    *it = static_cast<UUID::value_type>((random_value >> (i*8)) & 0xFF);
+    std::lock_guard<std::mutex> lock(distribution_mutex);
+    *reinterpret_cast<uint64_t*>(u.data) = generator();
```
svwilliams: Please pay extra attention to this refactor.

My interpretation of the original code is that a random number of N bytes is generated. Inside the loop, each random byte is copied into the next open slot in the UUID. When you run out of random bytes, generate N more bytes.

The refactor relies on the uint8_t type in the UUID and the uint64_t type generated by the random number generator being the expected sizes.

That seems like a strong guarantee in C++:
> unsigned integer type with width of exactly 8, 16, 32 and 64 bits respectively
(provided only if the implementation directly supports the type)

https://en.cppreference.com/w/cpp/types/integer

fuse_core/src/uuid.cpp
```diff
@@ -88,23 +88,21 @@ UUID generate()
...
-  static std::uniform_int_distribution<uint32_t> distibution;
-
-  UUID u;
+  static std::uniform_int_distribution<uint64_t> distibution;
+  static std::mutex distribution_mutex;
 
-  int i = 0;
-  auto random_value = distibution(generator);
-  for (UUID::iterator it = u.begin(); it != u.end(); ++it, ++i)
+  uint64_t random1;
```
svwilliams, line 94: Please pay extra attention to this refactor.

My interpretation of the original code is that a random number of N bytes is generated. Inside the loop, each random byte is copied into the next open slot in the UUID. When you run out of random bytes, generate N more bytes.

The refactor relies on the uint8_t type in the UUID and the uint64_t type generated by the random number generator being the expected sizes.

That seems like a strong guarantee in C++:
>unsigned integer type with width of exactly 8, 16, 32 and 64 bits respectively
(provided only if the implementation directly supports the type)

https://en.cppreference.com/w/cpp/types/integer

ayrton04 (approved):

fuse_core/test/test_uuid.cpp
```diff
@@ -146,6 +149,59 @@ TEST(UUID, Generate)
...
+void generateUUIDs(UUIDs& uuids)
+{
+  constexpr size_t uuid_count = 100000;
+  uuids.reserve(uuid_count);
+  for (size_t i = 0; i < uuid_count; ++i)
+  {
+    auto uuid = fuse_core::uuid::generate();
+    uuids.push_back(uuid);
+  }
+}
```
needphpsql, line 161: I think the general UUID generator does more work to ensure it is globally unique, which explains why it is slow. For test of 100K times, it may still get low probability to repeat, what if even larger numbers? So, this depends on application, what if there are duplicated UUID happens in the program? Any critical consequence?

svwilliams: I was running tests locally with 1 billion UUIDs, but I didn't want the unit test run by the server to take 10+ minutes.

The original code was taken from the Boost random UUID implementation. I feel like they are literally just populating the UUID with random bits.

But yes, duplicates are bad. The UUID is used to determine if two things are the same. Incorrectly saying thing1 is the same as thing2 is a critical error.

svwilliams: For whatever this random website I dug up is worth, it also says generate 16 bytes of random data.
https://www.cryptosys.net/pki/uuid-rfc4122.html

needphpsql: Since it is extensively tested both locally and by community, it should be fine.

needphpsql (approved):
looks good to me.

efernandez (approved):

PR #168, opened 2020-04-03
Fix motion model history

svwilliams:
The motion model history was being pruned too much. This PR modifies the pruning logic to ensure that entries are available for *at least* the requested history length. This will generally mean that one entry will remain in the buffer that is before the expiration time.

Because the motion model is expected to generate segments for anything inside the history length, it was previously possible for the motion model to be missing the required data. This is particularly likely in situations with large time gaps between requests.

As an example, consider a history buffer with the following time entries:
```
{0s, 1s, 9s, 11s, 12s}
```
If the history length is set to 10s, then the original code would purge the t=0s and t=1s entries, as both are older than the t=2s limit (t=12s - 10s length). However, that would leave just the t=9s, t=11s, and t=12s entries. This is only a 3s history, much shorter history than the requested 10s. In contrast, with this PR in place, only the t=0s entry would be purged; the t=1s entry would remain. Thus, the history will span 11s, which is at least as large as the requested span of 10s.

### DIFF
--- a/fuse_core/include/fuse_core/message_buffer_impl.h
+++ b/fuse_core/include/fuse_core/message_buffer_impl.h
@@ -96,7 +96,7 @@ typename MessageBuffer<Message>::message_range MessageBuffer<Message>::query(
   }
   // Find the entry that is strictly greater than the requested beginning stamp. If the extended range flag is true,
   // we will then back up one entry.
-  auto upper_bound_comparison = [](const ros::Time& stamp, const typename Buffer::value_type& element) -> bool
+  auto upper_bound_comparison = [](const auto& stamp, const auto& element) -> bool
   {
     return (element.first > stamp);
   };
@@ -107,7 +107,7 @@ typename MessageBuffer<Message>::message_range MessageBuffer<Message>::query(
   }
   // Find the entry that is greater than or equal to the ending stamp. If the extended range flag is false, we will
   // back up one entry.
-  auto lower_bound_comparison = [](const typename Buffer::value_type& element, const ros::Time& stamp) -> bool
+  auto lower_bound_comparison = [](const auto& element, const auto& stamp) -> bool
   {
     return (element.first < stamp);
   };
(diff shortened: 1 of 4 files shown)


fuse_core/include/fuse_core/message_buffer_impl.h
```diff
@@ -137,14 +137,32 @@ void MessageBuffer<Message>::purgeHistory()
...
   }
-  // Continue to remove the first entry from the buffer until we:
-  // (a) are left with only two entries, OR
-  // (b) the time delta between the beginning and end is within the buffer_length_
+
+  // Compute the expiration time carefully, as ROS can't handle negative times
   const ros::Time& ending_stamp = buffer_.back().first;
-  while ((buffer_.size() > 2)
-      && ((ending_stamp - buffer_.front().first) > buffer_length_))
+  ros::Time expiration_time;
```
efernandez: This is correct, but you could also do:

```suggestion
  const auto expiration_time = ending_stamp.toSec() > buffer_length_.toSec() ? ending_stamp - buffer_length_ : ros::Time(0);
```

So you can enforce correctness.

fuse_core/include/fuse_core/message_buffer_impl.h
```diff
@@ -137,14 +137,32 @@ void MessageBuffer<Message>::purgeHistory()
...
+  }
+  else
+  {
+    expiration_time = ending_stamp - buffer_length_;
+  }
+  // Remove buffer elements before the expiration time.
+  // Be careful to ensure that:
+  //  - at least two entries remains at all times
+  //  - the buffer covers *at least* until the expiration time. Longer is acceptable.
+  auto is_greater = [](const ros::Time& stamp, const typename Buffer::value_type& element) -> bool
```
efernandez: I believe with C++14, you can use `auto` for the arguments.

```suggestion
  auto is_greater = [](const auto& stamp, const auto& element) -> bool
```

efernandez (approved):
LGTM. Just mentioned two suggestions that could certainly be ignored.

ayrton04 (approved):

PR #169, opened 2020-04-06
Get positive param API change

efernandez:
This changes the `fuse_core::getPositiveParam` API so it matches the `getParam` and `getRequiredParam` ones. That is, the loaded value is not `return`ed, but set in the input/output argument that holds the default value.

This also moves all the helper functions related with ROS parameters into the `parameter.h` header, and updates the ros and unit test accordingly. Some of these functions where in the more generic header `util.h`.

In order to keep the tests for the `util.h` header, this adds a test for `fuse_core::wrapAngle2D`. There's another commit that updates the expected range from `(-Pi, +Pi]` to `[-Pi, +Pi)`, since that seems to be the actual behaviour. This is likely indifferent. Otherwise, the implementation should be changed.

### DIFF
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -402,9 +402,33 @@ if(CATKIN_ENABLE_TESTING)
       CXX_STANDARD_REQUIRED YES
   )
 
+  # Parameter tests
+  add_rostest_gtest(test_parameter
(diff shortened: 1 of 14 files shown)


efernandez:

svwilliams (approved):

ayrton04 (approved):

PR #170, opened 2020-04-06
Add ComposedLoss

efernandez:
This adds composition of two loss functions. The error is the result of first evaluating `g` followed by `f` to yeild the composition `f(g(s))` for the squared residual `s`: `rho(s) = f_rho(g_rho(s))`

If either the `f` or `g` loss functions are not given, the `TrivialLoss` is used by default.

### DIFF
--- a/fuse_loss/src/composed_loss.cpp
+++ b/fuse_loss/src/composed_loss.cpp
@@ -1,0 +1,88 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2020, Clearpath Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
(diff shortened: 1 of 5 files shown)


fuse_loss/src/composed_loss.cpp
```diff
@@ -0,0 +1,88 @@
...
+                           const std::shared_ptr<fuse_core::Loss>& g_loss)
+  : f_loss_(f_loss), g_loss_(g_loss)
+{
+}
+
+void ComposedLoss::initialize(const std::string& name)
+{
+  ros::NodeHandle private_node_handle(name);
+
+  f_loss_ = fuse_core::loadLossConfig(private_node_handle, "f_loss");
```
efernandez, line 60: I'm open to suggestions regarding the name of each loss function. I called them `f` and `g` because I couldn't come up with another pair of names that are clear enough, so they're not confusing.

efernandez:

svwilliams (approved):
Yeah...I don't have any better naming suggestions.

ayrton04 (approved):

PR #172, opened 2020-04-09
Fix thread synchronization issue in the fixed-lag smoother

svwilliams:
The started/ignited variables are accessed from multiple threads. Changing them to atomic<bool> instead.

### DIFF
--- a/fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother.h
+++ b/fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother.h
@@ -165,9 +165,9 @@ protected:
                                            //!< and processed by the optimization thread.
   std::mutex pending_transactions_mutex_;  //!< Synchronize modification of the pending_transactions_ container
   ros::Time start_time_;  //!< The timestamp of the first ignition sensor transaction
-  bool started_;  //!< Flag indicating the optimizer is ready/has received a transaction from an ignition sensor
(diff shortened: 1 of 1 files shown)


efernandez (approved):
I thought the `pending_transactions_mutex_` was protecting the `started_` and `ignited_` flags, but I see some methods where they are changed without locking that mutex:

- write access:
  - https://github.com/locusrobotics/fuse/blob/3b59771f2a71e3e67c6cc5f063b7d7c6ef716ad0/fuse_optimizers/src/fixed_lag_smoother.cpp#L381-L382
  - https://github.com/locusrobotics/fuse/blob/3b59771f2a71e3e67c6cc5f063b7d7c6ef716ad0/fuse_optimizers/src/fixed_lag_smoother.cpp#L130
- read access:
  -  https://github.com/locusrobotics/fuse/blob/3b59771f2a71e3e67c6cc5f063b7d7c6ef716ad0/fuse_optimizers/src/fixed_lag_smoother.cpp#L414
  -  https://github.com/locusrobotics/fuse/blob/3b59771f2a71e3e67c6cc5f063b7d7c6ef716ad0/fuse_optimizers/src/fixed_lag_smoother.cpp#L230
  -  https://github.com/locusrobotics/fuse/blob/3b59771f2a71e3e67c6cc5f063b7d7c6ef716ad0/fuse_optimizers/src/fixed_lag_smoother.cpp#L478 

So this LGTM.

ayrton04 (approved):

PR #175, opened 2020-04-13
Delay throttle no valid state message

efernandez:
This makes the `No valid state data yet. Delaying tf broadcast.` **WARN** message from the `Odometry2DPublisher` delayed, i.e. only printed after the throttle period, not immediately.

It also adds a `ros::console::FilterBase` child class that implements the equivalent to `ROS_*_DELAYED_THROTTLE` (see http://wiki.ros.org/roscpp/Overview/Logging and http://docs.ros.org/api/rosconsole/html/console_8h_source.html) and allows to reset it, so when `onStart` is called, it's delayed again.

The motivation is to avoid the first **WARN** message that always gets printed on startup or after a setting a pose.

You can `rosservice call` the `reset` service to force it print after the throttle period expires.

### DIFF
--- a/fuse_core/include/fuse_core/console.h
+++ b/fuse_core/include/fuse_core/console.h
@@ -1,0 +1,108 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2020, Clearpath Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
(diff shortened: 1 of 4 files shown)


efernandez:

fuse_core/include/fuse_core/console.h
```diff
@@ -0,0 +1,106 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_CORE_CONSOLE_H
+#define FUSE_CORE_CONSOLE_H
+
+#include <ros/console.h>
+#include <ros/time.h>
+
+#ifndef ROSCONSOLE_THROTTLE_CHECK
+#define ROSCONSOLE_THROTTLE_CHECK(now, last, period) (ROS_UNLIKELY(last + period <= now) || ROS_UNLIKELY(now < last))
```
svwilliams: Is this here because of ROS version issues?

fuse_models/include/fuse_models/odometry_2d_publisher.h
```diff
@@ -200,6 +201,11 @@ class Odometry2DPublisher : public fuse_core::AsyncPublisher
   std::unique_ptr<tf2_ros::TransformListener> tf_listener_;
 
   ros::Timer publish_timer_;
+
+  /**
+   * @brief A ros::console filter to print delayed throttle messages, that can be reset
+   */
+  std::unique_ptr<ros::console::FilterBase> delayed_throttle_filter_{ new fuse_core::DelayedThrottleFilter(10.0) };
```
svwilliams: I have to ask if this really needs to be base-class pointer. It seems to me this could be:
```
fuse_core::DelayedThrottleFilter delayed_throttle_filter_ { 10.0 };
```
And then the reset operation becomes simply
```
delayed_throttle_filter_.reset();
```
And the console call becomes
```
ROS_WARN_STREAM_FILTER(&delayed_throttle_filter_, "No valid state data yet. Delaying tf broadcast.");
```

I did a quick test and that compiles, which is admittedly different from "it works".

svwilliams: The `delayed_throttle_filter_` object is used within the `publish_timer_` callback. As such, it is safer to destroy the timer before destroying the throttle filter. Would you mind moving `delayed_throttle_filter_` above `publish_timer_` for that reason?

fuse_core/include/fuse_core/console.h
efernandez: Yes, it was introduced in https://github.com/ros/rosconsole/pull/12/files#diff-bc97742e4e32254946668ef27dceea4cR367 and released in 1.13.8.

I was using 1.13.7 in my system, so I had to do this.

I've added a comment regarding that, but I'm also going to request `rosconsole` version is bumped. Maybe it's better not to have this here?! :thinking:

efernandez: If I can bump `rosconsole` version, I'm leaning towards removing this, unless you think it's better to keep it.

fuse_models/include/fuse_models/odometry_2d_publisher.h
efernandez: Oh, IIRC it failed to compile for me in `ROS_WARN_STREAM_FILTER(&delayed_throttle_filter_`, but it actually works now for me.

Changed.

efernandez: Good catch. Moved.

fuse_core/include/fuse_core/console.h
efernandez: I've submitted another commit that removes this and also adds the following to `package.xml`:
`<depend version_gte="1.13.8">rosconsole</depend>`

fuse_core/package.xml
```diff
@@ -17,6 +17,7 @@
   <depend>fuse_msgs</depend>
   <depend>pluginlib</depend>
   <depend>roscpp</depend>
+  <depend version_gte="1.13.8">rosconsole</depend>
```
svwilliams, line 20: I didn't even know specifying versions was a thing. Thanks for that.

svwilliams (approved):

fuse_core/package.xml
efernandez: FYI https://www.ros.org/reps/rep-0140.html#build-depend-multiple

It only mentions `build_depend`, but I've seen it used in all _depend_ flavours.

BTW I tried to confirm it doesn't break anything, but `rospack depends1 fuse_core` didn't work in my system. It complained about not finding other dependencies, like `libceres-dev` or even `boost`. I'm pretty sure this is only in my system. I know other packages using this w/o issues for a while, going through the entire debian generation process.

I guess the automatic checks you have in this repo should guarantee it's fine, or maybe you need to run an additional test to confirm it's well supported in your infrastructure.

ayrton04 (approved):

PR #176, opened 2020-04-13
Sort transactions by minimum timestamp and purge wrt the maximum timestamp of all pending transactions

efernandez:
This is on top of https://github.com/locusrobotics/fuse/pull/173. I actually cherry-picked your commit, @svwilliams .

This stores the transaction's mininum involved stamp into the `pending_transactions_` queue, instead of the transaction stamp. And when purging, the `last_pending_time` or `current_time` is set to the maximum of all transactions's maximum involved stamp.

This gets rids of the following **WARN** and **ERROR** messages:
``` bash
[INFO]: Received a set_pose request (stamp: 173.570000000, x: 1.4, y: 0.525, yaw: -0.0350735)
[WARN ]: Unable to locate a state in this history with stamp <= 173.540000000. Variables will all be initialized to 0.

[ERROR]: The queued transaction with timestamp 2589.220000000 from sensor laser_odometry is not an ignition sensor transaction. This transaction will not be processed individually.
```

I could work on a unit test to exercise this. The idea would be to extend the existing fixed lag ignition test to create a transaction with a relative constraints that has a minimum involved timestamp older than the transaction with the ignition timestamp, but a transaction stamp (and maximum involved timestamp) newer than it. If you think this PR is a good change, I can work on that. :thinking: 

Consider this a quick change, not very well tested, and still open to discussion, that complements the comments in https://github.com/locusrobotics/fuse/pull/173. I'm still not sure if we really need the transaction `stamp()` method and `stamp_` attribute, or we could just rely on the involved stamps, and return `ros::Time(0, 0)` if there's none. I guess that's not an option if the reason there's no involved stamps is because the variables aren't `Stamped`. I guess this answers my question. :smile: 

Another thin still unclear to me is whether we should queue with the minimum involved stamp or not, or whether there are side effects if we do so. :thinking: 

There are two minor commits here as well.

### DIFF
--- a/fuse_core/include/fuse_core/transaction.h
+++ b/fuse_core/include/fuse_core/transaction.h
@@ -122,6 +122,22 @@ public:
   const_stamp_range involvedStamps() const { return involved_stamps_; }
 
   /**
+   * @brief Read-only access to the minimum (oldest), timestamp among the transaction's stamp and all involved
+   * timestamps, if any
+   *
+   * @return The minimum (oldest) timestamp.
+   */
+  const ros::Time& minStamp() const;
+
+  /**
+   * @brief Read-only access to the maximum (newest) timestamp among the transaction's stamp and all involved
+   * timestamps, if any
+   *
+   * @return The maximum (newest) timestamp.
+   */
+  const ros::Time& maxStamp() const;
+
+  /**
    * @brief Read-only access to the added constraints
    *
    * @return  An iterator range containing all added constraints
--- a/fuse_optimizers/src/fixed_lag_smoother.cpp
+++ b/fuse_optimizers/src/fixed_lag_smoother.cpp
@@ -334,7 +334,10 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction, const r
   }
 
   // Use the most recent transaction time as the current time
-  const auto current_time = pending_transactions_.front().stamp();
+  const auto current_time =
+      std::max_element(pending_transactions_.begin(), pending_transactions_.end(),
+                       [](const auto& lhs, const auto& rhs) { return lhs.maxStamp() < rhs.maxStamp(); })
+          ->maxStamp();
 
   // Attempt to process each pending transaction
   auto sensor_blacklist = std::vector<std::string>();
@@ -342,11 +345,7 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction, const r
   while (transaction_riter != pending_transactions_.rend())
   {
     auto& element = *transaction_riter;
-    auto min_stamp = element.stamp();
-    if (!element.transaction->involvedStamps().empty() && *element.transaction->involvedStamps().begin() < min_stamp)
-    {
-      min_stamp = *element.transaction->involvedStamps().begin();
-    }
+    const auto& min_stamp = element.minStamp();
     if (min_stamp < lag_expiration)
     {
       ROS_DEBUG_STREAM("The current lag expiration time is " << lag_expiration << ". The queued transaction with "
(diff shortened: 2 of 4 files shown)


efernandez:

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -353,14 +369,14 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction)
     {
       // The motion model processing failed.
       // Check the transaction timeout to determine if it should be removed or skipped.
-      if (element.transaction->stamp() + params_.transaction_timeout < current_time)
+      if (min_stamp + params_.transaction_timeout < current_time)
```
svwilliams, line 372: I feel like this should use the max stamp instead. With the assumption the transaction arrived at the smoother at maxStamp() or later.
```suggestion
      if (element.maxStamp() + params_.transaction_timeout < current_time)
```

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -332,13 +333,28 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction)
     }
   }
 
+  // Use the most recent transaction time as the current time
+  const auto current_time =
```
svwilliams, line 337: Having to do a linear search here is annoying. *If* the pending queue was sorted by max time, then this would just be the end of the pending queue. And the `(min_stamp < lag_expiration)` filtering logic will still purge out transactions involving old stamps.

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -426,12 +442,12 @@ void FixedLagSmoother::transactionCallback(
     // The pending set is arranged "smallest stamp last" to making popping off the back more efficient
     auto comparator = [](const ros::Time& value, const TransactionQueueElement& element)
     {
-      return value >= element.stamp();
+      return value >= element.minStamp();
```
svwilliams, line 445: This change I'm not sure about. I mentioned this a bit in another PR.
If we order the transactions by minStamp(), then the transactions will not be processed in "measurement order".
For example:
IMU measures the state at t1 and the constraint involves only t1.
Relative odometry measures the state at t2 and the constraint involves t0 and t2.
Sorting by min stamp adds the odometry constraint measured at t2 _before_ adding the IMU constraint measured at t1.

To be clear, I have no idea if this is a problem at all. It just _feels_ wrong somehow...to me.

I suppose the one argument I have is from the motion model stand point. By inserting the intermediate states first, the motion model has to do slightly less work.
Example 1:
* Current motion model stamp is t0.
* Process IMU measurement with involved stamps (t1) creates segment t0->t1
* Process the Odometry measurement with involved stamps (t0, t2) creates segment t1->t2

Example 2:
* Current motion model stamp is t0.
* Process the Odometry measurement with involved stamps (t0, t2) creates segment t0->t2
* Process IMU measurement with involved stamps (t1) deletes segment t0->t2, creates segment t0->t1 and t1->t2

The motion model is capable of handling either. And the above example may be contrived and doesn't happen in reality. Perhaps a more realistic scenario is:
* Current motion model stamp is t0.
* Process IMU measurement with involved stamps (t2) creates segment t0->t2
* Process the Odometry measurement with involved stamps (t1, t3) deletes segment t0->t2, creates segment t0->t1, t1->t2, t2->3

In which case, any ordering will involve a lot of deleting and adding.

The `processQueue()` method has also evolved quite a bit from its original design. With the additional of blacklisting transactions from a particular sensor and continuing processing, we no longer add transactions to the graph in a strict order. It may be the actual order of processing is fairly arbitrary at this point, as long as the filtering is in place.

Curious to hear any thoughts you have?

fuse_core/include/fuse_core/transaction.h
```diff
@@ -121,6 +121,22 @@ class Transaction
    */
   const_stamp_range involvedStamps() const { return involved_stamps_; }
 
+  /**
+   * @brief Read-only access to the minimum (oldest), timestamp among the transaction's stamp and all involved
+   * timestamps, if any
+   *
+   * @return The minimum (oldest) timestamp.
+   */
+  const ros::Time& minStamp() const;
```
svwilliams, line 130: I like these methods. We should definitely keep these.

fuse_optimizers/src/fixed_lag_smoother.cpp
efernandez: > The processQueue() method has also evolved quite a bit from its original design. With the additional of blacklisting transactions from a particular sensor and continuing processing, we no longer add transactions to the graph in a strict order. It may be the actual order of processing is fairly arbitrary at this point, as long as the filtering is in place.

My understanding is that a sensor is blacklisted only if `applyMotionModels` fails, and I think that rarely happens. So, unless I'm wrong, `processQueue()` would process all the transactions and try to apply the motion models iterating the `pending_transactions_` queue in reverse order, that is, from older to newer stamp.

> Curious to hear any thoughts you have?

I also feel there's something that doesn't smell well if we don't use the measurement time, as you've mentioned.

I think the processing order should still use the measurement time, i.e. the transaction `stamp()`, which is typically the `maxStamp()` of the transaction involved stamps.

But for purging, I feel like the`minStamp()` of the transaction involved stamps should also be taken into account.

efernandez:
Closing in favour of #173 

@svwilliams I don't think there's anything missed here that you didn't took into #173 . Let me know otherwise. :smiley:

PR #178, opened 2020-04-15
Add use_twist_covariance ROS param and logic to Imu2D differential orientation measurements

efernandez:
This simply adds the `use_twist_covariance` ROS param and logic to the `Imu2D` sensor model, which is equivalent to the one added to the `Odometry2D` sensor model in https://github.com/locusrobotics/fuse/pull/138

I also did two additional minor changes:
1. `std::move` the current `pose` into the `previous_pose_` in `Odometry2D`, so it's doing the same as `Imu2D`, which seems more efficient.
2. Move the `pose_relative_covariance` declaration in `processDifferentialPoseWithTwistCovariance()` closer to where it's used.


efernandez:

svwilliams (approved):

ayrton04 (approved):

PR #180, opened 2020-04-22
Validate unicycle 2d

efernandez:
# Changelog

This extends the validation checks from https://github.com/locusrobotics/fuse/pull/125 to the `Unicycle2D` motion model. In particular:

* Add `disable_check`s param (defaults to `false`)
* Validate `state1` and `state2` are finite
* Validate process noise covariance (after it's been scaled and multiplied by dt)

In a separate commit the unused `EPSILON` variable is removed.

# Concerns about dt being very small

During my test I haven't seen any validation checks failing, but I'm surprised that with small `dt` values the resulting covariance is still valid. :thinking: 

I haven't check the `covariance.invserse().llt().matrixU()` expression that computes the `sqrt_information_` matrix in the actual constraint is always succeeding though. But TBH I haven't seen anything suggesting it's wrong. I indeed printed the `unicycle_2d` motion model transactions, and the biggest value I've seen in the `sqrt_information_` matrix was around `1.0e+6`, usually associate to a small `dt`. The smallest `dt` I've seen was around `1.0e-5`.

I believe there's no check for `dt` being very small. I think we should have one, but I don't know what to do in that case. I believe we shouldn't create any constraint, but we still need to connect measurement variables with motion model constraints. :thinking: 

I also wonder if there should be a validation check to confirm the expression `covariance.invserse().llt().matrixU()` succeeded. Indeed, the `LLT` class allows to check that with `llt.info() == Eigen::Success` (see https://eigen.tuxfamily.org/dox/classEigen_1_1LLT.html#adb1295e7d2b1fd825a041345ae08be54). :thinking:

### DIFF
--- a/fuse_models/include/fuse_models/unicycle_2d.h
+++ b/fuse_models/include/fuse_models/unicycle_2d.h
@@ -105,6 +105,13 @@ protected:
     tf2_2d::Vector2 acceleration_linear;  //!< Body-frame linear acceleration
 
     void print(std::ostream& stream = std::cout) const;
+
+    /**
+     * @brief Validate the state components: pose, linear velocity, yaw velocity and linear acceleration.
+     *
+     * This validates the state components are finite. It throws an exception if any validation check fails.
+     */
+    void validate() const;
   };
   using StateHistory = std::map<ros::Time, StateHistoryElement>;
 
@@ -163,6 +170,19 @@ protected:
     StateHistory& state_history,
     const ros::Duration& buffer_length);
 
+  /**
+   * @brief Validate the motion model state #1, state #2 and process noise covariance
+   *
+   * This validates the motion model states and process noise covariance are valid. It throws an exception if any
+   * validation check fails.
+   *
+   * @param[in] state1                   The first/oldest state
+   * @param[in] state2                   The second/newest state
+   * @param[in] process_noise_covariance The process noise covariance, after it is scaled and multiplied by dt
+   */
+  static void validateMotionModel(const StateHistoryElement& state1, const StateHistoryElement& state2,
+                                  const fuse_core::Matrix8d& process_noise_covariance);
+
   ros::Duration buffer_length_;                    //!< The length of the state history
   fuse_core::UUID device_id_;                      //!< The UUID of the device to be published
   fuse_core::TimestampManager timestamp_manager_;  //!< Tracks timestamps and previously created motion model segments
@@ -171,7 +191,9 @@ protected:
                                                    //!< of the current state twist
   double velocity_norm_min_{ 1e-3 };               //!< The minimum velocity/twist norm allowed when scaling the
                                                    //!< process noise covariance
-  StateHistory state_history_;                     //!< History of optimized graph pose estimates
+  bool disable_checks_{ false };  //!< Whether to disable the validation checks for the current and predicted state,
+                                  //!< including the process noise covariance after it is scaled and multiplied by dt
+  StateHistory state_history_;    //!< History of optimized graph pose estimates
 };
 
 std::ostream& operator<<(std::ostream& stream, const Unicycle2D& unicycle_2d);
--- a/fuse_models/src/unicycle_2d.cpp
+++ b/fuse_models/src/unicycle_2d.cpp
@@ -63,8 +63,6 @@ PLUGINLIB_EXPORT_CLASS(fuse_models::Unicycle2D, fuse_core::MotionModel)
 namespace fuse_models
 {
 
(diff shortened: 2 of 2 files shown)


efernandez:

svwilliams (approved):

ayrton04 (approved):

svwilliams:
I see no harm in checking the `llt()` result. However, with the current interface it will be a bit tricky. That check would need to be done in the constraint class, which is not an overly convenient place to generate warnings and errors. I suppose it could throw in that case, and get caught by the sensor. But I'm wondering if it makes more sense to compute the inverse sqrt inside the sensor itself, and check the results in the `validate()` method?

Regardless, the changes in this PR are pretty straight-forward. I'll get this merged, and additional checks can be added in the future if desired.

efernandez:
> But I'm wondering if it makes more sense to compute the inverse sqrt inside the sensor itself, and check the results in the validate() method?

@svwilliams I had the same thought. Indeed, if you think about the case of the motion model process noise covariance, which is multiplied by `dt`, it's probably more numerically stable to compute the sqrt information matrix and pass it to the contraint, after it's been validated.

I'm ignoring the fact that we can also scale the process noise covariance by the norm of the state velocity for now, but that would make this point even stronger. So, ignoring this for now, if we had to pass the sqrt information for the `Unicycle2D` motion model kinematic constraint, instead of the covariance, things would be like this:

# If we pass the covariance

We're actually passing:

 `dt * process_noise_covariance_`

and that is used to compute:

`.inverse().llt().matrixU()`

# If we pass the sqrt information

In this case we can pre-compute this once (in the initialization, after loading the ROS params):

`process_noise_sqrt_information_ = process_noise_covariance_.inverse().llt().matrixU()`

and then pass this to the kinematic constraint, that would include the `dt` scaling:

`process_noise_sqrt_information_ / std::sqrt(dt)`

---

We can get a similar benefit if the process noise covariance is scaled by the state velocity norm `v`, since instead of multiplying by `v^2`, we'll be dividing by `v`.

This would required and API change though. Thoughts?

I'm happy to look into this in a future PR.

svwilliams:
The API change should be fine. I think it is only used in this one place anyway.

efernandez:
> The API change should be fine. I think it is only used in this one place anyway.

It's indeed used in almost all constraints, although only for the motion model there's a constant part that could be pre-computed.

``` bash
$ ack-grep 'inverse\(\)\.llt'
fuse_constraints/include/fuse_constraints/relative_constraint_impl.h
58:    sqrt_information_(covariance.inverse().llt().matrixU())
81:  fuse_core::MatrixXd partial_sqrt_information = partial_covariance.inverse().llt().matrixU();

fuse_constraints/include/fuse_constraints/absolute_constraint_impl.h
57:    sqrt_information_(covariance.inverse().llt().matrixU())
77:  fuse_core::MatrixXd partial_sqrt_information = partial_covariance.inverse().llt().matrixU();

fuse_constraints/src/absolute_pose_2d_stamped_constraint.cpp
68:  fuse_core::MatrixXd partial_sqrt_information = partial_covariance.inverse().llt().matrixU();

fuse_constraints/src/absolute_pose_3d_stamped_constraint.cpp
57:    sqrt_information_(covariance.inverse().llt().matrixU())

fuse_constraints/src/relative_pose_3d_stamped_constraint.cpp
60:    sqrt_information_(covariance.inverse().llt().matrixU())

fuse_constraints/src/absolute_orientation_3d_stamped_constraint.cpp
56:    sqrt_information_(covariance.inverse().llt().matrixU())

fuse_constraints/src/relative_pose_2d_stamped_constraint.cpp
71:  fuse_core::MatrixXd partial_sqrt_information = partial_covariance.inverse().llt().matrixU();

fuse_constraints/src/relative_orientation_3d_stamped_constraint.cpp
57:    sqrt_information_(covariance.inverse().llt().matrixU())

fuse_constraints/src/absolute_orientation_3d_stamped_euler_constraint.cpp
58:    sqrt_information_(covariance.inverse().llt().matrixU()),

fuse_models/src/unicycle_2d_state_kinematic_constraint.cpp
80:    sqrt_information_(covariance.inverse().llt().matrixU())

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
55:  const fuse_core::Matrix8d sqrt_information{ covariance.inverse().llt().matrixU() };

fuse_models/benchmark/benchmark_unicycle_2d_state_cost_function.cpp
123:const fuse_core::Matrix8d Unicycle2DStateCostFunction::sqrt_information(covariance.inverse().llt().matrixU());
```

svwilliams:
That's what happens when I make assertions without checking first. I'm still fine with the API change. Maybe make a utility function to compute it this time around. But I'll merge this PR in the mean time.

PR #181, opened 2020-04-22
Only call generator if motion model history empty

efernandez:
# Changelog

This simplifies the changes from https://github.com/locusrobotics/fuse/pull/154 by doing:
* Only call generator if motion model history empty.
* Assert `beginning stamp < ending stamp`. Or `beginning stamp == ending stamp` for the special case when the state history is empty.
* Handle the `dt == 0` special case in motion model, so no variables or constraints are generator, and no prediction is performed, or the same state is inserted twice into the history.
* Revert the changes to the `test_timestamp_manager.cpp` unit test
* Add a new test for the case when the transaction has a single involved stamp, since that was precisely the case not handled. Indeed, that test fails w/o #154 or this PR.
* [minor] Declare `state2` closer to where it is used.

The special case mentioned above is always the same.

From my tests locally, with a bunch of prints, I confirmed the state history already have the state for the given stamp, so IIRC that's all that was required for #154 . The `assert` added in this PR enforces the generator function isn't called for those cases it did before, so I think that constitutes a test for this change.

### DIFF
--- a/fuse_core/test/test_timestamp_manager.cpp
+++ b/fuse_core/test/test_timestamp_manager.cpp
@@ -554,7 +554,7 @@ TEST_F(TimestampManagerTestFixture, MultiSegment)
   ASSERT_EQ(0ul, generated_time_spans.size());
 }
 
-TEST_F(TimestampManagerTestFixture, MultiSegmentBeforBeginning)
+TEST_F(TimestampManagerTestFixture, MultiSegmentBeforeBeginning)
 {
   // Test:
   // Existing: |------111111112222222233333333-------> t
--- a/fuse_core/test/test_timestamp_manager.cpp
+++ b/fuse_core/test/test_timestamp_manager.cpp
@@ -112,12 +112,9 @@ TEST_F(TimestampManagerTestFixture, Empty)
   EXPECT_EQ(ros::Time(20, 0), *stamp_range_iter);
 
   // Verify the expected queries were performed
-  // And additional time span is generated for the last stamp as beginning and ending stamp
-  ASSERT_EQ(2ul, generated_time_spans.size());
+  ASSERT_EQ(1ul, generated_time_spans.size());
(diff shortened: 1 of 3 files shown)


efernandez:

fuse_core/test/test_timestamp_manager.cpp
```diff
@@ -542,7 +554,7 @@ TEST_F(TimestampManagerTestFixture, MultiSegment)
   ASSERT_EQ(0ul, generated_time_spans.size());
 }
 
-TEST_F(TimestampManagerTestFixture, MultiSegmentBeforeBeginning)
+TEST_F(TimestampManagerTestFixture, MultiSegmentBeforBeginning)
```
svwilliams: Did I lose an 'e' on purpose?
```suggestion
TEST_F(TimestampManagerTestFixture, MultiSegmentBeforeBeginning)
```

svwilliams (approved):
I like this better than always adding a dummy entry at the end. Thanks.

fuse_core/test/test_timestamp_manager.cpp
efernandez: Fixed :smile:

efernandez:
> I like this better than always adding a dummy entry at the end. Thanks.

And the unit test changes I did in #154 aren't needed any more. Indeed, I reverted all those changes and added a new test. :smiley:

ayrton04 (approved):

efernandez:
FYI Note that it's important to get this PR in now that https://github.com/locusrobotics/fuse/pull/180 has been merged, because you'll see ROS messages complaining of zero process noise covariance matrices. They'll be spamming at 0.1Hz, because they're throttled, but that could still be a lot of spam.

That's precisely the case of `dt == 0` that is handled as special case in this PR, that also makes `dt == 0` happen once per set pose or startup, as opposed to almost constantly.

PR #185, opened 2020-04-29
Remove variables on hold

efernandez:
I'm not really holding any variables atm, but I noticed that `removeVariable` didn't `erase` the `variable_uuid` from the `variables_on_hold_` set.

This fixes that and adds a unit test that required the `isVariableOnHold()` method to be created. The unit test failed before this change.

### DIFF
--- a/fuse_core/include/fuse_core/graph.h
+++ b/fuse_core/include/fuse_core/graph.h
@@ -317,6 +317,14 @@ public:
   virtual void holdVariable(const UUID& variable_uuid, bool hold_constant = true) = 0;
 
   /**
+   * @brief Check whether a variable is on hold or not
+   *
+   * @param[in] variable_uuid The variable to test
+   * @return True if the variable is on hold, false otherwise
+   */
+  virtual bool isVariableOnHold(const UUID& variable_uuid) const = 0;
+
+  /**
    * @brief Compute the marginal covariance blocks for the requested set of variable pairs.
    *
    * To compute the marginal variance of a single variable, simply supply the same variable UUID for both members of
--- a/fuse_graphs/src/hash_graph.cpp
+++ b/fuse_graphs/src/hash_graph.cpp
@@ -289,6 +289,11 @@ void HashGraph::holdVariable(const fuse_core::UUID& variable_uuid, bool hold_con
   }
 }
 
+bool HashGraph::isVariableOnHold(const fuse_core::UUID& variable_uuid) const
+{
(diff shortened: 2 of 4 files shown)


fuse_core/include/fuse_core/graph.h
```diff
@@ -316,6 +316,14 @@ class Graph
    */
   virtual void holdVariable(const UUID& variable_uuid, bool hold_constant = true) = 0;
 
+  /**
+   * @brief Check whether a variable is on hold or not
+   *
+   * @param[in] variable_uuid The variable to test
+   * @return True if the variable is on hold, false otherwise
+   */
+  virtual bool isVariableOnHold(const UUID& variable_uuid) const = 0;
```
ayrton04, line 325: I don't really care, but would `isVariableHeld` be a better name?

ayrton04 (approved):

fuse_core/include/fuse_core/graph.h
efernandez: Yep, I thought about that. The only reason I ended up calling it `isVariableOnHold` is because the graph maintains a set of variables held in an attribute called `variables_on_hold_`.

@svwilliams Do you have any preference?

I'm happy to make the change.

svwilliams: I have no opinion on the name either way. Fight amongst yourselves. 🙂

fuse_graphs/src/hash_graph.cpp
```diff
@@ -250,6 +250,7 @@ bool HashGraph::removeVariable(const fuse_core::UUID& variable_uuid)
   {
     constraints_by_variable_uuid_.erase(cross_reference_iter);
   }
+  variables_on_hold_.erase(variable_uuid);
```
svwilliams, line 253: Thank you for catching that.

svwilliams (approved):

fuse_core/include/fuse_core/graph.h
efernandez: lol

https://coinflip.com :thinking: :wink:

efernandez: @ayrton04 If you prefer `isVariableHeld`, simply let me know and I'll rename this method. No problem. Otherwise, I guess you can merge as is. :smiley:

ayrton04: Oh, this can be merged as-is. I don’t really have strong feelings at all.

PR #188, opened 2020-05-04
Check canTransform output and show error if false

efernandez:
This checks the output of `canTransform` and show the `error_msg` if it returns `false`.

There's also a typo in a message fixed.

### DIFF
--- a/fuse_models/include/fuse_models/common/sensor_proc.h
+++ b/fuse_models/include/fuse_models/common/sensor_proc.h
@@ -932,7 +932,7 @@ inline bool processTwistWithCovariance(
 
   if (!transformMessage(tf_buffer, twist, transformed_message))
   {
-    ROS_ERROR_STREAM("Cannot create constraint from pose message with stamp " << twist.header.stamp);
+    ROS_ERROR_STREAM("Cannot create constraint from twist message with stamp " << twist.header.stamp);
     return false;
   }
 
--- a/fuse_models/include/fuse_models/common/sensor_proc.h
+++ b/fuse_models/include/fuse_models/common/sensor_proc.h
@@ -223,30 +223,19 @@ inline void validatePartialMeasurement(
 template <typename T>
 bool transformMessage(const tf2_ros::Buffer& tf_buffer, const T& input, T& output)
 {
-  geometry_msgs::TransformStamped trans;
-  std::string error_msg;
-  if (tf_buffer.canTransform(output.header.frame_id, input.header.frame_id, input.header.stamp, &error_msg))
+  try
   {
-    try
-    {
-      trans = tf_buffer.lookupTransform(output.header.frame_id, input.header.frame_id, input.header.stamp);
-    }
-    catch (const tf2::TransformException& ex)
-    {
-      ROS_WARN_STREAM_THROTTLE(5.0, "Could not transform message from " << input.header.frame_id << " to " <<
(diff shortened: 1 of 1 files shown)


svwilliams (approved):

fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -237,6 +238,12 @@ bool transformMessage(const tf2_ros::Buffer& tf_buffer, const T& input, T& outpu
       return false;
     }
   }
+  else
```
ayrton04: Might be simpler to assume we return false and return true only if successful:

```
template <typename T>
bool transformMessage(const tf2_ros::Buffer& tf_buffer, const T& input, T& output)
{
  geometry_msgs::TransformStamped trans;
  std::string error_msg;
  if (tf_buffer.canTransform(output.header.frame_id, input.header.frame_id, input.header.stamp, &error_msg))
  {
    try
    {
      trans = tf_buffer.lookupTransform(output.header.frame_id, input.header.frame_id, input.header.stamp);
      tf2::doTransform(input, output, trans);
      return true;
    }
    catch (const tf2::TransformException& ex)
    {
      ROS_WARN_STREAM_THROTTLE(5.0, "Could not transform message from " << input.header.frame_id << " to " <<
        output.header.frame_id << ". Error was " << ex.what());
    }
  }
  else
  {
    ROS_WARN_STREAM("Could not transform message from " << input.header.frame_id << " to " <<
      output.header.frame_id << ". Error before lookup was " << error_msg);
  }

  return false;
}
```

I also wonder if it makes sense to just jump straight into the `try`. I imagine the exception and the error message we get are the same:

```
template <typename T>
bool transformMessage(const tf2_ros::Buffer& tf_buffer, const T& input, T& output)
{
  try
  {
    auto trans = tf_buffer.lookupTransform(output.header.frame_id, input.header.frame_id, input.header.stamp);
    tf2::doTransform(input, output, trans);
    return true;
  }
  catch (const tf2::TransformException& ex)
  {
    ROS_WARN_STREAM_THROTTLE(5.0, "Could not transform message from " << input.header.frame_id << " to " <<
      output.header.frame_id << ". Error was " << ex.what());
  }

  return false;
}
```

ayrton04 (approved):

fuse_models/include/fuse_models/common/sensor_proc.h
efernandez: Changed to only call `lookupTransform`.

I tested the case that failed for me, and it works:
`Could not transform message from sensor_link to base_link. Error was "sensor_link" passed to lookupTransform argument source_frame does not exist.`

That message is printed as a WARNing, and it's throttled.

However, there's another ERROR message with `transformMessage` returns `false`, that isn't throttled. Maybe that should also be changed. :thinking: 
`Cannot create constraint from sensor message with stamp 1587177144.513711276`

Thoughts? Maybe for another PR, or we just leave it as is.

efernandez:
Updated the PR with @ayrton04 suggestion. It's way cleaner and also catches the case that wasn't catch before. A static transform not available.

Hide white space when looking at the diff, since it's actually a simple change.

ayrton04 (approved):

ayrton04 (approved):

svwilliams (approved):

Issue #127, opened 2020-01-07
The fuse_models package needs an optional dependency on `benchmark`


svwilliams:
The fuse_models package uses the `benchmark` library as part of the test build. That library is not available for Xenial/Kinetic. The CMake build system conditionally compiles the benchmark test correctly, but I know of no way to specify a conditional/optional dependency in the `package.xml` file. As such, adding the `benchmark` dependency to the `fuse_models` `package.xml` breaks the Xenial/Kinetic buildfarm build. See PR #124.

PR #189, opened 2020-05-06
Conditionally test_depend on benchmark

efernandez:
This upgrades `package.xml` to version `3` for the `fuse_models`, so we can conditionally `test_depend` on `benchmark` if the ROS distro is `melodic` or newer. This addressed https://github.com/locusrobotics/fuse/issues/127.

### DIFF
--- a/fuse_models/package.xml
+++ b/fuse_models/package.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0"?>
-<package format="2">
+<?xml-model
+  href="http://download.ros.org/schema/package_format3.xsd"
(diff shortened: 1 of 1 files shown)


svwilliams (approved):

ayrton04 (approved):

ayrton04:
I wonder if we should bloom this package at some point.

efernandez:
> I wonder if we should bloom this package at some point.

@ayrton04 That's probably a good idea, so we can confirm this bump to package format 3 doesn't break anything related to the blooming tools.

Maybe it's worth pinging someone from you build infrastructure team into this PR, just in case this could break something on their side.

PR #195, opened 2020-05-22
Add fuse_models::Transaction sensor model

efernandez:
* Add `fuse_models::Transaction` sensor model

I'm going to send another PR with a `fuse_models::GraphIgnition` that complements this one and also adds a rostest. This should hopefully motivate https://github.com/locusrobotics/fuse/pull/183 a bit more, although I understand the last comments there and they might actually be an issue. I still have to look at that.

I'm open to other approaches to load a recorded graph and transactions.

PR #196, opened 2020-05-22
Add fuse_models::GraphIgnition sensor model

efernandez:
* Add `fuse_models::GraphIgnition` sensor model

This complements https://github.com/locusrobotics/fuse/pull/195, so we can configure a `fixed_lag_smoother` that ignites when it receives a recorded graph and then apply recorded transactions into it.

This should help motivate #183 a bit more, and also help to test things out, if we agree on using these two sensor models to play back recorded graphs and transactions. This is certainly open to discussion, mostly because #183 is required, and it also looks that even with #183 there are still some sharp edges, so things don't always work because we don't process the correct transaction after the graph. Remember the intention is to throttle the graphs while recording, because recording all graphs takes a lot of disk and bandwidth.

### DIFF
--- a/fuse_models/test/test_graph_ignition.cpp
+++ b/fuse_models/test/test_graph_ignition.cpp
@@ -1,0 +1,411 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2020, Clearpath Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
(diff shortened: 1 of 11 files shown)


efernandez:
@ayrton04 @svwilliams Do you mind to also copy-paste here the check errors for this one, please?

#195 Is passing now. Maybe this needs to be on top of it, but I don't think so. It could be another test failing, since I only made sure it compiles. Note that I moved both the **Transaction** and **GraphIgnition** sensor models from an internal repository, and I've already been using them w/o problems. I probably missed something else while moving things here.

fuse_models/test/test_graph_ignition.cpp
```diff
@@ -0,0 +1,411 @@
...
+    }
+    catch (const std::out_of_range& ex)
+    {
+      ADD_FAILURE() << ex.what();
+    }
+  }
+
+  // Since the variables in the graph do not have a stamp, the transaction should have a single involved stamp, equal to
+  // the transaction stamp, that should also be equal to the requested graph message stamp
+  ASSERT_EQ(1, boost::size(transaction->involvedStamps()));
```
ayrton04: Build doesn't like this for Bionic builds, because it's comparing signed and unsigned values:

```
ASSERT_EQ(1u, boost::size(transaction->involvedStamps()));
```

efernandez: Thanks. Fixed.

If that was the only error, now it should pass.

fuse_models/test/test_graph_ignition.cpp
```diff
@@ -0,0 +1,411 @@
...
+/**
+ * @brief Static variable to hold the last unit test error description
+ */
+static std::string failure_description;
+
+/**
+ * @brief Compare all the properties of two Variable objects
+ * @return True if all the properties match, false otherwise
+ */
+bool compareVariables(const fuse_core::Variable& expected, const fuse_core::Variable& actual)
```
svwilliams, line 79: Maybe I'll make gtest macros for these one day. Thanks.

svwilliams (approved):

ayrton04 (approved):

efernandez:
The tests passed now. :tada: 

So it looks like this is ready to merge. :smiley:

fuse_models/test/test_graph_ignition.cpp
efernandez: I wonder if you'd create a header in `fuse_core/include/fuse_core` for that, or create a new `fuse_test` package for that. :thinking: 

I don't think we can access headers define in `fuse_core/test`, if we move them there. If we actually can, then that would be another option. Probably the best approach.

PR #204, opened 2020-11-13
Support throttling serialized graph publisher

efernandez:
### Motivation

I've been throttling the serialized graph message published by the `fuse_publishers::SerializedPublisher` using the `throttle` node from `topic_tools`, so I only record a few graphs for debugging purposes. Recording all would take too much disk and bandwidth.

So far so good, but the serialized publisher is still serializing and publishing all the graphs it gets notified from the optimizer. It turns out that serializing the graph is quite fast, but it still takes a noticeable amount of time. If the optimization frequency is high, we could get too many graphs and the impact becomes more significant.

Since I'm throttling the serialized graph messages after anyway, it makes more sense to throttle the publisher directly in the serialized publisher, so we can save some cycles.

### Changelog

* Move the `fuse_models::ThrottledCallback` to `fuse_core`
* Support generic callbacks in `ThrottledCallback`. An alias template called `ThrottledMessageCallback` is provided for ROS messages. Unfortunately, the automatic type deduction in `ros::NodeHandle::subcribe()` no longer works due to the generic `ThrottledCallback::callback()` we have now. For this reason we now have to do `subscribe<Message>()` instead of just `subscribe()`. If you know how to make this cleaner, I'd like to know. :smiley: 
* Add support to throttle graph publishing. It doesn't throttle any graph by default. BTW, I'm happy to create a `parameters` folder and a class to encapsulate the `fuse_publishers::SerializedPublisher` configuration, as it's done in `fuse_models`.

### Results

If the optimization frequency or the rate of the sensor models inputs are low, the impact of this is very low. I've evaluated this with the following relevant configuration:
* 50Hz optimization frequency
* A bunch of sensor models inputs that when all are processed and the graph isn't throttled, the graph is published at approx. 30Hz

The plots below show the CPU% of the fuse optimizer node in blue, and the average in red:
* No graph publisher throttling:
C4-2020-11-13_18-58-25_01_plot
* Graph publisher throttled, with a throttle period of 30s, which is a reasonable value for use case described in the motivation section (I expect similar results with smaller throttling periods though):
C5-2020-11-13_19-01-16_01_plot

We save approx. 10% of CPU.

### DIFF
--- a/fuse_publishers/src/serialized_publisher.cpp
+++ b/fuse_publishers/src/serialized_publisher.cpp
@@ -36,6 +36,7 @@
 #include <fuse_core/async_publisher.h>
 #include <fuse_core/graph.h>
 #include <fuse_core/graph_deserializer.h>
+#include <fuse_core/parameter.h>
 #include <fuse_core/transaction.h>
 #include <fuse_core/transaction_deserializer.h>
 #include <fuse_msgs/SerializedGraph.h>
@@ -52,7 +53,9 @@ namespace fuse_publishers
 
 SerializedPublisher::SerializedPublisher() :
   fuse_core::AsyncPublisher(1),
-  frame_id_("map")
+  frame_id_("map"),
+  graph_publisher_throttled_callback_(
+      std::bind(&SerializedPublisher::graphPublisherCallback, this, std::placeholders::_1, std::placeholders::_2))
 {
 }
 
@@ -63,6 +66,17 @@ void SerializedPublisher::onInit()
 
   bool latch = false;
   private_node_handle_.getParam("latch", latch);
+
+  ros::Duration graph_throttle_period { 0.0 };
+  double graph_throttle_period_double = graph_throttle_period.toSec();
+  fuse_core::getPositiveParam(private_node_handle_, "graph_throttle_period", graph_throttle_period_double, false);
+  graph_throttle_period.fromSec(graph_throttle_period_double);
+
+  bool graph_throttle_use_wall_time { false };
+  private_node_handle_.getParam("graph_throttle_use_wall_time", graph_throttle_use_wall_time);
+
+  graph_publisher_throttled_callback_.setThrottlePeriod(graph_throttle_period);
+  graph_publisher_throttled_callback_.setUseWallTime(graph_throttle_use_wall_time);
 
   // Advertise the topics
   graph_publisher_ = private_node_handle_.advertise<fuse_msgs::SerializedGraph>("graph", 1, latch);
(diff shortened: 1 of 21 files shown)


fuse_publishers/src/serialized_publisher.cpp
```diff
@@ -64,6 +67,17 @@ void SerializedPublisher::onInit()
   bool latch = false;
   private_node_handle_.getParam("latch", latch);
 
+  ros::Duration graph_throttle_period { 0.0 };
+  double graph_throttle_period_double = graph_throttle_period.toSec();
+  fuse_core::getPositiveParam(private_node_handle_, "graph_throttle_period", graph_throttle_period_double, false);
```
ayrton04: Not for this PR, but as long as we have utility parameter loading methods, having one that loads durations would be handy.

ayrton04 (approved):

fuse_publishers/src/serialized_publisher.cpp
efernandez: That would be handy.

I tried to provide one, and also overload `fuse_core::getPositiveParam()` to support `ros::Duration`, but I didn't come with a good solution. I was indeed trying to support `ros::Duration` in the templated `fuse_core::getPositiveParam()` function, but I believe I ran into some issues and I didn't invest more time trying to make it work.

svwilliams: Can't we just do this?
```
/**
 * @brief Helper function that loads positive duration values from the parameter server
 *
 * @param[in] node_handle - The node handle used to load the parameter
 * @param[in] parameter_name - The parameter name to load
 * @param[in, out] default_value - A default value to use if the provided parameter name does not exist. As output it
 *                                 has the loaded (or default) value
 * @param[in] strict - Whether to check the loaded value is strictly positive or not, i.e. whether 0 is accepted or not
 */
void getPositiveParam(const ros::NodeHandle& node_handle, const std::string& parameter_name,
                      ros::Duration& default_value, const bool strict = true)
{
  double default_value_sec = default_value.toSec();
  getPositiveParam(node_handle, parameter_name, default_value_sec, strict);
  default_value.fromSec(default_value_sec);
}
```

svwilliams (approved):

fuse_publishers/src/serialized_publisher.cpp
efernandez: Done, that works if I make it `inline`.

svwilliams: You say "Done" but I don't see any new commits. :slightly_smiling_face:

efernandez: I pushed them to the wrong branch. Should be good now. :sweat_smile:

efernandez:
Not sure why the checks failed, so I've rebase this PR on top of https://github.com/locusrobotics/fuse/pull/207

svwilliams (approved):

PR #206, opened 2020-11-20
Add invert_tf to Odometry2DPublisher

efernandez:
This allows to publish the inverse TF transform.

This can be useful to skip the `odom->base_link` `lookupTransform()` when
braodcasting `map->base_link`, which must be broadcasted as `map->odom`
because TF tree doesn't support more than a single parent per frame,
`base_link` in this case.

This is particular relevant when
`predict_to_current_time` is enabled, because the `lookupTransform()` could
take a while, causing delays.

### DIFF
--- a/fuse_models/src/odometry_2d_publisher.cpp
+++ b/fuse_models/src/odometry_2d_publisher.cpp
@@ -47,6 +47,7 @@
 #include <tf2_ros/buffer.h>
 #include <tf2_ros/transform_listener.h>
 
+#include <algorithm>
 #include <memory>
 #include <mutex>
 #include <string>
@@ -75,7 +76,7 @@ void Odometry2DPublisher::onInit()
 
   params_.loadFromROS(private_node_handle_);
 
-  if (params_.world_frame_id == params_.map_frame_id)
+  if (!params_.invert_tf && params_.world_frame_id == params_.map_frame_id)
   {
     tf_buffer_ = std::make_unique<tf2_ros::Buffer>(params_.tf_cache_time);
     tf_listener_ = std::make_unique<tf2_ros::TransformListener>(*tf_buffer_, node_handle_);
@@ -446,15 +447,25 @@ void Odometry2DPublisher::publishTimerCallback(const ros::TimerEvent& event)
 
   if (params_.publish_tf)
   {
+    auto frame_id = odom_output.header.frame_id;
+    auto child_frame_id = odom_output.child_frame_id;
+
+    if (params_.invert_tf)
+    {
+      pose = pose.inverse();
+      std::swap(frame_id, child_frame_id);
+    }
+
     geometry_msgs::TransformStamped trans;
-    trans.header = odom_output.header;
-    trans.child_frame_id = odom_output.child_frame_id;
+    trans.header.stamp = odom_output.header.stamp;
+    trans.header.frame_id = frame_id;
+    trans.child_frame_id = child_frame_id;
     trans.transform.translation.x = pose.getX();
     trans.transform.translation.y = pose.getY();
     trans.transform.translation.z = odom_output.pose.pose.position.z;
     trans.transform.rotation = tf2::toMsg(pose.getRotation());
 
-    if (params_.world_frame_id == params_.map_frame_id)
+    if (!params_.invert_tf && params_.world_frame_id == params_.map_frame_id)
     {
       try
       {
--- a/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
+++ b/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
@@ -73,6 +73,7 @@ public:
   void loadFromROS(const ros::NodeHandle& nh) final
   {
     nh.getParam("publish_tf", publish_tf);
+    nh.getParam("invert_tf", invert_tf);
     nh.getParam("predict_to_current_time", predict_to_current_time);
     nh.getParam("predict_with_acceleration", predict_with_acceleration);
     nh.getParam("publish_frequency", publish_frequency);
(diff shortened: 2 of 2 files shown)


efernandez:
Is this failing because it can't find `gmock`?

It looks like the dependency is missed, and also the include headers and libraries path is not discovered in `CMakeLists.txt`.

efernandez:
If the checks are failing because of `gmock`, as they did for me locally, https://github.com/locusrobotics/fuse/pull/207 should hopefully make them pass. :crossed_fingers:

fuse_models/src/odometry_2d_publisher.cpp
```diff
@@ -75,7 +75,8 @@ void Odometry2DPublisher::onInit()
 
   params_.loadFromROS(private_node_handle_);
 
-  if (params_.world_frame_id == params_.map_frame_id)
+  const auto frame_id = params_.invert_tf ? params_.base_link_output_frame_id : params_.world_frame_id;
+  if (frame_id == params_.map_frame_id)
```
svwilliams: Regardless of the state of the `invert_tf` flag, the odom message will use the world_frame_id as the parent frame and base_link_output_frame_id as the child frame. That makes me think this check should continue to be:
```
if (params_.world_frame_id == params_.map_frame_id)
``` 
regardless of the `invert_tf` flag.

What I'm having trouble working out is what should this do?
```
world_frame_id: map
base_link_output_frame_id: base_link
invert_tf: true
```
The odometry message will use (parent frame "map", child frame "base_link").
Without the `invert_tf` flag, tf would publish (parent frame "map", child frame "odom").
Does that mean with the `invert_tf`, tf should publish (parent frame "odom", child frame "map")? But I have no idea why anyone would ever want that. Do you have a use-case in mind?

But with the code as-is, I do not think it is computing a valid transform. The base_link->odom lookup is not modified by the `invert_tf` flag, so it is always base_link->odom. The transform `trans` is modified by the `invert_tf` flag, so `trans` is really base_link->map. And the two get multiplied together here (https://github.com/locusrobotics/fuse/blob/devel/fuse_models/src/odometry_2d_publisher.cpp#L468):
```
base_link->map * base_link->odom
```
which I do not believe yields anything meaningful.

If publishing odom->map is the expected output, I think it would be easier to handle the invert logic just before the broadcast. So all of the code would remain identical, and right before this (https://github.com/locusrobotics/fuse/blob/devel/fuse_models/src/odometry_2d_publisher.cpp#L481), you add an `if(params_.invert_tf)` block that computes the inverse transform. It's a little annoying because you have to bounce back and forth between tf and geometry_msgs. There's probably a more clever way to do this with fewer translations. Converting the tf lookup results to a tf Transform maybe? And delaying the conversion from tf Transform to geometry_msgs until the very end?

svwilliams:
Build error:
```
fuse_models/src/odometry_2d_publisher.cpp:456:  Add #include <algorithm> for swap  [build/include_what_you_use] [4]
```

efernandez:
Added #include <algorithm> for swap, but I think #207 is also needed for the tests to pass

fuse_models/src/odometry_2d_publisher.cpp
efernandez: With `invert_tf` enabled I do `std::swap(frame_id, child_frame_id);`, so the block `if (frame_id == params_.map_frame_id)` compares `base_link` with `map`, so it doesn't compute `base_link->map * base_link->odom`.

The output in that case is simply `base_link->map`.

The whole purpose here is to avoid calling `lookupTransform()`, i.e. skip the need to broadcast wrt `odom` due to TF tree limitations.

ayrton04 (approved):

fuse_models/src/odometry_2d_publisher.cpp
svwilliams: I feel like what you are saying is, if the "invert_tf" flag it true, we should not create the tf listener or do the tf lookup. Can we just say that in the code, instead of swapping frame names around?

```
if (!params_.invert_tf && params_.world_frame_id == params_.map_frame_id)
```

fuse_models/src/odometry_2d_publisher.cpp
```diff
@@ -446,15 +448,25 @@ void Odometry2DPublisher::publishTimerCallback(const ros::TimerEvent& event)
...
+    trans.header.stamp = odom_output.header.stamp;
+    trans.header.frame_id = frame_id;
+    trans.child_frame_id = child_frame_id;
     trans.transform.translation.x = pose.getX();
     trans.transform.translation.y = pose.getY();
     trans.transform.translation.z = odom_output.pose.pose.position.z;
     trans.transform.rotation = tf2::toMsg(pose.getRotation());
 
-    if (params_.world_frame_id == params_.map_frame_id)
+    if (frame_id == params_.map_frame_id)
```
svwilliams: And then here as well
```
if (!params_.invert_tf && params_.world_frame_id == params_.map_frame_id)
```

fuse_models/src/odometry_2d_publisher.cpp
efernandez: Exactly, that's the intend. Changed, so it's easier to read it in the code. :+1:

fuse_models/src/odometry_2d_publisher.cpp
efernandez: Done, but in the code above I still `std::swap` the frames `if (params_.invert_tf)`, since I believe that makes the implementation simpler.

svwilliams: Yeah, sorry. That swap makes complete sense. But this is much easier to read now. Thank you.

svwilliams (approved):

PR #209, opened 2020-11-25
Add ability to throttle covariance computation

efernandez:
This PR adds the `covariance_throttle_period` ROS param to the `fuse_models::Odometry2DPublisher` to allow throttling the covariance computation. The motivation is to reduce the CPU usage by saving time not computing the covariance on every optimization cycle, i.e. every time `notifyCallback()` is called.

I've profiled the fixed-lag smoother, and the two major hotspots are Ceres solver optimization and the covariance computation, taking around half of the total computational time each.

The idea here is that we could reduce the CPU usage to almost half if we don't need the covariance at the same rate as the state. For example, if we set `covariance_throttle_period` to `2.0s`, we get approx. half the CPU usage.

By default `covariance_throttle_period` is `0.0`, which doesn't throttle the covariance computation, so we have the current behaviour.

Note that when the covariance isn't computed, we simply used the last one computed. We don't predict/propagate it forward (see code comment for more details and justification). It's still propagated if `predict_to_current_time` is enabled, but doesn't only for the small `dt` between the optimization and publish time.

### DIFF
--- a/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
+++ b/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
@@ -81,6 +81,8 @@ public:
     nh.param("scale_process_noise", scale_process_noise, scale_process_noise);
     nh.param("velocity_norm_min", velocity_norm_min, velocity_norm_min);
 
+    fuse_core::getPositiveParam(nh, "covariance_throttle_period", covariance_throttle_period, false);
+
     double tf_cache_time_double = tf_cache_time.toSec();
     nh.getParam("tf_cache_time", tf_cache_time_double);
     tf_cache_time.fromSec(tf_cache_time_double);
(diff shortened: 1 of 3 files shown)


svwilliams (approved):

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
```diff
@@ -81,6 +81,10 @@ struct Odometry2DPublisherParams : public ParameterBase
     nh.param("scale_process_noise", scale_process_noise, scale_process_noise);
     nh.param("velocity_norm_min", velocity_norm_min, velocity_norm_min);
 
+    double covariance_throttle_period_double = covariance_throttle_period.toSec();
+    fuse_core::getPositiveParam(nh, "covariance_throttle_period", covariance_throttle_period_double, false);
+    covariance_throttle_period.fromSec(covariance_throttle_period_double);
```
efernandez: I'll rebase and change this to 
```suggestion
    fuse_core::getPositiveParam(nh, "covariance_throttle_period", covariance_throttle_period, false);
```
when #204 is in.

efernandez: Done.

ayrton04 (approved):

PR #210, opened 2020-11-26
Use empty transaction

efernandez:
This is a trivial simplification, although I had to change the implementation of `Transaction::empty()` a bit so it doesn't check the `involved_stamps_`, and I updated the unit test accordingly.

### DIFF
--- a/fuse_optimizers/src/batch_optimizer.cpp
+++ b/fuse_optimizers/src/batch_optimizer.cpp
@@ -169,10 +169,7 @@ void BatchOptimizer::optimizerTimerCallback(const ros::TimerEvent& /*event*/)
   // Check if there is any pending information to be applied to the graph.
   {
     std::lock_guard<std::mutex> lock(combined_transaction_mutex_);
-    optimization_request_ = !combined_transaction_->addedConstraints().empty() ||
-                            !combined_transaction_->removedConstraints().empty() ||
-                            !combined_transaction_->addedVariables().empty() ||
-                            !combined_transaction_->removedVariables().empty();
+    optimization_request_ = !combined_transaction_->empty();
   }
   // If there is some pending work, trigger the next optimization cycle.
   // If the optimizer has not completed the previous optimization cycle, then it


fuse_core/test/test_transaction.cpp
```diff
@@ -339,14 +339,14 @@ TEST(Transaction, Empty)
     EXPECT_FALSE(transaction.empty());
   }
 
-  // A transaction with involved stamps cannot be empty
+  // A transaction with involved stamps but no added or removed variables or constraints is empty
```
efernandez: A transaction with involved stamps only doesn't make much sense though. Not sure if there's an easy and cheap way to enforce that. I believe it's better to leave it as is.

fuse_optimizers/src/batch_optimizer.cpp
```diff
@@ -169,10 +169,7 @@ void BatchOptimizer::optimizerTimerCallback(const ros::TimerEvent& /*event*/)
   // Check if there is any pending information to be applied to the graph.
   {
     std::lock_guard<std::mutex> lock(combined_transaction_mutex_);
-    optimization_request_ = !combined_transaction_->addedConstraints().empty() ||
-                            !combined_transaction_->removedConstraints().empty() ||
-                            !combined_transaction_->addedVariables().empty() ||
-                            !combined_transaction_->removedVariables().empty();
+    optimization_request_ = !combined_transaction_->empty();
```
svwilliams, line 172: Are you just looking to simplify this line? Because this line probably predates the creation of the `Transaction.empty()` method. The fixed-lag smoother uses the old definition of `Transaction.empty()` in a very similar way.
https://github.com/locusrobotics/fuse/blob/devel/fuse_optimizers/src/fixed_lag_smoother.cpp#L242

I think we should leave the `Transaction.empty()` method alone, and keep just this change to `optimization_request_`.

Unless you want to change the definition of `Transaction.empty()`. In which case, what is reason? I'm not necessarily against dropping the "involved_stamps" check, I just want to understand what is causing problems.

efernandez: Nothing was causing problems, just looking to simplify this line. `Transaction::empty()` was introduced after this line, yes.

I simply remove the check for `involved_stamps_.empty()` in `Transaction::empty()`, so it matches the checks done here, and because I don't think it's really needed.

svwilliams: If it wasn't causing any issues, I'd prefer to leave the `Transaction::empty()` definition as it was. So `empty()` really does mean empty. But I'm all for changing this line to `optimization_request_ = !combined_transaction_->empty();`, even though it changes the criteria slightly. This brings it in line with the fixed-lag smoother.

efernandez: Works for me. I've just removed the commit that changed the `Transaction::empty()` definition.

svwilliams (approved):

ayrton04 (approved):

PR #212, opened 2020-11-30
Use fuse_core::getPositiveParam for all ros::Duration parameters

efernandez:
When I tried to use `fuse_core::getPositiveParam` for all `ros::Duration` parameters before, I searched for `_sec`, so I missed a few spots where we use `_double` instead. Now all `ros::Duration` parameters make use of the new `fuse_core::getPositiveParam` that supports `ros::Duration`.

### DIFF
--- a/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
+++ b/fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
@@ -84,13 +84,8 @@ public:
 
     fuse_core::getPositiveParam(nh, "covariance_throttle_period", covariance_throttle_period, false);
 
-    double tf_cache_time_double = tf_cache_time.toSec();
-    nh.getParam("tf_cache_time", tf_cache_time_double);
-    tf_cache_time.fromSec(tf_cache_time_double);
-
-    double tf_timeout_double = tf_timeout.toSec();
-    nh.getParam("tf_timeout", tf_timeout_double);
-    tf_timeout.fromSec(tf_timeout_double);
+    fuse_core::getPositiveParam(nh, "tf_cache_time", tf_cache_time, false);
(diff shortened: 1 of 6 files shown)


fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
```diff
@@ -82,13 +82,8 @@ struct Odometry2DPublisherParams : public ParameterBase
...
     nh.param("velocity_norm_min", velocity_norm_min, velocity_norm_min);
 
-    double tf_cache_time_double = tf_cache_time.toSec();
-    nh.getParam("tf_cache_time", tf_cache_time_double);
-    tf_cache_time.fromSec(tf_cache_time_double);
-
-    double tf_timeout_double = tf_timeout.toSec();
-    nh.getParam("tf_timeout", tf_timeout_double);
-    tf_timeout.fromSec(tf_timeout_double);
+    fuse_core::getPositiveParam(nh, "tf_cache_time", tf_cache_time, false);
```
efernandez, line 87: I've also used `fuse_core::getPositiveParam` for these two, that didn't required the input to be positive.

I believe they must be positive, and I've allowed `0.0` because:
* tf_cache_time: A TF buffer with a time cache that only spans for `0.0` seconds probably doesn't make much sense. I believe in this case only the latest transform is kept in the buffer.
* tf_timeout: A `lookupTransform` with `0.0` timeout will simply not wait for the transform to be available.

If you don't agree with any of this, I'm happy to change it.

svwilliams: Looks good to me.

svwilliams (approved):

ayrton04 (approved):

svwilliams:
Looks like this needs a rebase.

efernandez:
Rebased

PR #216, opened 2020-12-08
Transform message in differential mode

efernandez:
This is important because the relative transformation is not the same if the **sensor** and **target** frame are different.

Consider for example the case of an IMU sensor upside down:
* The robot base frame is `base_link`
* The IMU sensor frame is `imu_link`
* The `imu_link` transformation wrt `base_link` is `180` degrees wrt the `y` or `x` axis
* The angular velocity around the `z` axis has **opposite sign** in the **sensor** frame wrt the **target** frame

This only impacts the `Imu2D`, `Pose2D` and `Odometry2D` sensor models. The changes are similar for each, but with some differences. For this reason I've refrained from trying to factorize things out somehow. TBH it doesn't look trivial and probably not worth it, but at the same time I think there are too many `if-else` blocks and indentation levels. Any thoughts?

I'd be happy to make the implementation of this logic cleaner/simpler. :smiley: 

For now I advise you to hide the whitespace changes when you inspect the changes.

### DIFF
--- a/fuse_models/include/fuse_models/parameters/odometry_2d_params.h
+++ b/fuse_models/include/fuse_models/parameters/odometry_2d_params.h
@@ -89,11 +89,12 @@ struct Odometry2DParams : public ParameterBase
         fuse_core::getParamRequired(nh, "twist_target_frame", twist_target_frame);
       }
 
-      if (!differential && (!position_indices.empty() || !orientation_indices.empty()))
+      if (!position_indices.empty() || !orientation_indices.empty())
       {
         fuse_core::getParamRequired(nh, "pose_target_frame", pose_target_frame);
       }
-      else
+
+      if (differential)
       {
         nh.getParam("independent", independent);
         nh.getParam("use_twist_covariance", use_twist_covariance);
--- a/fuse_models/src/imu_2d.cpp
+++ b/fuse_models/src/imu_2d.cpp
@@ -136,41 +136,63 @@ void Imu2D::process(const sensor_msgs::Imu::ConstPtr& msg)
 
   if (params_.differential)
   {
-    if (previous_pose_)
-    {
-      if (params_.use_twist_covariance)
-      {
-        common::processDifferentialPoseWithTwistCovariance(
-          name(),
-          device_id_,
-          *previous_pose_,
-          *pose,
-          twist,
-          params_.minimum_pose_relative_covariance,
-          params_.pose_loss,
-          {},
-          params_.orientation_indices,
-          validate,
-          *transaction);
-      }
-      else
-      {
-        common::processDifferentialPoseWithCovariance(
-          name(),
-          device_id_,
-          *previous_pose_,
-          *pose,
-          params_.independent,
-          params_.minimum_pose_relative_covariance,
-          params_.pose_loss,
-          {},
-          params_.orientation_indices,
-          validate,
-          *transaction);
-      }
-    }
+    auto transformed_pose = std::make_unique<geometry_msgs::PoseWithCovarianceStamped>();
+    transformed_pose->header.frame_id = params_.orientation_target_frame;
 
-    previous_pose_ = std::move(pose);
+    if (!common::transformMessage(tf_buffer_, *pose, *transformed_pose))
+    {
+      ROS_ERROR_STREAM("Cannot transform pose message with stamp "
+                       << pose->header.stamp << " to orientation target frame " << params_.orientation_target_frame);
+    }
+    else
+    {
+      if (previous_pose_)
+      {
+        if (params_.use_twist_covariance)
+        {
+          geometry_msgs::TwistWithCovarianceStamped transformed_twist;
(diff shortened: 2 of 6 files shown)


fuse_models/include/fuse_models/parameters/odometry_2d_params.h
```diff
@@ -89,11 +89,12 @@ struct Odometry2DParams : public ParameterBase
         fuse_core::getParamRequired(nh, "twist_target_frame", twist_target_frame);
       }
 
-      if (!differential && (!position_indices.empty() || !orientation_indices.empty()))
+      if (!position_indices.empty() || !orientation_indices.empty())
```
efernandez, line 92: This change makes the `pose_target_frame` required for the `differential` mode too.

Another option would be to skip the transformation is `pose_target_frame.empty()`, i.e. if not set (it defaults to empty). That could be a change applied to all instances of any type of target frame parameter. Thoughts?

efernandez:
What's the error in the CI checks, please? I can't reproduce it offline.

fuse_models/include/fuse_models/parameters/odometry_2d_params.h
ayrton04: I like making it required.

ayrton04 (approved):

fuse_models/include/fuse_models/parameters/odometry_2d_params.h
efernandez: That's fine, but I wonder why you prefer it required as opposed to optional with the default falling back to the input message `header.frame_id`.

In a configuration with multiple odometry sources (e.g. wheel, laser, visual) we'd have to specify the `pose_target_frame` for each different odometry frame (e.g. `wheel_odom`, `laser_odom`, `visual_odom`), even for the `differential` mode with this PR. This means that any existing configuration should be updated, so the node doesn't terminate during the parameter loading step. If this is well understood and acceptable, then I'm happy to update my configurations adding the `pose_target_frame` where needed.

efernandez:
Still don't know why the checks failed on this one. They passed on #217 , which is on top of this one. Maybe the CI test job should be re-kicked. :thinking:

svwilliams:
The build error is: 
> C++ exception with description "Could not find required parameter pose_target_frame in namespace /Optimizer/wheel_odometry" thrown in the test body.

from here: https://github.com/locusrobotics/fuse/blob/devel/fuse_optimizers/test/test_optimizer.cpp#L49

This PR makes the "pose_target_frame" parameter required for the Odometry2D sensor. Previously it was not required when in differential mode. The configuration for the above test was not updated to include that required parameter:
https://github.com/locusrobotics/fuse/blob/devel/fuse_optimizers/test/config/struct/common_robot_config.yaml#L28-L33

efernandez:
Thanks, that makes sense. Fixed.

fuse_models/include/fuse_models/parameters/odometry_2d_params.h
ayrton04: I suppose I just like being explicit. We could change the parameter loading so that it does a `find`, too. Then you could have a common `pose_target_frame` that is used by all pose sources, and if you want to override it for any of them, you can.

To be clear, I don't feel strongly about this _at all_, so if you like using the "empty means skip transformation" paradigm, then that's fine with me. :+1:

svwilliams (approved):
>  I think there are too many if-else blocks and indentation levels

The only thing that's coming to mind is factoring out some of the `if-else` blocks into a separate function. For example, in `imu_2d.cpp`, if all of the differential logic was contained inside a function call, then you would end up with a pleasing:
```
  if (params_.differential)
  {
    some_new_function(probably, with, lots, of, parameters);
  }
  else
  {
    common::processAbsolutePoseWithCovariance(
      name(),
      device_id_,
      *pose,
      params_.pose_loss,
      params_.orientation_target_frame,
      {},
      params_.orientation_indices,
      tf_buffer_,
      validate,
      *transaction);
  }
```

Mostly that function call would just be moving/hiding the nesting levels. But it would allow you to exit early if needed.
```
void some_new_function(probably, with, lots, of, parameters)
{
  if (!common::transformMessage(tf_buffer_, *pose, *transformed_pose))
  {
    ROS_ERROR_STREAM("Cannot transform pose message with stamp "
      << pose->header.stamp << " to orientation target frame "
      <<  params_.orientation_target_frame);
    return;
  }
  if (!previous_pose_)
  {
    previous_pose_ = std::move(transformed_pose);
    return;
  }
  if (params_.use_twist_covariance)
  {
    // Do real work here
  }
  else
  {
    // Do other real work here
  }
``` 

To be very clear, I'm not requesting you do this as part of this PR. I'm happy enough as-is. But you did ask.

efernandez:
Yes, I agree it'd look simpler and cleaner moving things a helper function.

Since both this and #217 are approved, I'll make that refactoring change on another PR. :smiley:

PR #217, opened 2020-12-08
Target frame optional

efernandez:
On top of https://github.com/locusrobotics/fuse/pull/216, this also makes the target frame parameters optional.

If not set, the target frame parameters default to `empty()`. In that case not transformation is performed. In this implementation, I actually fallback to the message `header.frame_id`, because it looks cleaner, but we could also consider bigger changes to factorize things out.

### DIFF
--- a/fuse_models/include/fuse_models/common/sensor_proc.h
+++ b/fuse_models/include/fuse_models/common/sensor_proc.h
@@ -273,12 +273,19 @@ inline bool processAbsolutePoseWithCovariance(
   }
 
   geometry_msgs::PoseWithCovarianceStamped transformed_message;
-  transformed_message.header.frame_id = target_frame;
-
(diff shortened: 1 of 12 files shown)


ayrton04 (approved):

svwilliams (approved):

svwilliams:
This PR now needs a rebase. I really wish git would handle these squash-and-merge cases more gracefully.

efernandez:
That's ok. Indeed, in this case three 3-way merges got applied when I rebased it, so I guess that prevents github from running an automatic rebase/merge itself. At least in this case, my experience tells me that's better to review the result of the the rebase/3-way merge. That's what I've done and I believe things are fine.

efernandez:
I've just sent a commit to remove the optional `target_frame` parameters in the `fuse_optimizers` tests.

PR #219, opened 2020-12-15
Factorize differential mode processing

efernandez:
As discussed in #216 and suggested in https://github.com/locusrobotics/fuse/pull/216#pullrequestreview-549689323, this moves the `differential` mode processing into a separate method, so its a bit cleaner.

### DIFF
--- a/fuse_models/src/imu_2d.cpp
+++ b/fuse_models/src/imu_2d.cpp
@@ -221,8 +221,9 @@ void Imu2D::processDifferential(const geometry_msgs::PoseWithCovarianceStamped& 
 
   if (!common::transformMessage(tf_buffer_, pose, *transformed_pose))
   {
-    ROS_ERROR_STREAM("Cannot transform pose message with stamp "
-                     << pose.header.stamp << " to orientation target frame " << params_.orientation_target_frame);
+    ROS_WARN_STREAM_THROTTLE(5.0, "Cannot transform pose message with stamp " << pose.header.stamp
+                                                                              << " to orientation target frame "
+                                                                              << params_.orientation_target_frame);
     return;
   }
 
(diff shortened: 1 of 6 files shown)


fuse_models/src/imu_2d.cpp
```diff
@@ -269,4 +211,71 @@ void Imu2D::process(const sensor_msgs::Imu::ConstPtr& msg)
...
+                                const geometry_msgs::TwistWithCovarianceStamped& twist, const bool validate,
+                                fuse_core::Transaction& transaction)
+{
+  auto transformed_pose = std::make_unique<geometry_msgs::PoseWithCovarianceStamped>();
+  transformed_pose->header.frame_id =
+      params_.orientation_target_frame.empty() ? pose.header.frame_id : params_.orientation_target_frame;
+
+  if (!common::transformMessage(tf_buffer_, pose, *transformed_pose))
+  {
+    ROS_ERROR_STREAM("Cannot transform pose message with stamp "
```
svwilliams: Should these be throttled? On one hand, they should not happen. On the other hand, if tf is misconfigured/having issues it tends to be very chatty.

svwilliams (approved):

fuse_models/src/imu_2d.cpp
efernandez: Fair enough. Changed so it matches the logging message that already is in `common::transformMessage()`:
https://github.com/locusrobotics/fuse/blob/9119624be8f0837a6d43c3b70b7c33d118d3a6da/fuse_models/include/fuse_models/common/sensor_proc.h#L234-L235
That is, `5s` throttling and `WARN` level.

Maybe we could simply remove the logging message from here, but then I believe it'd be hard to figure out which sensor model generated it. Just thinking out loud.

efernandez: @ayrton04 Does it look good for you like this?

ayrton04 (approved):

efernandez:
ping

fuse_models/src/imu_2d.cpp
ayrton04: LGTM!

ayrton04 (approved):

PR #221, opened 2021-01-28
Use rviz::Axes::updateAlpha

efernandez:
This is required with the latest `rviz` version because now it does not have the `static` `getDefaultXColor()`, `getDefaultYColor()` and `getDefaultZColor()` methods.

The change was introduced in https://github.com/ros-visualization/rviz/commit/a4667509f8a9649039cbce5f6e1b3f053657bf05#diff-7825aa999d905edcd690f7b42fee90ed281e573de8481dfa915af235e07004f7R108-R113

This is part of https://github.com/ros-visualization/rviz/tree/noetic-devel, which also requires https://github.com/ros-visualization/interactive_markers/tree/noetic-devel

This PR might fail the tests if those aren't the versions in the buildfarm. I've confirmed this PR builds fine in a new workspace with the `noetic-devel` branches for `rviz` and `interactive_markers`, and this branch for `fuse_viz`.

If bumping `rviz` (and `interactive_markers`) is not an option for you atm, that's fine. I'd simply have this in my fork. I don't know if it's possible add some pre-processor `if-else` blocks depending on the `rviz` version, but that could be another option.

### DIFF
--- a/fuse_viz/src/pose_2d_stamped_visual.cpp
+++ b/fuse_viz/src/pose_2d_stamped_visual.cpp
@@ -113,13 +113,7 @@ void Pose2DStampedVisual::setSphereColor(const float r, const float g, const flo
 
 void Pose2DStampedVisual::setAxesAlpha(const float alpha)
 {
-  static const auto& default_x_color_ = rviz::Axes::getDefaultXColor();
-  static const auto& default_y_color_ = rviz::Axes::getDefaultYColor();
-  static const auto& default_z_color_ = rviz::Axes::getDefaultZColor();
-
-  axes_->setXColor(Ogre::ColourValue{ default_x_color_.r, default_x_color_.g, default_x_color_.b, alpha });  // NOLINT
-  axes_->setYColor(Ogre::ColourValue{ default_y_color_.r, default_y_color_.g, default_y_color_.b, alpha });  // NOLINT
-  axes_->setZColor(Ogre::ColourValue{ default_z_color_.r, default_z_color_.g, default_z_color_.b, alpha });  // NOLINT
+  axes_->updateAlpha(alpha);
 }
 
 void Pose2DStampedVisual::setScale(const Ogre::Vector3& scale)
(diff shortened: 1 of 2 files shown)


svwilliams:
Build is currently failing with:
```
fuse_viz/src/pose_2d_stamped_visual.cpp:116:10: error: ‘using element_type = class rviz::Axes {aka class rviz::Axes}’ has no member named ‘updateAlpha’
```
I'm guessing this is that version mismatch you warned about. We recently updated several upstream packages. I'll look into exactly what version of rviz we are running. But at a minimum, I'll spin off a melodic-devel branch to maintain compatibility.

efernandez:
@svwilliams Yes, that error is because the _old_ `rviz` version doesn't have `updateAlpha`, and the _new_ one doesn't have the `static` methods.

FYI I couldn't easily find the exact `rviz` version that broke the API. It's definitely in `noetic-devel` 1.14.4 (last version released), and it's not in `melodic-devel`

ayrton04 (approved):

efernandez:
@svwilliams Any update on this? Just curious about status and whether there's a way I can help. I don't have any time pressure to have this in, since I'm already using it from our fork.

svwilliams:
Sorry for the delay. We are working on testing this with noetic. Hopefully I can get this merged in the near future.

efernandez:
@ayrton04 Is this PR still required after https://github.com/locusrobotics/fuse/pull/230/commits/e71bc04365a1be6469ce5aedb192fc01dd1251e7, from https://github.com/locusrobotics/fuse/pull/230 ?

I just tried it locally and it seems that everything builds fine now.

If that's the case and expectation, this can be closed.

The approach in this PR is cleaner and simpler, so I still think it'd be great to switch to this, but it isn't backwards compatible, unfortunately.

ayrton04:
I'd rather not have to split off a `melodic-devel` if we don't have to. If it was something in the core that was related to Ceres, I could see the need, but if we can support both Melodic and Noetic with the same code base, IMO that's preferable. Maybe @svwilliams disagrees, though.

svwilliams:
If we don't have to split melodic and noetic, then I'd prefer not having two active branches to maintain. Melodic still has another two years until EOL.

efernandez:
Feel free to close this PR

It'd be great if it could be hidden and postponed in 2 years from now :wink:

PR #222, opened 2021-02-05
Substract minimum twist covariance from twist covariance

efernandez:
This might sound a bit awkward at first, but I believe there's a use case for it, that I explain in a bit below.

### First, what's the issue this is trying to solve?

If the twist covariance already had a minimum twist covariance added to it in order to prevent ill-conditioned covariance matrices, we need a way to substract that minimum twist covariance from it before we compute the pose relative covariance.

Otherwise, we cannot get the original pose relative covariance because the minimum twist covariance term is
multiplied by the time delta, which could actually make the resulting pose relative covariance ill-conditioned or very small, i.e.
overconfident.

### How can we fix this?

In practice, this means that we can use the same minimum twist covariance that was added to the _raw_ twist covariance to obtain the _raw_ twist covariance and then compute the pose relative covariance and, if needed, add a minimum pose relative covariance to the result, which could be the same as the minimum twist covariance or different.

### Use case

If we don't care too much about having a pose relative covariance that matches exactly what it should be for the _raw_ twist covariance, it's true that with a minimum pose relative covariance we should be fine to guarantee the pose relative covariance isn't ill-conditioned.

However, we might not be able to remove the minimum twist covariance because other consumers might not have protection for ill-conditioned covariance matrices. For context, https://github.com/clearpathrobotics/ros_controllers/blob/38e474f3d05e42de1500f8486eb180803613d79d/diff_drive_controller/src/odometry.cpp#L203 is an example of an odometry producer that does this, and its output could be consumed by multiple consumers, not all able to handle the twist covariance if it were zero or very close to it.

### DIFF
--- a/fuse_models/include/fuse_models/common/sensor_proc.h
+++ b/fuse_models/include/fuse_models/common/sensor_proc.h
@@ -725,7 +725,7 @@ inline bool processDifferentialPoseWithCovariance(
  * @param[in] twist - The second (and temporally later) TwistWithCovarianceStamped message
  * @param[in] minimum_pose_relative_covariance - The minimum pose relative covariance that is always added to the
  *                                               resulting pose relative covariance
- * @param[in] minimum_twist_covariance - The minimum twist covariance that was added to the twist covariance and must be
+ * @param[in] twist_covariance_offset - The twist covariance offset that was added to the twist covariance and must be
  *                                       substracted from it before computing the pose relative covariance from it
  * @param[in] loss - The loss function for the 2D pose constraint generated
  * @param[in] validate - Whether to validate the measurements or not. If the validation fails no constraint is added
@@ -739,7 +739,7 @@ inline bool processDifferentialPoseWithTwistCovariance(
   const geometry_msgs::PoseWithCovarianceStamped& pose2,
   const geometry_msgs::TwistWithCovarianceStamped& twist,
   const fuse_core::Matrix3d& minimum_pose_relative_covariance,
-  const fuse_core::Matrix3d& minimum_twist_covariance,
+  const fuse_core::Matrix3d& twist_covariance_offset,
   const fuse_core::Loss::SharedPtr& loss,
   const std::vector<size_t>& position_indices,
   const std::vector<size_t>& orientation_indices,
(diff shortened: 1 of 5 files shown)


fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -819,6 +822,9 @@ inline bool processDifferentialPoseWithTwistCovariance(
   //
   // In some cases the twist covariance T12 is very small and it could yield to an ill-conditioned C12 covariance. For
   // that reason a minimum covariance is added to [2].
+  //
+  // It is also common that for the same reason, the twist covariance T12 has already been added a minimum covariacne to
+  // it, so we have to remove it before using it.
```
svwilliams:
```suggestion
  //
  // It is also common that for the same reason, the twist covariance T12 already has a minimum covariance added to
  // it by the publisher, so we have to remove it before using it.
```

fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -737,6 +739,7 @@ inline bool processDifferentialPoseWithTwistCovariance(
   const geometry_msgs::PoseWithCovarianceStamped& pose2,
   const geometry_msgs::TwistWithCovarianceStamped& twist,
   const fuse_core::Matrix3d& minimum_pose_relative_covariance,
+  const fuse_core::Matrix3d& minimum_twist_covariance,
```
svwilliams: My only concern here is similarity in name to "minimum_pose_relative_covariance". The pose covariance is added to the output, but the twist covariance is subtracted. For people like me that don't read the manual thoroughly, can we change the name to .... "twist_covariance_offset" ... or anything else really? Just a name that is not parallel to "minimum_pose_relative_covariance", to emphasize the difference in functionality.

svwilliams (approved):

fuse_models/include/fuse_models/common/sensor_proc.h
efernandez: I agree. I admit I also had the feeling it has a name very similar to `minimum_pose_relative_covariance` and could lead to confusion. I'm going to rename it to `twist_covariance_offset_diagonal`; I can't think of a better name and I'm just prefixing it with `_diagonal` for consistency with other _similar_ parameters. Although, maybe `twist_covariance_diagonal_offset` sounds better. :thinking:

efernandez: Done.

Finally using `twist_covariance_offset_diagonal` for the parameter and `twist_covariance_offset` for the variable/attribute.

I also added the word `offset` to the comment above. The one you suggested a change for.

ayrton04 (approved):

Issue #223, opened 2021-02-06
Cannot create constraints for IMU Twist messages with less than 2ms delay


meghdeepj:
Board: Nvidia Jetson Nano (Jetpack 4.4)
OS: Ubuntu 18.04
ROS: Melodic

Fuse sounds very promising and I am eager to use it for my AMR navigation stack. I am using fuse to combine data from Wheel odometry and IMU readings. IMU data is coming from Intel D435i via Madgwick Filter.
My launch file and `smoother.yaml` file is as follows:
 
 ```
<node name="smoother" pkg="fuse_optimizers" type="fixed_lag_smoother_node" output="screen">
        <rosparam command="load" file="$(find loc_mod)/params/fuse_imu.yaml" subst_value="true"/>
```
```
#Fixed-lag smoother configuration
optimization_frequency: 10  # The best-effort frequency (in Hz) of the optimization loop. This means that the optimizer will run every 0.1s and trigger all of the publishers when it is done
lag_duration: 0.5  # The size of the fixed-lag window, in seconds. The larger the window, the more variables will be inside the optimizer at any given time, which affects CPU usage.

# Define the set of motion model plugins to load
# Currently fuse only includes a 2D motion model. This is roughly equivalent to the robot_localization state, including the linear and angular position, the linear and angular velocity, and the linear acceleration. A turtlebot is a 2D robot, so this is fine.
# Multiple motion models can be loaded, and each sensor defines which set of motion models are used. We only have one motion model currently, and all sensors should use it.
motion_models:
  unicycle_motion_model:
    type: fuse_models::Unicycle2D

# Next define the set of sensor plugins to load
# You naturally want an IMU sensor and an Odometry sensor
# Additionally, you may want an "ignition" sensor. Ignition sensors tell the optimizer when to start running and what the initial state should be. fuse ships with one ignition sensor that initializes the Unicycle2D state variables.
sensor_models:
  ignition_sensor:
    type: fuse_models::Unicycle2DIgnition
    motion_models: [unicycle_motion_model]
    ignition: true
  odometry_sensor:
    type: fuse_models::Odometry2D
    motion_models: [unicycle_motion_model]
  imu_sensor:
    type: fuse_models::Imu2D
    motion_models: [unicycle_motion_model]

# And finally, define the set of publishers. fuse_models includes a robot_localization like publisher that publishes the latest state to the odom and tf topics.
publishers:
  odom_publisher:
    type: fuse_models::Odometry2DPublisher

[...]

PR #224, opened 2021-02-07
[Issue #223] Add an optional tf_timeout parameter to the sensor models

svwilliams:
In Issue #223 the sensor model failed to create a constraint due to the lack of an available tf transform. This PR adds an optional `tf_timeout` to all relevant sensor models. If the timeout is non-zero, the tf `lookupTransform()` signature with a timeout parameter will be used instead. This allows the user to optionally add a timeout duration, allowing the sensor processing code to wait for a transform is one is not available immediately. Note that this will potentially delay the sensor processing, and may cause dropped messages if the tf frame is not available in a timely fashion.

### DIFF
--- a/fuse_models/include/fuse_models/common/sensor_proc.h
+++ b/fuse_models/include/fuse_models/common/sensor_proc.h
@@ -218,14 +218,27 @@ inline void validatePartialMeasurement(
  * @param[in] tf_buffer - The transform buffer with which we will lookup the required transform
  * @param[in] input - The message to transform. Source frame and stamp are dictated by its header.
  * @param[in,out] output - The transformed message. Target frame is dictated by its header.
+ * @param [in] timeout - Optional. The maximum time to wait for a transform to become available.
  * @return true if the transform succeeded, false otherwise
  */
 template <typename T>
-bool transformMessage(const tf2_ros::Buffer& tf_buffer, const T& input, T& output)
+bool transformMessage(
+  const tf2_ros::Buffer& tf_buffer,
+  const T& input,
+  T& output,
+  const ros::Duration& tf_timeout = ros::Duration(0, 0))
 {
   try
   {
-    const auto trans = tf_buffer.lookupTransform(output.header.frame_id, input.header.frame_id, input.header.stamp);
+    auto trans = geometry_msgs::TransformStamped();
+    if (tf_timeout.isZero())
+    {
+      trans = tf_buffer.lookupTransform(output.header.frame_id, input.header.frame_id, input.header.stamp);
+    }
+    else
+    {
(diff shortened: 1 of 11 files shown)


svwilliams:
@ayrton04 Do you remember any reason why we did not allow the sensor processing to wait for a tf transform to begin with?

ayrton04:
Almost certainly because it had the potential to mess with the output rate.

fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -218,14 +218,27 @@ inline void validatePartialMeasurement(
...
+  T& output,
+  const ros::Duration& tf_timeout = ros::Duration(0, 0))
 {
   try
   {
-    const auto trans = tf_buffer.lookupTransform(output.header.frame_id, input.header.frame_id, input.header.stamp);
+    auto trans = geometry_msgs::TransformStamped();
+    if (tf_timeout.isZero())
+    {
+      trans = tf_buffer.lookupTransform(output.header.frame_id, input.header.frame_id, input.header.stamp);
```
ayrton04, line 236: I'm too lazy to look, but does this call just end up calling the other overload with a duration of 0 anyway? What happens if you always just pass `tf_timeout`, even if it's 0?

fuse_models/include/fuse_models/common/sensor_proc.h
```diff
@@ -1094,9 +1110,11 @@ inline bool processAccelWithCovariance(
   {
     transformed_message.header.frame_id = target_frame;
 
-    if (!transformMessage(tf_buffer, acceleration, transformed_message))
+    if (!transformMessage(tf_buffer, acceleration, transformed_message, tf_timeout))
     {
-      ROS_ERROR_STREAM("Cannot create constraint from pose message with stamp " << acceleration.header.stamp);
+      ROS_ERROR_STREAM_THROTTLE(
+        10.0,
+        "Cannot create constraint from pose message with stamp " << acceleration.header.stamp);
```
ayrton04: *acceleration message

ayrton04 (approved):

fuse_models/include/fuse_models/common/sensor_proc.h
svwilliams: It does not. The `lookupTransform()` method with the timeout calls the `canTransform()` version with a timeout internally, before calling the base class `lookupTransform()` method without the timeout. 
https://github.com/ros/geometry2/blob/noetic-devel/tf2_ros/src/buffer.cpp#L119

Whereas the non-timeout version of `lookupTransform()` _is_ the base class version.

There seemed like enough unneeded operations in the `canTransform()` implementation to make this additional bit of ugliness worthwhile.

matthewbwhitaker (approved):

PR #230, opened 2021-07-11
Fixing viz build issues in noetic

ayrton04:
This change *might* be compatible with melodic and noetic. I'd have to see if the melodic version of Ogre is compatible with this change.


svwilliams (approved):

efernandez (commented):
LGTM

PR #233, opened 2021-07-21
Add unstamped 3D point variable

jakemclaughlin6:

PR #239, opened 2021-07-23
Port the unstamped 3D point variable into devel

svwilliams:
Port the unstamped 3D point variable from kinetic-devel into devel
* Add unstamped 3D landmark variable
* Add landmark test and new uuid generator
Co-authored-by: Stephen Williams <swilliams@locusrobotics.com>


svwilliams:
The PR was previously reviewed here: #233

PR #243, opened 2021-08-11
[RST-4390] Allow variables to be held constant during optimization

svwilliams:
Ceres Solver allows you to mark variables as constant. Ceres will not alter the values of those variables during the optimization. In fact, it actually removes those variables from the linear optimization problem when solving for the best variable values. Add support for constant variables to fuse as follows:
* Provide an accessor method in the Variable base class for checking if a variable is held constant. Default to return false.
* Update the graph implementation to check the new variable property during insert and add them to the set of constant variables.
* Update the marginalization code to check for constant variables and set their jacobians to zero. This is the rough equivalent of removing them from the linear problem entirely.
* Add a unit test to verify the marginalization code works -- verify the remaining variables optimize to the same value before and after a constant variable is marginalized out.
* Add a "fixed landmark" variable that uses the new "hold constant" feature

### DIFF
--- a/fuse_constraints/src/marginalize_variables.cpp
+++ b/fuse_constraints/src/marginalize_variables.cpp
@@ -180,6 +180,12 @@ fuse_core::Transaction marginalizeVariables(
   const fuse_core::Graph& graph,
   const fuse_constraints::UuidOrdering& elimination_order)
 {
+  // TODO(swilliams) The method used to marginalize variables assumes that all variables are fully constrained.
+  //                 However, with the introduction of "variables held constant", it is possible to have a well-behaved
+  //                 system that is not fully-constrained. Ceres handles this issue by removing constant variables from
+  //                 the problem before the linearization and solve steps. A similar approach should be implemented
+  //                 here, but that will require a major refactor.
+
   assert(std::all_of(marginalized_variables.begin(),
                      marginalized_variables.end(),
                      [&elimination_order, &marginalized_variables](const fuse_core::UUID& variable_uuid)
@@ -321,19 +327,23 @@ LinearTerm linearize(
                              "during the jacobian computation.");
   }
 
-  // Update the jacobians with the local parameterizations.
+  // Update the jacobians with the local parameterizations. And clear the jacobians is the variable is held constant.
   for (size_t index = 0ul; index < variable_count; ++index)
   {
     const auto& variable_uuid = variable_uuids[index];
     const auto& variable = graph.getVariable(variable_uuid);
     auto local_parameterization = variable.localParameterization();
+    auto& jacobian = result.A[index];
     if (local_parameterization)
     {
-      auto& jacobian = result.A[index];
       fuse_core::MatrixXd J(local_parameterization->GlobalSize(), local_parameterization->LocalSize());
       local_parameterization->ComputeJacobian(variable_values[index], J.data());
       delete local_parameterization;
       jacobian *= J;
+    }
+    if (variable.holdConstant())
+    {
+      jacobian.setZero();
     }
   }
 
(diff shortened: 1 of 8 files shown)


ayrton04 (approved):

fuse_constraints/src/marginalize_variables.cpp
```diff
@@ -321,20 +327,24 @@ LinearTerm linearize(
                              "during the jacobian computation.");
   }
 
-  // Update the jacobians with the local parameterizations.
+  // Update the jacobians with the local parameterizations. And clear the jacobians is the variable is held constant.
```
needphpsql: This sentence is hard to grasp the meaning. Consider changing to another way?

svwilliams:
```
  // Update the jacobians with the local parameterizations. This potentially changes the size of the Jacobian block.
  // The classic example is a quaternion parameter, which has 4 components but only 3 degrees of freedom. The Jacobian
  // will be transformed from 4 columns to 3 columns after the local parameterization is applied.
  // We also check for variables that have been marked as constants. Since these variables cannot change value, their
  // derivatives/Jacobians should be zero.
```
Is that better?

needphpsql: Very clear.

fuse_constraints/src/marginalize_variables.cpp
```diff
@@ -321,20 +327,28 @@ LinearTerm linearize(
...
-      auto& jacobian = result.A[index];
       fuse_core::MatrixXd J(local_parameterization->GlobalSize(), local_parameterization->LocalSize());
       local_parameterization->ComputeJacobian(variable_values[index], J.data());
       delete local_parameterization;
       jacobian *= J;
     }
+    if (variable.holdConstant())
+    {
+      jacobian.setZero();
+    }
```
needphpsql: Shouldn't this be checked before "if (local_parameterization)"? If it is hold constant, any need to compute the jacobian for this variable?

svwilliams: Yes, it is wasteful. And yes, that does bother me.

The problem is the local parameterization potentially changes the size the Jacobian matrix. By applying the local parameterization even for constant variables, we ensure the block of zeros is the correct size.

But since you bring it up, we could do something like:
```
if (variable.holdConstant())
{
  if (local_parameterization)
  {
    jacobian.resize(jacobian.rows(), local_parameterization->LocalSize());
  }
  jacobian.setZero();
}
else if (local_parameterization)
{
  ...
}
```
Let me see if I can make that work without too much trouble.

svwilliams: There's even a fancy `Eigen::NoChange` option so I don't need to look up the row count.
```
jacobian.resize(Eigen::NoChange, local_parameterization->LocalSize());
```

needphpsql (approved):

fuse_variables/src/point_3d_fixed_landmark.cpp
```diff
@@ -0,0 +1,73 @@
...
+         << "  landmark id: " << id() << "\n"
+         << "  data:\n"
+         << "  - x: " << x() << "\n"
+         << "  - y: " << y() << "\n"
+         << "  - z: " << z() << "\n";
+}
+
+bool Point3DFixedLandmark::holdConstant() const
+{
+  return true;
```
efernandez, line 67: @svwilliams Just curious, I wonder if this is the only difference wrt `Point3DLandmark`, why `Point3DFixedLandmark` wasn't simply implemented as:
```c++
class Point3DFixedLandmark : public Point3DLandmark
{
public:
  FUSE_VARIABLE_DEFINITIONS(Point3DFixedLandmark);

  bool holdConstant() const
  {
    return true;
  }
};
```

and the corresponding code for the serialization and pluginlib in the cpp file. :thinking:

svwilliams: Honestly, I think just an accident of history.

svwilliams: https://github.com/locusrobotics/fuse/pull/259

PR #253, opened 2021-10-16
Add files for the simple tutorial

svwilliams:
Add files for the simple tutorial that is already documented here: http://docs.ros.org/en/noetic/api/fuse_doc/html/getting_started.html

This prevents people from having to download code from an AWS S3 bucket.


fuse_tutorials/config/range_sensor_tutorial.rviz
```diff
@@ -0,0 +1,222 @@
+Panels:
+- Class: rviz/Displays
+  Help Height: 70
+  Name: Displays
+  Property Tree Widget:
+    Expanded:
+    - /True Robot Pose1/Shape1
+    - /Estimated Robot Pose1/Covariance1/Position1
+    Splitter Ratio: 0.6058823466300964
```
needphpsql: Curious, how is this .rviz file created? Manual or export from GUI? This looks like not very readable and memorized.

needphpsql (approved):

fuse_tutorials/config/range_sensor_tutorial.rviz
svwilliams: Configured the display using the rviz GUI, then saved the config file. We also have a script to strip out _some_ of the inconsequential data. It's not great, but I don't know of a good way to generate a minimal rviz config file.

anhillman (approved):

PR #259, opened 2021-11-21
Clean up the landmark implementations

svwilliams:
From https://github.com/locusrobotics/fuse/pull/243#discussion_r753010333

The "fixed" landmarks are identical to the normal landmarks, expect for the implementation of `holdConstant()`. Make the "fixed" landmarks inherit from the normal landmarks to reduce code duplication. This also means that any constraint that accepts a normal landmark variable will also accept a fixed landmark variable.


efernandez (approved):
LGTM

needphpsql (approved):

ayrton04 (approved):

Issue #248, opened 2021-09-20
Attempting to remove a variable ... that is used by existing constraints ...


jakemclaughlin6:
When running a fixed lag smoother occasionally the following error will occur at random times, despite no client side code removing any variables, so I am assuming this is from the marginalization. This error is uncatchable in the sensor model code. Is it possible to have this be changed to be a warning message so that it does not crash the program, or does this error cause a total failure of the problem being optimized? Is it possible to have a setting where if this is the case, then said constraints can be removed?

svwilliams:
This sounds like a bug in the marginalization code. Is there any way you can shared your fuse configuration and maybe a bagfile that I can replay to reproduce the problem? If I can reproduce the error, I'll get the bug fixed as soon as possible.

I have it configured to throw because this error simply should not happen. But I will review the code to determine if this error is truly catastrophic, or if I can reduce that exception to a warning/error log.

jakemclaughlin6:
I'm not able to share the fuse config as I am running custom sensor models in a private repository right now. However the error seems to occur more frequently when the lag duration is small (~5 seconds) and the optimization period is low (~0.5 seconds). I'm currently adding new position/orientation variables at a rate of 5hz

svwilliams:
That makes finding the issue harder. Is there anything else you can tell me about your setup:
* How long the system runs between errors
* Number of active sensors
* What variables are involved in the sensor constraints? e.g. A fuse_models IMU sensor only uses the current angular velocity. A custom odometry constraint might use the current position and orientation and an older position and orientation. And a visual odometry system might use the current position and orientation as well as some unstamped visual landmarks.
* Are you using the fuse_models motion model, or a custom motion model? If custom, can you tell me anything about the involved variables?

If I know more details, I could create some trivial sensors that roughly mimic the connectivity of your configuration. I can then run that setup with some simulated data and see if I can reproduce your issue. At a minimum, I'd be able to share that configuration with you, and we could iterate on that.

jakemclaughlin6:
1. Completely different each time, sometimes it'll occur after about 5 seconds, sometimes it may not occur at all
2. Currently running with just one visual odom sensor model
[...]

PR #261, opened 2021-12-06
Issue258 add fixed size optimizer

jakemclaughlin6:


jakemclaughlin6:
@svwilliams Can this be reviewed soon, not sure if you would like some tests written for it since the other optimizers dont have any either, but the changes I made to timestamp tracker have been tested and I've also been using this optimizer for my own code and it seems to work well.

svwilliams:
It may be a couple of weeks before I can take a look. But I'll do my best to get to it sooner.

jakemclaughlin6:
This PR implements a fixed size window optimizer where the window is determined by the number of "states" in it. The number of states are determined by the total # of unique timestamps exist in the graph.

Another addition from this PR is the fix to the following issue: https://github.com/locusrobotics/fuse/issues/248. It would seem that what was occurring in a fringe edge case was a sensor model was sending a transaction and adding constraints to variables that were already chosen to be marginalized out in the previous cycle, the fix involved just checking for these added constraints and removing them. I am curious if this fix is suitable or if there should be another method to actually re marginalize the constraints that have been added @svwilliams.

Lastly, a duplicate of the reset service was implemented but as a message callback, this is for the case where a system would like to call a reset however only knows when it would like to reset in its own message callbacks. Calling the service inside the sensors callback created a deadlock, making a message callback version allows for the sensor model to ask for a reset and entrust that all onStart() and onStop() functions perform the required logic to reset the system

fuse_optimizers/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(fuse_optimizers)
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp")
```
svwilliams: You should never set the CMAKE_CXX_FLAGS directly in a catkin package. If you are using catkin_make, then a single makefile is created for the entire workspace. Modifying CXX_FLAGS like this pollutes the entire workspace. It's better to add the required compile flags on a per-target basis.

```suggestion
```

And you should `find_package(OpenMP)` before assuming OpenMP is available.

And then add something like this to the library config
```
target_link_libraries(${PROJECT_NAME}
  ${OpenMP_CXX_LIBRARIES}
)
target_compile_options(${PROJECT_NAME}
  PRIVATE ${OpenMP_CXX_FLAGS}
)
```

fuse_optimizers/include/fuse_optimizers/fixed_size_smoother.h
```diff
@@ -0,0 +1,197 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2019, Locus Robotics
```
svwilliams:
```suggestion
 *  Copyright (c) 2022, Locus Robotics
```

fuse_optimizers/include/fuse_optimizers/fixed_size_smoother.h
```diff
@@ -0,0 +1,197 @@
...
+#include <ros/time.h>
+
+#include <mutex>
+#include <string>
+#include <vector>
+
+namespace fuse_optimizers
+{
+/**
+ * @brief A fixed-Size smoother implementation that marginalizes out variables that are older than a defined Size time
```
svwilliams:
```suggestion
 * @brief A fixed-size smoother implementation that marginalizes out variables that are older than a defined size
```
Or maybe?
```suggestion
 * @brief A fixed-size smoother implementation that marginalizes out variables that are older than a defined buffer length
```

Also, it looks like find&replace got away from you. Review all of the comments and make sure `Size` is properly capitalized and makes sense in the sentence.

fuse_optimizers/include/fuse_optimizers/fixed_size_smoother.h
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * This implementation assumes that all added variable types are either derived from the fuse_variables::Stamped class,
+ * or are directly connected to at least one fuse_variables::Stamped variable via a constraint. The current time of
+ * the fixed-Size smoother is determined by the newest stamp of all added fuse_variables::Stamped variables.
+ *
+ * During optimization:
+ *  (1) new variables and constraints are added to the graph
+ *  (2) the augmented graph is optimized and the variable values are updated
+ *  (3) all motion models, sensors, and publishers are notified of the updated graph
+ *  (4) all variables older than "current time - Size duration" are marginalized out.
```
svwilliams: I'm guessing this statement should be revised

fuse_optimizers/include/fuse_optimizers/fixed_size_smoother_params.h
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * @brief Defines the set of parameters required by the fuse_optimizers::FixedSizeSmoother class
+ */
+struct FixedSizeSmootherParams : public WindowedOptimizerParams
+{
+public:
+  SMART_PTR_DEFINITIONS(FixedSizeSmootherParams);
+
+  /**
+   * @brief The duration of the smoothing window in seconds
```
svwilliams: This statement needs to be revised

fuse_optimizers/include/fuse_optimizers/fixed_size_smoother.h
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * Optimization is performed at a fixed frequency, controlled by the \p optimization_frequency parameter. Received
+ * sensor transactions are queued while the optimization is processing, then applied to the graph at the start of the
+ * next optimization cycle. If the previous optimization is not yet complete when the optimization period elapses,
+ * then a warning will be logged but a new optimization will *not* be started. The previous optimization will run to
+ * completion, and the next optimization will not begin until the next scheduled optimization period.
+ *
+ * Parameters:
+ *  - Size_duration (float, default: 5.0) The duration of the smoothing window in seconds
```
svwilliams:
```suggestion
 *  - num_states (int, default: 10) The maximum number of states in the smoothing window
```
Or something similar

fuse_optimizers/include/fuse_optimizers/fixed_size_smoother_params.h
```diff
@@ -0,0 +1,80 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2019, Locus Robotics
```
svwilliams:
```suggestion
 *  Copyright (c) 2022, Locus Robotics
```

fuse_optimizers/include/fuse_optimizers/fixed_size_smoother_params.h
```diff
@@ -0,0 +1,80 @@
...
+#include <fuse_core/parameter.h>
+#include <fuse_optimizers/windowed_optimizer_params.h>
+#include <ros/duration.h>
+#include <ros/node_handle.h>
+
+#include <ceres/solver.h>
+
+#include <algorithm>
+#include <string>
+#include <vector>
```
svwilliams: Most of these includes are not required
```suggestion
#include <fuse_core/parameter.h>
#include <fuse_optimizers/windowed_optimizer_params.h>
#include <ros/node_handle.h>
```

fuse_optimizers/src/fixed_size_smoother_node.cpp
```diff
@@ -0,0 +1,51 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2019, Locus Robotics
```
svwilliams:
```suggestion
 *  Copyright (c) 2022, Locus Robotics
```

fuse_optimizers/src/variable_stamp_index.cpp
```diff
@@ -47,19 +47,29 @@ namespace fuse_optimizers
...
-  auto compare_stamps = [](const StampedMap::value_type& lhs, const StampedMap::value_type& rhs)
-  {
-    return lhs.second < rhs.second;
-  };
-  auto iter = std::max_element(stamped_index_.begin(), stamped_index_.end(), compare_stamps);
-  if (iter != stamped_index_.end())
-  {
-    return iter->second;
+  if(!unique_stamps_.empty()){
+    return *unique_stamps_.end();
```
svwilliams: Doesn't this throw or segfault or something? The `end()` iterator is not dereferenceable.
```suggestion
    return *unique_stamps_.rbegin();
```

fuse_optimizers/src/variable_stamp_index.cpp
```diff
@@ -47,19 +47,29 @@ namespace fuse_optimizers
...
   }
+  return ros::Time(0, 0);
+} 
+
+ros::Time VariableStampIndex::ithStamp(size_t idx) const{
+  if(idx < unique_stamps_.size()){
+    std::set<ros::Time>::iterator it = unique_stamps_.begin();
+    for(size_t i = 0; i < idx; i++){
+      it++;
+    }
```
svwilliams: There's a function for this exact purpose
```suggestion
    std::advance(it, idx);
```

fuse_optimizers/src/variable_stamp_index.cpp
```diff
@@ -47,19 +47,29 @@ namespace fuse_optimizers
...
+ros::Time VariableStampIndex::firstStamp() const
+{
+  if(!unique_stamps_.empty()){
+    return *unique_stamps_.begin();
   }
+  return ros::Time(0, 0);
+} 
+
+ros::Time VariableStampIndex::ithStamp(size_t idx) const{
+  if(idx < unique_stamps_.size()){
```
svwilliams: Adhere to the existing style
```suggestion
ros::Time VariableStampIndex::ithStamp(size_t idx) const
{
  if (idx < unique_stamps_.size())
  {
```

fuse_optimizers/src/variable_stamp_index.cpp
```diff
@@ -47,19 +47,29 @@ namespace fuse_optimizers
...
+
+ros::Time VariableStampIndex::firstStamp() const
+{
+  if(!unique_stamps_.empty()){
+    return *unique_stamps_.begin();
   }
+  return ros::Time(0, 0);
+} 
+
+ros::Time VariableStampIndex::ithStamp(size_t idx) const{
```
svwilliams: What are your thoughts on the proper action if `idx` exceeds the size of the container? Stdlib avoids the issue and just claims it is undefined behavior. We could do that as well and blindly call `std::advance()` without checking sizes first. Or we could throw an exception to provide some debugging information. Or we can return Time(0) as some sort of magic code for "does not exist".

fuse_optimizers/include/fuse_optimizers/variable_stamp_index.h
```diff
@@ -86,13 +91,24 @@ class VariableStampIndex
...
 
+  /**
+   * @brief Returns the oldest/first stamp 
+   */
+  ros::Time firstStamp() const;
+
+  /**
+   * @brief Returns the ith timestamp
+   */
+  ros::Time ithStamp(size_t idx) const;
```
svwilliams: If you need access to the first, last, and i-th element of the unique stamps set, would it make more sense just to provide read-only access to the container itself?
```
const std::set<ros::Time>& stamps() const { return unique_stamps_; }
```

fuse_optimizers/src/variable_stamp_index.cpp
```diff
@@ -99,6 +109,10 @@ void VariableStampIndex::applyAddedVariables(const fuse_core::Transaction& trans
     if (stamped_variable)
     {
       stamped_index_[variable.uuid()] = stamped_variable->stamp();
+      // add to unique stamps
+      if(unique_stamps_.find(stamped_variable->stamp()) == unique_stamps_.end()){
+        unique_stamps_.insert(stamped_variable->stamp());
+      }
```
svwilliams: I don't think there is a need to query of rht e stamp first. If it already exists, `insert()` will not insert it again.
```suggestion
      unique_stamps_.insert(stamped_variable->stamp());
```

fuse_optimizers/src/variable_stamp_index.cpp
```diff
@@ -120,6 +134,12 @@ void VariableStampIndex::applyRemovedVariables(const fuse_core::Transaction& tra
 {
   for (const auto& variable_uuid : transaction.removedVariables())
   {
+    // remove from unique stamps
+    auto stamp = stamped_index_[variable_uuid];
+    if(unique_stamps_.find(stamp) != unique_stamps_.end()){
+      unique_stamps_.erase(stamp);
+    }
```
svwilliams: This comment is important:
Multiple variables may be associated with the same timestamp. The motion model in the fuse_models package uses position, orientation, linear velocity, angular velocity, and linear acceleration variables all at the exact same timestamp. Just because you remove the linear acceleration variable doesn't mean that the other variables are also removed. It feels like you really need to keep track of the individual variables associated with each timestamp, and only remove a timestamp once all variables have been removed.
Some sort of structure like:
```
using StampToVariablesMap = std::map<ros::Time, std::unordered_set<fuse_core::UUID>>;
StampToVariablesMap unique_stamps_;
```

fuse_optimizers/src/windowed_optimizer.cpp
```diff
@@ -169,6 +175,25 @@ void WindowedOptimizer::optimizationLoop()
...
+            }
+          }
+        }
+      }
+      if(faulty_constraints.size() > 0){
+        ROS_WARN_STREAM("Removing invalid constraints.");
+        for(auto& faulty_constraint: faulty_constraints){
+          new_transaction->removeConstraint(faulty_constraint);
+        }
+      }
```
svwilliams: Interesting. There is indeed a timing problem. At the end of Loop1 we compute a set of variables to marginalize. But those variables are not removed until the graph is updated in Loop2. The sensor models have no way of knowing a variable was removed in Loop1 until after Loop2 has been processed and the updated graph is sent back to the sensor models. So a sensor model can very reasonably send new constraints involving variables that should have been removed.

I originally implemented it this way to avoid optimizing the graph twice. Once when new constraints are added, and a second time to remove the marginalized variables. In retrospect, marginalizing variables from the graph should not alter the optimized values of the remaining variables. So I could just remove the marginalized variables at the end of the loop without reoptimizing. But I wouldn't want to make that change without a lot of testing first.

For now, your fix should be valid. Thanks.

fuse_optimizers/src/windowed_optimizer.cpp
```diff
@@ -397,6 +422,40 @@ void WindowedOptimizer::processQueue(fuse_core::Transaction& transaction)
...
+    graph_->clear();
+    marginal_transaction_ = fuse_core::Transaction();
+  }
+  // Perform any required reset operations for derived classes
+  onReset();
+  // Tell all the plugins to start
+  startPlugins();
+  // Test for auto-start
+  autostart();
+  return;
```
svwilliams: This is identical to the code in the `resetServiceCallback()`. Can you pull this out into a `reset()` function, and call that function from both `resetServiceCallback()` and `resetMessageCallback()`?

fuse_optimizers/test/test_variable_stamp_index.cpp
```diff
@@ -282,6 +282,102 @@ TEST(VariableStampIndex, CurrentStamp)
...
+
+  for(int i = 0; i < N; i++){
+    auto x = StampedVariable::make_shared(ros::Time(i, 0));
+    auto transaction = fuse_core::Transaction();
+    transaction.addVariable(x);
+    index.addNewTransaction(transaction);
+  }
+
+  // Verify the number of unique stamps
+  EXPECT_EQ((size_t)N, index.numStates());
```
svwilliams: If you use unsigned types for `N` and `i`, you won't need to typecast anything.
```suggestion
  unsigned int N = 10u;

  for(auto i = 0u; i < N; i++)
  {
    auto x = StampedVariable::make_shared(ros::Time(i, 0));
    auto transaction = fuse_core::Transaction();
    transaction.addVariable(x);
    index.addNewTransaction(transaction);
  }

  // Verify the number of unique stamps
  EXPECT_EQ(N, index.numStates());
```

fuse_optimizers/test/test_variable_stamp_index.cpp
```diff
@@ -282,6 +282,102 @@ TEST(VariableStampIndex, CurrentStamp)
...
+    auto x = StampedVariable::make_shared(ros::Time(i, 0));
+    auto transaction = fuse_core::Transaction();
+    transaction.addVariable(x);
+    index.addNewTransaction(transaction);
+  }
+
+  // Verify the stamps of the states using ithStamp function
+  for(int i = 0; i < N; i++){
+    EXPECT_EQ(ros::Time(i, 0), index.ithStamp((size_t)i));  
+  }
```
svwilliams: Same here
```suggestion
  unsigned int N = 10u;
  for(auto i = 0u; i < N; i++)
  {
    auto x = StampedVariable::make_shared(ros::Time(i, 0));
    auto transaction = fuse_core::Transaction();
    transaction.addVariable(x);
    index.addNewTransaction(transaction);
  }

  // Verify the stamps of the states using ithStamp function
  for(auto i = 0u; i < N; i++)
  {
    EXPECT_EQ(ros::Time(i, 0), index.ithStamp(i));  
  }
```

svwilliams (changes requested):
My big concern is in the VariableStampIndex. It seems to me that timestamps can be removed while they are still in use by other variables.

Additionally, there is a bit of cleanup needed in CMakelists and numerous linter errors that need to be fixed to conform to ROS coding conventions:
```
/fuse_optimizers/include/fuse_optimizers/variable_stamp_index.h:195:  At least two spaces is best between code and comments  [whitespace/comments] [2]
/fuse_optimizers/include/fuse_optimizers/variable_stamp_index.h:195:  Add #include <set> for set<>  [build/include_what_you_use] [4]
/fuse_optimizers/include/fuse_optimizers/windowed_optimizer.h:188:  At least two spaces is best between code and comments  [whitespace/comments] [2]
/fuse_optimizers/src/fixed_size_smoother.cpp:89:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/fixed_size_smoother.cpp:89:  Missing space before ( in if(  [whitespace/parens] [5]
/fuse_optimizers/src/fixed_size_smoother.cpp:89:  Missing space before {  [whitespace/braces] [5]
/fuse_optimizers/src/fixed_size_smoother.cpp:89:  Using C-style cast.  Use static_cast<int>(...) instead  [readability/casting] [4]
/fuse_optimizers/src/variable_stamp_index.cpp:50:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/variable_stamp_index.cpp:50:  Missing space before ( in if(  [whitespace/parens] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:50:  Missing space before {  [whitespace/braces] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:58:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/variable_stamp_index.cpp:58:  Missing space before ( in if(  [whitespace/parens] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:58:  Missing space before {  [whitespace/braces] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:62:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
/fuse_optimizers/src/variable_stamp_index.cpp:64:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/variable_stamp_index.cpp:64:  Missing space before {  [whitespace/braces] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:65:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/variable_stamp_index.cpp:65:  Missing space before ( in if(  [whitespace/parens] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:65:  Missing space before {  [whitespace/braces] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:67:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/variable_stamp_index.cpp:67:  Missing space before ( in for(  [whitespace/parens] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:67:  Missing space before {  [whitespace/braces] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:113:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/variable_stamp_index.cpp:113:  Missing space before ( in if(  [whitespace/parens] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:113:  Missing space before {  [whitespace/braces] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:139:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/variable_stamp_index.cpp:139:  Missing space before ( in if(  [whitespace/parens] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:139:  Missing space before {  [whitespace/braces] [5]
/fuse_optimizers/src/variable_stamp_index.cpp:66:  Add #include <set> for set<>  [build/include_what_you_use] [4]
/fuse_optimizers/src/windowed_optimizer.cpp:110:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
/fuse_optimizers/src/windowed_optimizer.cpp:180:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/windowed_optimizer.cpp:180:  Missing space around colon in range-based for loop  [whitespace/forcolon] [2]
/fuse_optimizers/src/windowed_optimizer.cpp:180:  Missing space before ( in for(  [whitespace/parens] [5]
/fuse_optimizers/src/windowed_optimizer.cpp:180:  Missing space before {  [whitespace/braces] [5]
/fuse_optimizers/src/windowed_optimizer.cpp:181:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/windowed_optimizer.cpp:181:  Missing space around colon in range-based for loop  [whitespace/forcolon] [2]
/fuse_optimizers/src/windowed_optimizer.cpp:181:  Missing space before ( in for(  [whitespace/parens] [5]
/fuse_optimizers/src/windowed_optimizer.cpp:181:  Missing space before {  [whitespace/braces] [5]
/fuse_optimizers/src/windowed_optimizer.cpp:182:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/windowed_optimizer.cpp:183:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/windowed_optimizer.cpp:190:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/windowed_optimizer.cpp:190:  Missing space before ( in if(  [whitespace/parens] [5]
/fuse_optimizers/src/windowed_optimizer.cpp:190:  Missing space before {  [whitespace/braces] [5]
/fuse_optimizers/src/windowed_optimizer.cpp:192:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]
/fuse_optimizers/src/windowed_optimizer.cpp:192:  Missing space around colon in range-based for loop  [whitespace/forcolon] [2]
```

fuse_optimizers/src/variable_stamp_index.cpp
jakemclaughlin6: The previous behaviour for firstStamp would return Time(0) for the failure cases so I would stick with this for the rest. However I see how Time(0) makes sense for firstStamp and even currentStamp when the time index is empty. I think time(0) should be returned when empty, and an exception should be thrown if index is greater than the size.

fuse_optimizers/include/fuse_optimizers/variable_stamp_index.h
jakemclaughlin6: I was thinking of overriding the [] operator to perform the function of these 3 functions instead, this way it keeps the validity checking to the stamp index rather than to the client using it

PR #270, opened 2022-05-07
Compile in Ubuntu 22.04, 24.04 & C++17

lucasw:
Work-in-progress, there are a couple of warnings I didn't eliminate- one requires rviz changes (OgreVector3.h includes), another in eigen:

```
In file included from /usr/include/eigen3/Eigen/Core:337,
                 from /usr/include/eigen3/Eigen/Dense:1,
                 from /home/lucasw/catkin_ws/src/fuse/fuse_viz/include/fuse_viz/mapped_covariance_visual.h:41,
                 from /home/lucasw/catkin_ws/src/fuse/fuse_viz/src/mapped_covariance_visual.cpp:30:
/usr/include/eigen3/Eigen/src/Core/products/SelfadjointMatrixVector.h: In function ‘static void Eigen::internal::selfadjoint_product_impl<Lhs, LhsMode, false, Rhs, 0, true>::run(Dest&, const Lhs&, const Rhs&, const Scalar&) [with Dest = Eigen::Block<Eigen::Matrix<double, 1, 1, 0, 1, 1>, -1, 1, false>; Lhs = Eigen::Block<Eigen::Matrix<double, 2, 2>, -1, -1, false>; int LhsMode = 17; Rhs = Eigen::CwiseBinaryOp<Eigen::internal::scalar_product_op<double, double>, const Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<double>, const Eigen::Matrix<double, -1, 1, 0, 2, 1> >, const Eigen::Block<Eigen::Block<Eigen::Matrix<double, 2, 2>, 2, 1, true>, -1, 1, false> >]’:
/usr/include/eigen3/Eigen/src/Core/products/SelfadjointMatrixVector.h:229:7: error: ‘result’ may be used uninitialized [-Werror=maybe-uninitialized]
  227 |     internal::selfadjoint_matrix_vector_product<Scalar, Index, (internal::traits<ActualLhsTypeCleaned>::Flags&RowMajorBit) ? RowMajor : ColMajor,
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  228 |                                                 int(LhsUpLo), bool(LhsBlasTraits::NeedToConjugate), bool(RhsBlasTraits::NeedToConjugate)>::run
      |                                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  229 |       (
      |       ^
  230 |         lhs.rows(),                             // size
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  231 |         &lhs.coeffRef(0,0),  lhs.outerStride(), // lhs info
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  232 |         actualRhsPtr,                           // rhs info
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  233 |         actualDestPtr,                          // result info
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  234 |         actualAlpha                             // scale factor
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  235 |       );
      |       ~
/usr/include/eigen3/Eigen/src/Core/products/SelfadjointMatrixVector.h:41:6: note: by argument 3 of type ‘const double*’ to ‘static void Eigen::internal::selfadjoint_matrix_vector_product<Scalar, Index, StorageOrder, UpLo, ConjugateLhs, ConjugateRhs, Version>::run(Index, const Scalar*, Index, const Scalar*, Scalar*, Scalar) [with Scalar = double; Index = long int; int StorageOrder = 0; int UpLo = 1; bool ConjugateLhs = false; bool ConjugateRhs = false; int Version = 0]’ declared here
   41 | void selfadjoint_matrix_vector_product<Scalar,Index,StorageOrder,UpLo,ConjugateLhs,ConjugateRhs,Version>::run(
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

That looks like it's entirely in Eigen but I haven't looked closely.

Current workaround is to remove the fuse_viz `Werror`

### DIFF
--- a/fuse_optimizers/CMakeLists.txt
+++ b/fuse_optimizers/CMakeLists.txt
@@ -28,7 +28,7 @@ catkin_package(
 ###########
 ## Build ##
 ###########
-add_compile_options(-Wall -Werror)
+add_compile_options(-Wall)  # -Werror)
 
 ## fuse_optimizers library
 add_library(${PROJECT_NAME}
--- a/fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp
+++ b/fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp
@@ -343,7 +343,7 @@ Ogre::ColourValue RelativePose2DStampedConstraintVisual::computeLossErrorLineCol
   // Get the error line color as HSB:
   Ogre::ColourValue error_line_color(color.r, color.g, color.b);
   Ogre::Real hue, saturation, brightness;
-  error_line_color.getHSB(&hue, &saturation, &brightness);
+  error_line_color.getHSB(hue, saturation, brightness);
 
   // We should correct the color brightness if it is smaller than minimum brightness. Otherwise, we would get an
   // incorrect loss brightness.
--- a/fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp
+++ b/fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp
@@ -46,6 +46,7 @@
 #include <tf2/utils.h>
 #include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 
+#include <OgrePrerequisites.h>
 #include <OgreQuaternion.h>
 #include <OgreSceneManager.h>
 #include <OgreSceneNode.h>
@@ -343,7 +344,13 @@ Ogre::ColourValue RelativePose2DStampedConstraintVisual::computeLossErrorLineCol
   // Get the error line color as HSB:
   Ogre::ColourValue error_line_color(color.r, color.g, color.b);
   Ogre::Real hue, saturation, brightness;
+#if (OGRE_VERSION < ((1 << 16) | (11 << 8) | 0))
+  // 1.10 or earlier
+  error_line_color.getHSB(&hue, &saturation, &brightness);
+# else
+  // 1.11 or later
   error_line_color.getHSB(hue, saturation, brightness);
+#endif
 
   // We should correct the color brightness if it is smaller than minimum brightness. Otherwise, we would get an
   // incorrect loss brightness.
--- a/fuse_viz/CMakeLists.txt
+++ b/fuse_viz/CMakeLists.txt
@@ -33,7 +33,8 @@ add_definitions(-DQT_NO_KEYWORDS)
 ###########
 ## Build ##
 ###########
-add_compile_options(-Wall -Werror)
+# add_compile_options(-Wall -Werror)
+add_compile_options(-Wall)
 
 catkin_package(
   INCLUDE_DIRS include
--- a/fuse_viz/include/fuse_viz/conversions.h
+++ b/fuse_viz/include/fuse_viz/conversions.h
@@ -46,7 +46,7 @@
 
 #include <OgreColourValue.h>
 #include <OgreQuaternion.h>
-#include <OgreVector3.h>
+#include <Ogre.h>
 
 #include <boost/array.hpp>
 
--- a/fuse_viz/include/fuse_viz/mapped_covariance_visual.h
+++ b/fuse_viz/include/fuse_viz/mapped_covariance_visual.h
@@ -41,7 +41,7 @@
 #include <Eigen/Dense>
 
 #include <OgreColourValue.h>
-#include <OgreVector3.h>
+#include <Ogre.h>
 
 namespace Ogre
 {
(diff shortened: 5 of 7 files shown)


lucasw:
I'm trying to make this build in Ubuntu 22.10 but getting:

```
/home/lucasw/base_catkin_ws/src/other/fuse/fuse_graphs/include/fuse_graphs/hash_graph.h:426:21: error: ‘ceres::Problem::Options::local_parameterization_ownership’ is deprecated: Local Parameterizations are deprecated. Use Manifold and manifold_ownership instead. [-Werror=deprecated-declarations]
  426 |   archive & options.local_parameterization_ownership;
```

22.10 has ceres 2.1.0

svwilliams:
I swear I've written this before, but I'm not seeing it in this PR history. So either this will be a repeat of a message you have seen before, or I never clicked "send". So apologies either way.

Ceres 2.1 has dramatically changed the implementation of what was previously called LocalParameterization and is now called a Manifold. Making `fuse` support Ceres 2.1 will require a significant time investment. It is something that I intend to support, but likely only in the ROS2 branch. If migrating all of the `fuse` LocalParameterization code into the new Manifold classes is something you want to tackle, we can create a noetic-22.10-devel branch or similar to hold those updates. But it is not something I can implement or maintain at this time.

lucasw:
How about I peel out the commits that don't break ci, get those into another pr- probably just the hpp and maybe ogre.h includes- that'll reduce the diff noise with breaking changes here, make it easier for me to maintain a 22.04+ branch.

svwilliams:
Sounds good. I did push a commit a couple of months ago to get fuse compiling on Ubuntu Jammy. That included some of these changes already.

lucasw:
Right, https://github.com/locusrobotics/fuse/pull/326 has the https://github.com/locusrobotics/fuse/pull/270/commits/d045cc226a391a78bc62ee78065e8fed12c68403 and https://github.com/locusrobotics/fuse/pull/270/commits/98dee7fd6e8d6a2c20d5adf6a7f073a917078426 changes, which only leaves these from https://github.com/locusrobotics/fuse/pull/270/commits/b3da98f0c0146cc3eab5cdd1082bbcb5cd4014de:

```
-#include <OgreVector3.h>
+#include <Ogre.h>
```

fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp
```diff
@@ -343,7 +343,13 @@ Ogre::ColourValue RelativePose2DStampedConstraintVisual::computeLossErrorLineCol
   // Get the error line color as HSB:
   Ogre::ColourValue error_line_color(color.r, color.g, color.b);
   Ogre::Real hue, saturation, brightness;
+#if (OGRE_VERSION < ((1 << 16) | (11 << 8) | 0))
```
svwilliams, line 347: Please include the `OgrePrerequisites.h` header to support the `OGRE_VERSION` macro.

svwilliams (changes requested):

fuse_viz/src/relative_pose_2d_stamped_constraint_visual.cpp
lucasw: Done in https://github.com/locusrobotics/fuse/pull/270/commits/5b4dd949514cef2a89dbcd02003094ba186b5262

fuse_optimizers/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ catkin_package(
 ###########
 ## Build ##
 ###########
-add_compile_options(-Wall -Werror)
+add_compile_options(-Wall)  # -Werror)
```
lucasw, line 31: I haven't tried restoring Werror here since rebasing these commits on latest, maybe it's not needed now?

PR #271, opened 2022-06-10
Port to ROS2 Galactic

BrettRD:
This is my work so far, with cherry-picked commits from `devel` and `ros2`
I could use some help making decisions about Time and Parameters

I'm stuck in the weeds in the optimizers, I'm looking forward to getting the unit-tests running for them because I've made a few small mistakes.

The Parameters in particular need updating, ROS2 offers a parameter description available at CLI, and a callback to validate and update the parameters at runtime a bit like the old dynamic_reconfigure. ROS2 also allows nodes to reject new parameters in the update callback.  ~~b0027f12fb0b5ae21231d728bc8502872de60bcb limits passing of the node to better isolate plugins, but requires the use of the full parameter API, breaking some of 73217768087a94ade0f839b64ea982c79f76ac7b~~

I'm not completely convinced by the approach I'm taking with the Time APIs, and I'd like some discussion there.
ROS2's major strength is multi-agent robotics, and Fuse has the right feature set to do better than NTP/PTP for clock synchronisation between machines.
Using a rough timestamp to order entries in the graph is probably fine, but I feel there needs to be a convention that allows a Constraint to describe a measurement of a remote clock's time, and time-stamping the measurement with the local clock.

ROS2 has migrated the core packages to Apache2 licenses, I'm happy for my work to remain under the existing BSD or to migrate to  Apache2 for consistency with ROS2


methylDragon:
With the completion of https://github.com/locusrobotics/fuse/issues/276 , closing this PR.
Thanks again @BrettRD for all the work, it was an AWESOME start to build off of!

Issue #276, opened 2022-10-06
Fuse -> ROS 2 Rolling Porting Progress


methylDragon:
# Description

This issue tracks the progress of the ROS 2 port for Fuse, targeting the `rolling` release of ROS 2.
Some of the work here will be based off the work mentioned in https://github.com/locusrobotics/fuse/issues/257.

# Architectural Decisions
I will be prioritising API consistency between the ROS 1 and ROS 2 versions of Fuse. This means I won't be looking at high level architectural changes at this time.

# Approach
PRs will be sent against the `rolling` branch on [fuse](https://github.com/locusrobotics/fuse) and [tf2_2d](https://github.com/locusrobotics/tf2_2d), working up the dependency graph and doubling back as necessary. (Or more likely, finding features to port as I work up the dependency graph, and port them for downstream packages as well.)

Changes in dependent packages that have to be made as a result of a change in a dependency packages will be made on a best effort basis, but can't be confirmed until they're buildable, so expect bugs in those PRs.

I fully expect `fuse_core` to take up the bulk of the migration effort. And I also fully expect needing to **circle back** after things finally build and run.

![image](https://user-images.githubusercontent.com/20265670/194437811-f1792ebc-9d9f-47f6-abb9-412b562db46f.png)

# Progress

## General
- [x] Logging https://github.com/locusrobotics/fuse/pull/279
- [x] Resolve macro usage warnings https://github.com/locusrobotics/fuse/pull/280
- [x] Port Time:
  - [x] https://github.com/locusrobotics/fuse/pull/283
  - [x] https://github.com/locusrobotics/fuse/pull/303
- [x] ~~Port messages https://github.com/locusrobotics/fuse/pull/285~~ (DEFERRED to on a per-package basis)
- [x] Linting

BONUS (if time permits)
- (Probably bonus, too many parts to shake up) Port the pluginlib usage to rclcpp_components
- "Dynamic reconfigure" for params

## Down (Up?) The Dependency Tree
- [x] fuse_msgs
	- [x] https://github.com/locusrobotics/fuse/pull/277
- [x] fuse_core
  - [x] https://github.com/locusrobotics/fuse/pull/281
  - [x] https://github.com/locusrobotics/fuse/pull/284 (depends on #283)
  - [x] https://github.com/locusrobotics/fuse/pull/286
  - [x] https://github.com/locusrobotics/fuse/pull/292
[...]

PR #277, opened 2022-10-07
fuse -> ROS 2 fuse_msgs : Port package and ignore unported packages for now

methylDragon:

This is just an extremely minor port.

It moves the message definitions and uses the ROS 2 message generation pipeline and adds COLCON_IGNORE files to the remaining unported packages.

I also added linting (fixing a `package.xml` linting issue).

PS: Should I be bumping the version number? It should be API breaking, so does that mean it should be a new MAJOR version? Or will it have to wait for release first?

### DIFF
--- a/fuse_msgs/CMakeLists.txt
+++ b/fuse_msgs/CMakeLists.txt
@@ -1,26 +1,26 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.5)
 project(fuse_msgs)
 
-find_package(catkin REQUIRED COMPONENTS
-  message_generation
-  std_msgs
-)
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
 
-add_message_files(
-  DIRECTORY
-    msg
-  FILES
-    SerializedGraph.msg
-    SerializedTransaction.msg
-)
+find_package(ament_cmake REQUIRED)
+find_package(rosidl_default_generators REQUIRED)
+find_package(std_msgs REQUIRED)
 
-generate_messages(
+rosidl_generate_interfaces(
+  ${PROJECT_NAME}
+    "msg/SerializedGraph.msg"
+    "msg/SerializedTransaction.msg"
   DEPENDENCIES
     std_msgs
+  ADD_LINTER_TESTS
 )
 
-catkin_package(
-  CATKIN_DEPENDS
-    message_runtime
-    std_msgs
-)
+ament_export_dependencies(rosidl_default_runtime)
+
+ament_package()
--- a/fuse_msgs/CMakeLists.txt
+++ b/fuse_msgs/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.5)
 project(fuse_msgs)
 
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
 endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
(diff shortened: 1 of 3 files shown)


methylDragon:
Hmm, looks like jenkins is failing on me, but I'm missing some permissions to see what's wrong?
(If I were to guess, it's probably failing because it's trying to catkin build it)

svwilliams:
In case it is helpful, the buildserver error is:
```
[2022-10-07T00:19:40.026Z] ERROR: the following packages/stacks could not have their rosdep keys resolved
[2022-10-07T00:19:40.026Z] to system dependencies:
[2022-10-07T00:19:40.026Z] fuse_msgs: Cannot locate rosdep definition for [rosidl_default_runtime]
script returned exit code 1
```
Pinging @paulbovbel who designed our build system. We are starting the ROS2 port of the fuse package. Are there any steps required to make our build system work with ROS2? Or is the issue that the port is targeting ROS2 Rolling and Locus is using Galactic or whatever official release?

sloretz:
> PS: Should I be bumping the version number? It should be API breaking, so does that mean it should be a new MAJOR version? Or will it have to wait for release first?

Usually we bump the version number at the time of release, but in this case I think we could do it at any time. @svwilliams does Locus use a specific versioning system (such as semver.org?) If so, what version number would you prefer for Rolling? (current is 0.4.2)

> Pinging @paulbovbel who designed our build system. We are starting the ROS2 port of the fuse package. Are there any steps required to make our build system work with ROS2? Or is the issue that the port is targeting ROS2 Rolling and Locus is using Galactic or whatever official release?

@svwilliams How would you feel about us adding a [Rolling PR job](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Build-Farms.html#id6) and using that as CI?

fuse_msgs/CMakeLists.txt
```diff
@@ -1,26 +1,33 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.5)
 project(fuse_msgs)
 
-find_package(catkin REQUIRED COMPONENTS
-  message_generation
-  std_msgs
-)
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
```
sloretz: [C++17 in ROS Rolling](https://www.ros.org/reps/rep-2000.html#rolling-ridley-june-2020-ongoing)

sloretz (changes requested):

svwilliams:
> How would you feel about us adding a Rolling PR job and using that as CI?

That probably makes the most sense. Locus is not currently targeting Rolling with our CI builds anyway. So even if we get the build configuration worked out to build fuse against ROS2, there is still potential for version compatibility problems.

svwilliams:
>Usually we bump the version number at the time of release, but in this case I think we could do it at any time. @svwilliams does Locus use a specific versioning system (such as semver.org?) If so, what version number would you prefer for Rolling? (current is 0.4.2)

Locus also bumps the version numbers at the time of release, but Locus has its own release schedule. Again, I'm going to defer to @paulbovbel for version number recommendations.

fuse_msgs/CMakeLists.txt
methylDragon: Ahh, let me change it

methylDragon: https://github.com/locusrobotics/fuse/compare/5c22c91210203a8b48afe2aaf01409c1c8aa8626..5c8940af69c840a6f7fe4a16af9a7420be658ccd

sloretz (approved):
This looks good to me, though I recommend adding a note to the main repository README saying the Rolling branch is a work in progress with a link to https://github.com/locusrobotics/fuse/issues/276.

In absence of a [PR job](https://github.com/ros/rosdistro/pull/34871) I manually built this repo with `colcon build` using ROS Rolling on Ubuntu Jammy. Only `fuse_msgs` was built. `colcon test` + `colcon test-result` shows all tests passed.

I manually ran a few commands to double check, and the output looks correct.

```bash
jammy:osrf> ros2 interface package fuse_msgs
fuse_msgs/msg/SerializedGraph
fuse_msgs/msg/SerializedTransaction
jammy:osrf> python3 -c 'import fuse_msgs.msg; print(fuse_msgs.msg.SerializedTransaction())'
fuse_msgs.msg.SerializedTransaction(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=0, nanosec=0), frame_id=''), data=[])
jammy:osrf> python3 -c 'import fuse_msgs.msg; print(fuse_msgs.msg.SerializedGraph())'
fuse_msgs.msg.SerializedGraph(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=0, nanosec=0), frame_id=''), plugin_name='', data=[])
```

methylDragon:
> This looks good to me, though I recommend adding a note to the main repository README saying the Rolling branch is a work in progress with a link to #276.
> 
> In absence of a [PR job](https://github.com/ros/rosdistro/pull/34871) I manually built this repo with `colcon build` using ROS Rolling on Ubuntu Jammy. Only `fuse_msgs` was built. `colcon test` + `colcon test-result` shows all tests passed.
> 
> I manually ran a few commands to double check, and the output looks correct.
> 
> ```shell
> jammy:osrf> ros2 interface package fuse_msgs
> fuse_msgs/msg/SerializedGraph
> fuse_msgs/msg/SerializedTransaction
> jammy:osrf> python3 -c 'import fuse_msgs.msg; print(fuse_msgs.msg.SerializedTransaction())'
> fuse_msgs.msg.SerializedTransaction(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=0, nanosec=0), frame_id=''), data=[])
> jammy:osrf> python3 -c 'import fuse_msgs.msg; print(fuse_msgs.msg.SerializedGraph())'
> fuse_msgs.msg.SerializedGraph(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=0, nanosec=0), frame_id=''), plugin_name='', data=[])
> ```

Okie, I'll add that note and then merge! Thanks for the review!

PR #278, opened 2022-10-07
fuse -> ROS 2 : Doc Generation

methylDragon:

Another minor port. This uses rosdoc2 to do the document generation.

Most likely the CI/CD has to be updated to use it!
Also, I tried my best to preserve the settings between rosdoc and rosdoc2, but it's not a one-to-one mapping.

Also, you'll notice that I added `\` to some of the package descriptions in `package.xml`, that's because without them doxygen will truncate the description in the generated docs.

I'll leave it to the maintainers to tweak the doc generation configuration!

Try it out by invoking the following on anything other than fuse_docs (since that uses Sphinx instead of Doxygen):
```shell
$ mkdir some_doc_dir_to_not_pollute_the_workspace
$ cd some_doc_dir_to_not_pollute_the_workspace
$ rosdoc2 build --package-path <path_to_rosdoc2.yaml_directory>
$ rosdoc2 open ./docs_output/fuse_constraints/api/html/index.html
```

Pinging @svwilliams for visibility


fuse_core/package.xml
```diff
@@ -41,7 +41,6 @@
   <test_depend>rclcpp</test_depend>
 
   <export>
-    <build_type>ament_cmake</build_type>
```
sloretz, line 44: need this line to prevent colcon from trying to build this as a catkin pacakge

sloretz: e20fd5d

fuse_doc/package.xml
```diff
@@ -7,15 +8,13 @@
...
-  <buildtool_depend>catkin</buildtool_depend>
-
-  <buildtool_depend condition="$ROS_PYTHON_VERSION == 2">python-catkin-pkg</buildtool_depend>
-  <buildtool_depend condition="$ROS_PYTHON_VERSION == 3">python3-catkin-pkg</buildtool_depend>
+  <buildtool_depend>ament_cmake_ros</buildtool_depend>
 
   <export>
-    <rosdoc config="rosdoc.yaml"/>    
+    <build_type>ament_cmake</build_type>
+    <rosdoc2>"rosdoc2.yaml"<rosdoc2/>
```
sloretz: `<rosdoc2/>`  -> `</rosdoc2>`

sloretz: e20fd5d

fuse_core/rosdoc2.yaml
```diff
@@ -0,0 +1,14 @@
...
+version: 1
+---
+settings:
+  output_dir: ''
+builders:
+  - doxygen: {
+    name: 'doxygen',
+    output_dir: 'api',
+    extra_doxyfile_statements: [
+      "FILE_PATTERNS = '*.cpp *.h",
```
sloretz: `.hpp` now too

sloretz: Actually, I ended up removing the rosdoc2.yaml files except for `fuse_doc`. I found the C++ api docs weren't being generated with the custom configs, but seem to work fine with the default settings.

fuse_doc/doc/.templates/full_globaltoc.html
```diff
@@ -5,16 +5,16 @@ <h3><a href="{{ pathto(master_doc) }}">{{ _('Further Documentation') }}</a></h3>
...
-    <li class="toctree-l1"><a href="../../fuse_optimizers/api/index.html"/>fuse_optimizers</a></li>
-    <li class="toctree-l1"><a href="../../fuse_publishers/api/index.html"/>fuse_publishers</a></li>
-    <li class="toctree-l1"><a href="../../fuse_variables/api/index.html"/>fuse_variables</a></li>
-    <li class="toctree-l1"><a href="../../fuse_viz/api/index.html"/>fuse_viz</a></li>
+    <li class="toctree-l1"><a href="../fuse_constraints/index.html"/>fuse_constraints</a></li>
+    <li class="toctree-l1"><a href="../fuse_core/index.html"/>fuse_core</a></li>
+    <li class="toctree-l1"><a href="../fuse_graphs/index.html"/>fuse_graphs</a></li>
+    <li class="toctree-l1"><a href="../fuse_loss/index.html"/>fuse_loss</a></li>
+    <li class="toctree-l1"><a href="../fuse_models/index.html"/>fuse_models</a></li>
+    <li class="toctree-l1"><a href="../fuse_msgs/index.html"/>fuse_msgs</a></li>
```
sloretz, line 13: This one is special, and a broken link for now. rosdoc2 doesn't seem to generate any files for interface packages. I left it in so that it will hopefully start working once rosdoc2 supports that.

sloretz (approved):
These look good to me locally.

There are a couple issues we can't resolve at the moment.

* The first is fuse_doc requires an upstream fix in `rosdoc2` to render correctly: https://github.com/ros-infrastructure/rosdoc2/pull/47
* The second is `rosdoc2` doesn't currently generate anything for interface packages.

However, doc jobs run regularly without needing a new release, so I don't think we should block on those issues. They will get resolved automatically when those two are resolved upstream.

fuse_core/rosdoc2.yaml
methylDragon: That's very strange.. but not a problem I suppose

PR #279, opened 2022-10-08
fuse -> ROS 2 : Port Logging

methylDragon:

This PR ports logging calls to RCLCPP across the entire stack.

Note that it won't compile since only logging's been ported (and that requires ROS 2 constructs like node pointers, which have yet to be integrated.)

Additionally, the delayed throttle filter was ported to use std::chrono to separate its functionality from ROS. (Also because ROS doesn't support negative time (: )

It's recommended to take a look at how the logging calls have been ported: https://github.com/locusrobotics/fuse/pull/279/commits/3581bd7d96d056a1fa76a13b4cda0d31096fd95b
I also left behind comments for future edits (mostly porting node handles -> nodes)

PS: Would a std::chrono::steady_clock [be preferable for the console throttler](https://github.com/locusrobotics/fuse/pull/279/files#diff-7252082b533722f0a0542d2c417f86dfc98c57db8df36e0d12b036eaebe68009)? It's a bit of a roundabout to be using it if yes, so if it's not critical I'd continue using system_clock.

### DIFF
--- a/fuse_core/include/fuse_core/async_publisher.h
+++ b/fuse_core/include/fuse_core/async_publisher.h
@@ -38,6 +38,7 @@
 #include <fuse_core/fuse_macros.h>
 #include <fuse_core/publisher.h>
 #include <fuse_core/transaction.h>
+// #include <rclcpp/rclcpp.hpp>  TODO(CH3): Uncomment when ready
 #include <ros/callback_queue.h>
 #include <ros/node_handle.h>
 #include <ros/spinner.h>
@@ -135,6 +136,7 @@ public:
 protected:
   ros::CallbackQueue callback_queue_;  //!< The local callback queue used for all subscriptions
   std::string name_;  //!< The unique name for this publisher instance
+  // rclcpp::Node::SharedPtr node_;  //!< The node for this publisher (TODO(CH3): Uncomment when it's time)
   ros::NodeHandle node_handle_;  //!< A node handle in the global namespace using the local callback queue
   ros::NodeHandle private_node_handle_;  //!< A node handle in the private namespace using the local callback queue
   ros::AsyncSpinner spinner_;  //!< A single/multi-threaded spinner assigned to the local callback queue
--- a/fuse_core/include/fuse_core/async_publisher.h
+++ b/fuse_core/include/fuse_core/async_publisher.h
@@ -136,7 +136,7 @@ public:
 protected:
   ros::CallbackQueue callback_queue_;  //!< The local callback queue used for all subscriptions
   std::string name_;  //!< The unique name for this publisher instance
-  // rclcpp::Node::SharedPtr node_;  //!< The node for this publisher (TODO(CH3): Uncomment when it's time)
+  // rclcpp::NodeHandle<???>::SharedPtr nh_;  //!< The node handle for this publisher (TODO(CH3): Uncomment when it's time)
   ros::NodeHandle node_handle_;  //!< A node handle in the global namespace using the local callback queue
   ros::NodeHandle private_node_handle_;  //!< A node handle in the private namespace using the local callback queue
   ros::AsyncSpinner spinner_;  //!< A single/multi-threaded spinner assigned to the local callback queue
--- a/fuse_core/include/fuse_core/async_publisher.h
+++ b/fuse_core/include/fuse_core/async_publisher.h
@@ -136,7 +136,7 @@ public:
 protected:
   ros::CallbackQueue callback_queue_;  //!< The local callback queue used for all subscriptions
   std::string name_;  //!< The unique name for this publisher instance
-  // rclcpp::NodeHandle<???>::SharedPtr nh_;  //!< The node handle for this publisher (TODO(CH3): Uncomment when it's time)
+  // rclcpp::Node::SharedPtr node_;  //!< The node for this publisher (TODO(CH3): Uncomment when it's time)
   ros::NodeHandle node_handle_;  //!< A node handle in the global namespace using the local callback queue
   ros::NodeHandle private_node_handle_;  //!< A node handle in the private namespace using the local callback queue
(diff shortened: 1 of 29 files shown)


fuse_core/include/fuse_core/async_publisher.h
```diff
@@ -135,6 +136,7 @@ class AsyncPublisher : public Publisher
 protected:
   ros::CallbackQueue callback_queue_;  //!< The local callback queue used for all subscriptions
   std::string name_;  //!< The unique name for this publisher instance
+  // rclcpp::Node::SharedPtr node_;  //!< The node for this publisher (TODO(CH3): Uncomment when it's time)
```
sloretz, line 139: This might end up being a shared pointer to [the `NodeTopicsInterface`](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/node_interfaces/node_topics_interface.hpp) , so that lifecycle nodes can be used too

fuse_core/include/fuse_core/ceres_options.h
```diff
@@ -190,7 +190,7 @@ CERES_OPTION_STRING_DEFINITIONS(VisibilityClusteringType)
  * @param[in] default_value - A default value to use if the provided parameter name does not exist
  * @return The loaded (or default) value
  */
-template <class T>
+template <class T>  // TODO(CH3): Replace nh with a node
```
sloretz, line 193: I'd recommend it takes a `NodeLoggingInterface` and `NodeParamsInterface` with another convenience method that gets those off of "Node like" objects. That will allow it to work with Lifecycle nodes too.

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
```diff
@@ -70,7 +70,7 @@ struct Odometry2DPublisherParams : public ParameterBase
    *
    * @param[in] nh - The ROS node handle with which to load parameters
    */
-  void loadFromROS(const ros::NodeHandle& nh) final
+  void loadFromROS(const ros::NodeHandle& nh) final  // TODO(CH3): Replace with a NodePtr
```
sloretz, line 73: NodeParamsInterface probably

fuse_publishers/src/pose_2d_publisher.cpp
```diff
@@ -84,12 +85,14 @@ bool findPose(
   }
   catch (const std::exception& e)
   {
-    ROS_WARN_STREAM_THROTTLE(10.0, "Failed to find a pose at time " << stamp << ". Error" << e.what());
+    RCLCPP_WARN_STREAM_THROTTLE(rclcpp::get_logger("pose_2d_publisher"), rclcpp::Clock(), 10.0 * 1000,
```
sloretz: Any places where we don't have a logger, I'd recommend using the package name as the logger name. I think that will make it easier for users to know where messages are coming from.

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
methylDragon: Comment updated https://github.com/locusrobotics/fuse/pull/279/commits/771e4f4b228a1bb1c4dea24ad07fbcd46502ea71

Pending merge of https://github.com/ros2/rclcpp/pull/2041 , but if we expect that to take too long I can duplicate an implementation in fuse_core...

fuse_core/include/fuse_core/ceres_options.h
methylDragon: Comment updated https://github.com/locusrobotics/fuse/pull/279/commits/771e4f4b228a1bb1c4dea24ad07fbcd46502ea71

Pending merge of https://github.com/ros2/rclcpp/pull/2041 , but if we expect that to take too long I can duplicate an implementation in fuse_core...

fuse_core/include/fuse_core/async_publisher.h
methylDragon: Comment updated https://github.com/locusrobotics/fuse/pull/279/commits/771e4f4b228a1bb1c4dea24ad07fbcd46502ea71

Pending merge of https://github.com/ros2/rclcpp/pull/2041 , but if we expect that to take too long I can duplicate an implementation in fuse_core...

fuse_publishers/src/pose_2d_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/279/commits/56b274b6752572f8314dcaf32bea79c54ae213e5 !

fuse_core/include/fuse_core/async_publisher.h
methylDragon: Reverting the comment update since it turns out we do have to store a SharedPtr to a node after all (for now at least)
https://github.com/locusrobotics/fuse/pull/279/commits/e9a0a40139bd24da809a03adc714ffd415c58683

fuse_core/include/fuse_core/ceres_options.h
methylDragon: Reverting the comment update since it turns out we do have to store a SharedPtr to a node after all (for now at least)
https://github.com/locusrobotics/fuse/pull/279/commits/e9a0a40139bd24da809a03adc714ffd415c58683

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
methylDragon: Reverting the comment update since it turns out we do have to store a SharedPtr to a node after all (for now at least)
https://github.com/locusrobotics/fuse/pull/279/commits/e9a0a40139bd24da809a03adc714ffd415c58683

sloretz (approved):
Approving knowing the changes won't build yet, but are necessary in the long run.

PR #281, opened 2022-10-08
Fuse -> ROS 2 fuse_core: Partial port of fuse_core

methylDragon:
See: #276

This PR gets fuse_core building partially. It is NOT a complete port, since there are some changes that need to be made that affect many libraries downstream (e.g. time and waitables.)

This is a common branching off point for many of those downstream features, so it'll be extremely nice to get this in so that those PRs can use this as a base.

I've incorporated a lot of changes from @BrettRD (credited in the commit messages as appropriate.) And edited some of them to fit with the rest of the commits/PRs. (I know there are some (mostly time related) changes here that are still nascent (and don't reflect the final state of either of the existing branches), I intend to tackle those and repackage the changes in a separate time related PR.)

Also, there are a whole lot of warnings that get emitted, they'll get dealt with in downstream PRs. It'll get messier before it gets neater :>

One way to help reduce the number of warnings is to get https://github.com/locusrobotics/fuse/pull/280 merged.

Also currently I think this should play well with the rest of the PRs currently open? But it'll be good to try to get this and #280 merged as quickly as possible. (Probably before https://github.com/locusrobotics/fuse/pull/279 , though https://github.com/locusrobotics/fuse/pull/279 should get in immediately after, I think)

### DIFF
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -8,12 +8,13 @@ if(NOT CMAKE_CXX_STANDARD)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-Wall -Werror -Wextra -Wpedantic)
+  add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(fuse_msgs REQUIRED)
 find_package(pluginlib REQUIRED)
+find_package(rcl_interfaces REQUIRED)
 find_package(rclcpp REQUIRED)
 
 find_package(Boost REQUIRED COMPONENTS serialization)
@@ -22,6 +23,17 @@ find_package(Eigen3 REQUIRED)
 find_package(PkgConfig REQUIRED)
 pkg_check_modules(libglog REQUIRED)  # Used downstream
 
+set(ament_dependencies
+  fuse_msgs
+  pluginlib
+  rcl_interfaces
+  rclcpp
+)
+
+###########
+## Build ##
+###########
+
 ## fuse_core library
 add_library(
   ${PROJECT_NAME} SHARED
@@ -29,23 +41,18 @@ add_library(
 #    src/async_publisher.cpp
 #    src/async_sensor_model.cpp
 #    src/ceres_options.cpp
-#    src/constraint.cpp
-#    src/graph.cpp
-#    src/graph_deserializer.cpp
-#    src/loss.cpp
-#    src/serialization.cpp
-#    src/timestamp_manager.cpp
-#    src/transaction.cpp
-#    src/transaction_deserializer.cpp
-#    src/uuid.cpp
-#    src/variable.cpp
+    src/constraint.cpp
+    src/graph.cpp
+    src/graph_deserializer.cpp
+    src/loss.cpp
+    src/serialization.cpp
+    src/timestamp_manager.cpp
+    src/transaction.cpp
+    src/transaction_deserializer.cpp
+    src/uuid.cpp
+    src/variable.cpp
 )
-ament_target_dependencies(
-  ${PROJECT_NAME}
-    fuse_msgs
-    rclcpp
-    std_msgs
-)
+ament_target_dependencies(${PROJECT_NAME} ${ament_dependencies})
 target_include_directories(
   ${PROJECT_NAME} PUBLIC
     include
@@ -61,22 +68,12 @@ target_link_libraries(
     ${GLOG_LIBRARIES}
 )
 
+
 # fuse_echo executable
-# add_executable(fuse_echo
-#   src/fuse_echo.cpp
-# )
-# target_include_directories(fuse_echo
-#   PUBLIC
-#     include
-# )
-# target_link_libraries(fuse_echo
-#   ${PROJECT_NAME}
-# )
-# ament_target_dependencies(fuse_echo
-#   fuse_msgs
-#   rclcpp
-#   pluginlib
-# )
+add_executable(fuse_echo src/fuse_echo.cpp)
+ament_target_dependencies(fuse_echo ${ament_dependencies})
+target_include_directories(fuse_echo PUBLIC include)
+target_link_libraries(fuse_echo ${PROJECT_NAME})
 
 #############
 ## Testing ##
@@ -490,25 +487,19 @@ endif()
 ## Install ##
 #############
 
-set(dependencies
-  fuse_msgs
-  pluginlib
-  rclcpp
-)
-
 ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
-ament_export_dependencies(${dependencies})
+ament_export_dependencies(${ament_dependencies})
 
 install(
   TARGETS ${PROJECT_NAME}
   DESTINATION lib/${PROJECT_NAME}
 )
 
-# install(
-#   TARGETS fuse_echo
-#   RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
-# )
+install(
+  TARGETS fuse_echo
+  DESTINATION lib/${PROJECT_NAME}
+)
 
 install(DIRECTORY include/
   DESTINATION include
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.5)
 project(fuse_core)
 
-# Default to C++14
+# Default to C++17
 if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 17)
   set(CXX_STANDARD_REQUIRED YES)
@@ -18,27 +18,18 @@ find_package(rcl_interfaces REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rclcpp_components REQUIRED)
 
-find_package(Boost REQUIRED COMPONENTS serialization)
 find_package(Ceres REQUIRED)
 find_package(Eigen3 REQUIRED)
-find_package(PkgConfig REQUIRED)
-pkg_check_modules(libglog REQUIRED)  # Used downstream
+find_package(glog REQUIRED)  # Used downstream
 
-set(ament_dependencies
-  fuse_msgs
-  pluginlib
-  rcl_interfaces
-  rclcpp
-  rclcpp_components
-)
+include(boost-extras.cmake)
 
 ###########
 ## Build ##
 ###########
 
 ## fuse_core library
-add_library(
-  ${PROJECT_NAME} SHARED
+add_library(${PROJECT_NAME} SHARED
 #    src/async_motion_model.cpp
 #    src/async_publisher.cpp
 #    src/async_sensor_model.cpp
@@ -54,28 +45,27 @@ add_library(
     src/uuid.cpp
     src/variable.cpp
 )
-ament_target_dependencies(${PROJECT_NAME} ${ament_dependencies})
-target_include_directories(
-  ${PROJECT_NAME} PUBLIC
-    include
-    ${Boost_INCLUDE_DIRS}
-    ${CERES_INCLUDE_DIRS}
-    ${EIGEN3_INCLUDE_DIRS}
-    ${GLOG_INCLUDE_DIRS}
+target_include_directories(${PROJECT_NAME} PUBLIC
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+  "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>"
 )
-target_link_libraries(
-  ${PROJECT_NAME}
-    ${Boost_LIBRARIES}
-    ${CERES_LIBRARIES}
-    ${GLOG_LIBRARIES}
+target_link_libraries(${PROJECT_NAME} PUBLIC
+  Boost::serialization
+  Ceres::ceres
+  Eigen3::Eigen
+  ${fuse_msgs_TARGETS}
+  pluginlib::pluginlib
+  rclcpp::rclcpp
+  ${rcl_interfaces_TARGETS}
+)
+target_link_libraries(${PROJECT_NAME} PRIVATE
+  rclcpp_components::component
 )
 
-
-# fuse_echo executable
+## fuse_echo executable
 add_executable(fuse_echo src/fuse_echo.cpp)
-ament_target_dependencies(fuse_echo ${ament_dependencies})
-target_include_directories(fuse_echo PUBLIC include)
-target_link_libraries(fuse_echo ${PROJECT_NAME})
+target_link_libraries(fuse_echo PUBLIC ${PROJECT_NAME})
+target_link_libraries(fuse_echo PRIVATE rclcpp_components::component)
 
 #############
 ## Testing ##
@@ -489,11 +479,7 @@ endif()
 ## Install ##
 #############
 
-ament_export_include_directories(include)
-ament_export_libraries(${PROJECT_NAME})
-ament_export_dependencies(${ament_dependencies})
-
-install(TARGETS ${PROJECT_NAME}
+install(TARGETS ${PROJECT_NAME} EXPORT ${PROJECT_NAME}-export
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
@@ -505,9 +491,10 @@ install(
 )
 
 install(DIRECTORY include/
-  DESTINATION include
+  DESTINATION include/${PROJECT_NAME}
 )
 
-ament_package(
-  CONFIG_EXTRAS
-)
+ament_export_targets(${PROJECT_NAME}-export HAS_LIBRARY_TARGET)
+ament_export_dependencies(${ament_dependencies})
+
+ament_package(CONFIG_EXTRAS boost-extras.cmake)
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -16,6 +16,7 @@ find_package(fuse_msgs REQUIRED)
 find_package(pluginlib REQUIRED)
 find_package(rcl_interfaces REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(rclcpp_components REQUIRED)
 
 find_package(Boost REQUIRED COMPONENTS serialization)
 find_package(Ceres REQUIRED)
@@ -28,6 +29,7 @@ set(ament_dependencies
   pluginlib
   rcl_interfaces
   rclcpp
+  rclcpp_components
 )
 
 ###########
@@ -491,9 +493,10 @@ ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
 ament_export_dependencies(${ament_dependencies})
 
-install(
-  TARGETS ${PROJECT_NAME}
-  DESTINATION lib/${PROJECT_NAME}
+install(TARGETS ${PROJECT_NAME}
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
 )
 
 install(
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -20,7 +20,6 @@ find_package(rclcpp_components REQUIRED)
 
 find_package(Ceres REQUIRED)
 find_package(Eigen3 REQUIRED)
-find_package(glog REQUIRED)  # Used downstream
 
 include(boost-extras.cmake)
 
@@ -63,9 +62,15 @@ target_link_libraries(${PROJECT_NAME} PRIVATE
 )
 
 ## fuse_echo executable
-add_executable(fuse_echo src/fuse_echo.cpp)
-target_link_libraries(fuse_echo PUBLIC ${PROJECT_NAME})
-target_link_libraries(fuse_echo PRIVATE rclcpp_components::component)
+add_library(fuse_echo_component SHARED src/fuse_echo.cpp)
+target_link_libraries(fuse_echo_component PUBLIC ${PROJECT_NAME})
+target_link_libraries(fuse_echo_component PRIVATE rclcpp_components::component)
+
+rclcpp_components_register_node(fuse_echo_component
+  PLUGIN "fuse_core::FuseEcho"
+  EXECUTABLE fuse_echo
+  EXECUTOR SingleThreadedExecutor
+)
 
 #############
 ## Testing ##
@@ -485,9 +490,10 @@ install(TARGETS ${PROJECT_NAME} EXPORT ${PROJECT_NAME}-export
   RUNTIME DESTINATION bin
 )
 
-install(
-  TARGETS fuse_echo
-  DESTINATION lib/${PROJECT_NAME}
+install(TARGETS fuse_echo_component
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
 )
 
 install(DIRECTORY include/
@@ -495,6 +501,15 @@ install(DIRECTORY include/
 )
 
 ament_export_targets(${PROJECT_NAME}-export HAS_LIBRARY_TARGET)
-ament_export_dependencies(${ament_dependencies})
+ament_export_dependencies(
+  ament_cmake
+  fuse_msgs
+  pluginlib
+  rcl_interfaces
+  rclcpp
+  rclcpp_components
+  Ceres
+  Eigen3
+)
 
 ament_package(CONFIG_EXTRAS boost-extras.cmake)
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -1,519 +1,519 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.5)
 project(fuse_core)
 
-set(build_depends
-  fuse_msgs
-  pluginlib
-  roscpp
-)
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 17)
+  set(CXX_STANDARD_REQUIRED YES)
+endif()
 
-find_package(catkin REQUIRED COMPONENTS
-  ${build_depends}
-)
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Werror -Wextra -Wpedantic)
+endif()
+
+find_package(ament_cmake REQUIRED)
+find_package(fuse_msgs REQUIRED)
+find_package(pluginlib REQUIRED)
+find_package(rclcpp REQUIRED)
+
 find_package(Boost REQUIRED COMPONENTS serialization)
 find_package(Ceres REQUIRED)
 find_package(Eigen3 REQUIRED)
 find_package(PkgConfig REQUIRED)
-pkg_check_modules(libglog REQUIRED)
-
-catkin_package(
-  INCLUDE_DIRS
-    include
-  LIBRARIES
-    ${PROJECT_NAME}
-  CATKIN_DEPENDS
-    ${build_depends}
-  DEPENDS
-    Boost
-    CERES
-    EIGEN3
-    GLOG
-)
-
-###########
-## Build ##
-###########
-add_compile_options(-Wall -Werror)
+pkg_check_modules(libglog REQUIRED)  # Used downstream
 
 ## fuse_core library
-add_library(${PROJECT_NAME}
-  src/async_motion_model.cpp
-  src/async_publisher.cpp
-  src/async_sensor_model.cpp
-  src/ceres_options.cpp
-  src/constraint.cpp
-  src/graph.cpp
-  src/graph_deserializer.cpp
-  src/loss.cpp
-  src/serialization.cpp
-  src/timestamp_manager.cpp
-  src/transaction.cpp
-  src/transaction_deserializer.cpp
-  src/uuid.cpp
-  src/variable.cpp
+add_library(
+  ${PROJECT_NAME} SHARED
+#    src/async_motion_model.cpp
+#    src/async_publisher.cpp
+#    src/async_sensor_model.cpp
+#    src/ceres_options.cpp
+#    src/constraint.cpp
+#    src/graph.cpp
+#    src/graph_deserializer.cpp
+#    src/loss.cpp
+#    src/serialization.cpp
+#    src/timestamp_manager.cpp
+#    src/transaction.cpp
+#    src/transaction_deserializer.cpp
+#    src/uuid.cpp
+#    src/variable.cpp
 )
-add_dependencies(${PROJECT_NAME}
-  ${catkin_EXPORTED_TARGETS}
+ament_target_dependencies(
+  ${PROJECT_NAME}
+    fuse_msgs
+    rclcpp
+    std_msgs
 )
-target_include_directories(${PROJECT_NAME}
-  PUBLIC
+target_include_directories(
+  ${PROJECT_NAME} PUBLIC
     include
     ${Boost_INCLUDE_DIRS}
-    ${catkin_INCLUDE_DIRS}
     ${CERES_INCLUDE_DIRS}
     ${EIGEN3_INCLUDE_DIRS}
     ${GLOG_INCLUDE_DIRS}
 )
-target_link_libraries(${PROJECT_NAME}
-  ${Boost_LIBRARIES}
-  ${catkin_LIBRARIES}
-  ${CERES_LIBRARIES}
-  ${GLOG_LIBRARIES}
-)
-set_target_properties(${PROJECT_NAME}
-  PROPERTIES
-    CXX_STANDARD 14
-    CXX_STANDARD_REQUIRED YES
+target_link_libraries(
+  ${PROJECT_NAME}
+    ${Boost_LIBRARIES}
+    ${CERES_LIBRARIES}
+    ${GLOG_LIBRARIES}
 )
 
 # fuse_echo executable
-add_executable(fuse_echo
-  src/fuse_echo.cpp
-)
-add_dependencies(fuse_echo
-  ${catkin_EXPORTED_TARGETS}
-)
-target_include_directories(fuse_echo
-  PUBLIC
-    include
-    ${catkin_INCLUDE_DIRS}
-)
-target_link_libraries(fuse_echo
-  ${PROJECT_NAME}
-  ${catkin_LIBRARIES}
-)
-set_target_properties(fuse_echo
-  PROPERTIES
-    CXX_STANDARD 14
-    CXX_STANDARD_REQUIRED YES
-)
-
-#############
-## Install ##
-#############
-
-install(
-  TARGETS ${PROJECT_NAME}
-  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION}
-)
-
-install(
-  TARGETS fuse_echo
-  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
-)
-
-install(
-  DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
-)
+# add_executable(fuse_echo
+#   src/fuse_echo.cpp
+# )
+# target_include_directories(fuse_echo
+#   PUBLIC
+#     include
+# )
+# target_link_libraries(fuse_echo
+#   ${PROJECT_NAME}
+# )
+# ament_target_dependencies(fuse_echo
+#   fuse_msgs
+#   rclcpp
+#   pluginlib
+# )
 
 #############
 ## Testing ##
 #############
 
-if(CATKIN_ENABLE_TESTING)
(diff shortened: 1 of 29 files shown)


methylDragon:
@BrettRD could I double check what you meant by `unsafe time packing` and `deprecated time API` here: https://github.com/locusrobotics/fuse/pull/281/files#diff-4d7269f5ad306527f06d77e64f5230b27087a3be72516585a9e0dc9286db6116 ?

I'm guessing deprecated time API is from the ROS1 split of time into nsec and sec? What about packing?

BrettRD:
Unsafe packing is endianness portability, the same timestamp will produce different results on different processor architectures

The time API is just the sec/nsec split, ROS2 and chrono offer clearer options

methylDragon:
Oof, I suppose it's time to use boost endian then... (we're on C++17 so we can't use endian from std yet :( )

I'm also pretty tempted to just use the nanoseconds from ROS2 instead of both nanoseconds and seconds. Does anyone from Locus know if the byte buffer inputs to the UUID generator HAVE to encode seconds AND nanoseconds? I would presume no 🤔

BrettRD:
We don't need an endianness lib, just an explicit byte packing order instead of the technically undefined behaviour of reinterpret_cast.
The compiler will produce the same behaviour and binary on little endian, but it'll have consistent output when compiled for other architectures. (Good habits from the embedded systems world)

The uuid functions as a hash, so it only needs to use every meaningful bit in the time value

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,513 @@
...
   TARGETS fuse_echo
-  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+  DESTINATION lib/${PROJECT_NAME}
 )
 
-install(
-  DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+install(DIRECTORY include/
+  DESTINATION include
```
sloretz: Install to `include/${PROJECT_NAME}` to avoid incorrect headers being found when overriding packages

```CMake
install(DIRECTORY include/ DESTINATION include/${PROJECT_NAME})
```

https://colcon.readthedocs.io/en/released/user/overriding-packages.html#install-headers-to-a-unique-include-directory

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,513 @@
...
-  ${catkin_LIBRARIES}
-)
-set_target_properties(fuse_echo
-  PROPERTIES
-    CXX_STANDARD 14
-    CXX_STANDARD_REQUIRED YES
-)
+add_executable(fuse_echo src/fuse_echo.cpp)
+ament_target_dependencies(fuse_echo ${ament_dependencies})
+target_include_directories(fuse_echo PUBLIC include)
```
sloretz: Use `$<BUILD_INTERFACE:>` and `$<INSTALL_INTERFACE:..>` so we can export modern CMake targets that have include directory info built in

```CMake
target_include_directories(${PROJECT_NAME} INTERFACE
  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
  "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>")
```

Install interface has extra `${PROJECT_NAME}` to allow correct headers to be found when overriding packages.

https://colcon.readthedocs.io/en/released/user/overriding-packages.html#install-headers-to-a-unique-include-directory

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,513 @@
...
 ## Install ##
 #############
 
-install(
-  TARGETS ${PROJECT_NAME}
-  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION}
+ament_export_include_directories(include)
+ament_export_libraries(${PROJECT_NAME})
```
sloretz: No need for `ament_export_include_directories` or `ament_export_libraries` these days. Those use old-style CMake variables intead of modern CMake targets. Instead use `ament_export_targets`

```
ament_export_targets(${PROJECT_NAME}-export HAS_LIBRARY_TARGET)
```

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,513 @@
...
+ament_export_dependencies(${ament_dependencies})
+
+install(TARGETS ${PROJECT_NAME}
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
```
sloretz: Export a modern CMake target


```cmake
install(TARGETS ${PROJECT_NAME} EXPORT ${PROJECT_NAME}-export
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
  RUNTIME DESTINATION bin
)
```

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,513 @@
...
 #############
 
-install(
-  TARGETS ${PROJECT_NAME}
-  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION}
+ament_export_include_directories(include)
+ament_export_libraries(${PROJECT_NAME})
+ament_export_dependencies(${ament_dependencies})
```
sloretz: Need to export anything we `find_package()` that downstream packages also need `find_package()`'d, not just depependencies using ament CMake. At a minimum this is any package whose headers we include in our own public headers.

For example, need to `ament_export_dependencies()` Ceres: https://github.com/locusrobotics/fuse/blob/07676a8a47ad00b605153d6d98aa58dce663a39e/fuse_core/include/fuse_core/autodiff_local_parameterization.h#L41

Also need to do the same for Eigen: https://github.com/locusrobotics/fuse/blob/07676a8a47ad00b605153d6d98aa58dce663a39e/fuse_core/include/fuse_core/eigen.h#L37-L38

We also need to do the same for `boost`: https://github.com/locusrobotics/fuse/blob/07676a8a47ad00b605153d6d98aa58dce663a39e/fuse_core/include/fuse_core/constraint.h#L42-L44  - but `ament_export_dependencies()` doesn't support `find_package()`ing components, so we may need to do that in an extra CMake file.

I think we can skip the `gtest` one though: https://github.com/locusrobotics/fuse/blob/07676a8a47ad00b605153d6d98aa58dce663a39e/fuse_core/include/fuse_core/eigen_gtest.h#L38 as gtest likes to be built in every project rather than being `find_package()`'d.  Any downstream package using `fuse_core/eigen_gtest.h` will have to be aware that it needs to depend on gtest on it's own.

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,513 @@
...
+    src/transaction_deserializer.cpp
+    src/uuid.cpp
+    src/variable.cpp
 )
-add_dependencies(${PROJECT_NAME}
-  ${catkin_EXPORTED_TARGETS}
-)
-target_include_directories(${PROJECT_NAME}
-  PUBLIC
+ament_target_dependencies(${PROJECT_NAME} ${ament_dependencies})
```
sloretz: Now that ROS 2 is using modern CMake targets days I'd recommend using `target_link_libraries()` only instead of `ament_target_dependencies()` and `target_include_directories()`. https://github.com/ament/ament_cmake/issues/292

```CMake
target_link_libraries(${PROJECT_NAME} PUBLIC
    boost::serialization
    Ceres::ceres
    Eigen3::Eigen
    ${fuse_msgs_TARGETS}
    pluginlib::pluginlib
    rclcpp::rclcpp
    rclcpp_components::component
    ${rcl_interfaces_TARGETS})
```

If any of these are only used in source files, not public header files, then we could move them to a second `target_link_libraries()` call with the keyword `PRIVATE`. Probably `rclcpp_components::component` is that way, but I didn't look at all the header files to make sure.

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,513 @@
...
 
-find_package(catkin REQUIRED COMPONENTS
-  ${build_depends}
-)
 find_package(Boost REQUIRED COMPONENTS serialization)
 find_package(Ceres REQUIRED)
 find_package(Eigen3 REQUIRED)
 find_package(PkgConfig REQUIRED)
-pkg_check_modules(libglog REQUIRED)
+pkg_check_modules(libglog REQUIRED)  # Used downstream
```
sloretz: It looks like the version of `libgoogle-glog-dev` comes with a `<project>-config.cmake` :tada:

That means you can replace this with `find_packge(glog REQUIRED)` and link against it with `target_link_libraries(my_target PUBLIC|PRIVATE glog::glog)`

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,513 @@
...
-  ${PROJECT_NAME}
-  ${catkin_LIBRARIES}
-)
-set_target_properties(fuse_echo
-  PROPERTIES
-    CXX_STANDARD 14
-    CXX_STANDARD_REQUIRED YES
-)
+add_executable(fuse_echo src/fuse_echo.cpp)
+ament_target_dependencies(fuse_echo ${ament_dependencies})
```
sloretz: Same comment about not using `ament_target_dependencies()`.

fuse_core/include/fuse_core/console.h
```diff
@@ -68,14 +67,16 @@ class DelayedThrottleFilter : public ros::console::FilterBase
    */
   bool isEnabled() override
   {
-    const auto now = ros::Time::now().toSec();
+    #warn "migrated from ros1, using default clock"
+    const auto now = rclcpp::Clock().now().seconds();
```
sloretz, line 71: Ah. I'd recommend requiring an `rclcpp::Clock::SharedPtr` be passed via the constructor.

At least it looks like this is only used in one place: https://github.com/locusrobotics/fuse/blob/07676a8a47ad00b605153d6d98aa58dce663a39e/fuse_models/include/fuse_models/odometry_2d_publisher.h#L207

fuse_core/include/fuse_core/console.h
```diff
@@ -68,14 +67,16 @@ class DelayedThrottleFilter : public ros::console::FilterBase
...
 
     if (last_hit_ < 0.0)
     {
       last_hit_ = now;
+      return true;
     }
 
-    if (ROSCONSOLE_THROTTLE_CHECK(now, last_hit_, period_))
+    if (now > (last_hit_ + period_))
```
sloretz, line 79: Seems like it should be `now >= ...` instead of `now > ...`

fuse_core/include/fuse_core/graph_deserializer.h
```diff
@@ -73,7 +73,7 @@ class GraphDeserializer
    * @param[in]  msg  The SerializedGraph message to be deserialized
    * @return          A unique_ptr to a derived Graph object
    */
-  fuse_core::Graph::UniquePtr deserialize(const fuse_msgs::SerializedGraph::ConstPtr& msg) const;
+  fuse_core::Graph::UniquePtr deserialize(const fuse_msgs::msg::SerializedGraph::SharedPtr msg) const;
```
sloretz: I think the `ConstPtr` type in ROS 1 maps to the `ConstSharedPtr` type in ROS 2. That said, it looks like this is a convenience API that calls the lower one. It would make sense to me to make this function `inline` so a compiler could choose to optimize it out.

fuse_core/include/fuse_core/message_buffer_impl.h
```diff
@@ -131,17 +131,17 @@ template<class Message>
...
+  if ((buffer_length_ == rclcpp::Duration::max()) || (buffer_.size() <= 2))
   {
     return;
   }
 
   // Compute the expiration time carefully, as ROS can't handle negative times
   const auto& ending_stamp = buffer_.back().first;
   auto expiration_time =
-      ending_stamp.toSec() > buffer_length_.toSec() ? ending_stamp - buffer_length_ : ros::Time(0, 0);
+      ending_stamp.toSec() > buffer_length_.toSec() ? ending_stamp - buffer_length_ : rclcpp::Time(0, 0);
```
sloretz, line 144: `rclcpp::Time(0,0)` is tricky because [the default clock with this constructor is system time](https://github.com/ros2/rclcpp/blob/3aca271ef5d25c1b8c12e67ecf85166856a88000/rclcpp/include/rclcpp/time.hpp#L46), and that would cause a comparison failure if a different clock was used in a time in the `buffer_`. I'd recommend `rclcpp::Time(0, 0, ending_stamp.get_clock_type())` instead so the comparisons will work assuming all times in the buffer have the same clock type.

fuse_core/include/fuse_core/throttled_callback.h
```diff
@@ -60,11 +60,12 @@ class ThrottledCallback
...
    * @param[in] drop_callback   The callback to call when dropped because of the throttling. Defaults to nullptr
    * @param[in] throttle_period The throttling period duration in seconds. Defaults to 0.0, i.e. no throttling
-   * @param[in] use_wall_time   Whether to use ros::WallTime or not. Defaults to false
+   * @param[in] use_wall_time   Whether to use wall time or not. Defaults to false
    */
   ThrottledCallback(Callback&& keep_callback = nullptr,  // NOLINT(whitespace/operators)
                     Callback&& drop_callback = nullptr,  // NOLINT(whitespace/operators)
-                    const ros::Duration& throttle_period = ros::Duration(0.0), const bool use_wall_time = false)
+                    const rclcpp::Duration& throttle_period = rclcpp::Duration(0,0),
+                    const bool use_wall_time = false)
```
sloretz, line 73: I'd recommend requiring the clock to be passed in instead of a `use_wall_time` flag.

fuse_core/src/fuse_echo.cpp
```diff
@@ -31,64 +31,76 @@
...
   fuse_core::GraphDeserializer graph_deserializer_;
   fuse_core::TransactionDeserializer transaction_deserializer_;
-  ros::NodeHandle node_handle_;
-  ros::Subscriber graph_subscriber_;
-  ros::Subscriber transaction_subscriber_;
+  rclcpp::Subscription<fuse_msgs::msg::SerializedGraph>::SharedPtr graph_subscriber_;
+  rclcpp::Subscription<fuse_msgs::msg::SerializedTransaction>::SharedPtr transaction_subscriber_;
 
-  void graphCallback(const fuse_msgs::SerializedGraph::ConstPtr& msg)
+  void graphCallback(const fuse_msgs::msg::SerializedGraph::SharedPtr msg)
```
sloretz: I'm a little rusty on what type is most efficient for intraprocess, but I think it' might be `const MsgT& msg)` here instead of `const MsgT::SharedPtr`. That way if the intra-process publisher published the message with a different type then `rclcpp` wouldn't have to create a `SharedPtr` instance to pass to this function. William would be the best person to ask though.

fuse_core/src/fuse_echo.cpp
```diff
@@ -31,64 +31,76 @@
...
     std::cout << "-------------------------" << std::endl;
     std::cout << "GRAPH:" << std::endl;
-    std::cout << "received at: " << ros::Time::now() << std::endl;
+    std::cout << "received at: " << rclcpp::Clock().now().seconds() << std::endl;
     auto graph = graph_deserializer_.deserialize(msg);
     graph->print();
   }
 
-  void transactionCallback(const fuse_msgs::SerializedTransaction::ConstPtr& msg)
+  void transactionCallback(const fuse_msgs::msg::SerializedTransaction::SharedPtr msg)
```
sloretz: Same comment about callback type.

fuse_core/src/fuse_echo.cpp
```diff
@@ -31,64 +31,76 @@
...
   {
     std::cout << "-------------------------" << std::endl;
     std::cout << "TRANSACTION:" << std::endl;
-    std::cout << "received at: " << ros::Time::now() << std::endl;
+    std::cout << "received at: " << rclcpp::Clock().now().seconds() << std::endl;
```
sloretz: Since `FuseEcho` is a node, I'd use the node's clock with [`this->now()`](https://github.com/ros2/rclcpp/blob/3aca271ef5d25c1b8c12e67ecf85166856a88000/rclcpp/include/rclcpp/node.hpp#L1399).

fuse_core/src/uuid.cpp
```diff
@@ -65,30 +59,47 @@ UUID generate()
...
 }
 
-UUID generate(const std::string& namespace_string, const ros::Time& stamp, const UUID& id)
+UUID generate(const std::string& namespace_string, const rclcpp::Time& stamp, const UUID& id)
 {
-  constexpr size_t buffer_size = sizeof(stamp.sec) + sizeof(stamp.nsec) + UUID::static_size();
+  // replicate ros1 behaviour for now
+  #warning "deprecated time API"
+  rcl_time_point_value_t stamp_val = stamp.nanoseconds();
+  int32_t stamp_s = stamp_val / (rcl_time_point_value_t)1e9;
```
sloretz: Use [`RCUTILS_NS_TO_S`](https://github.com/ros2/rcutils/blob/c83c4c7b197e72527fc9e0b438f9f1e6225f70af/include/rcutils/time.h#L39). IIRC it avoids some weird behavior on some compilers.

sloretz (changes requested):
Looks like a good start

fuse_core/include/fuse_core/message_buffer_impl.h
BrettRD: The only places I've found time zero used are when:
* the logic is actually testing for uninitialised variables, not time values
* the logic has set a time value to an invalid state to force the result of a conditional statement (this example)
* both

`StampedVariableSynchronizer` uses it quite a bit and gets twisty as a result.
It might make sense to try patching out the use of uninitialised time in the ROS1 version while we figure out the rest of the requirements for a fuse_core time class

fuse_core/CMakeLists.txt
methylDragon: https://github.com/locusrobotics/fuse/pull/281/commits/232dd5ae01033a662e18294b0ca7d029c92bc239

fuse_core/CMakeLists.txt
methylDragon: https://github.com/locusrobotics/fuse/pull/281/commits/232dd5ae01033a662e18294b0ca7d029c92bc239

fuse_core/CMakeLists.txt
methylDragon: https://github.com/locusrobotics/fuse/pull/281/commits/232dd5ae01033a662e18294b0ca7d029c92bc239

fuse_core/CMakeLists.txt
methylDragon: https://github.com/locusrobotics/fuse/pull/281/commits/232dd5ae01033a662e18294b0ca7d029c92bc239

fuse_core/CMakeLists.txt
methylDragon: https://github.com/locusrobotics/fuse/pull/281/commits/232dd5ae01033a662e18294b0ca7d029c92bc239

fuse_core/CMakeLists.txt
methylDragon: https://github.com/locusrobotics/fuse/pull/281/commits/232dd5ae01033a662e18294b0ca7d029c92bc239

fuse_core/CMakeLists.txt
methylDragon: https://github.com/locusrobotics/fuse/pull/281/commits/232dd5ae01033a662e18294b0ca7d029c92bc239

fuse_core/CMakeLists.txt
methylDragon: https://github.com/locusrobotics/fuse/pull/281/commits/232dd5ae01033a662e18294b0ca7d029c92bc239

fuse_core/src/uuid.cpp
methylDragon: There's no RCUTILS macro for getting the nanoseconds section, but I tried to use similar logic.
https://github.com/locusrobotics/fuse/pull/281/commits/4473a1d7231941e2b5081041e229405b0f398552

uuid will continue to change in future PRs though, some of these changes might get obsoleted.

fuse_core/src/fuse_echo.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/281/commits/c32719142e724537a9fe8060aa216d6d8fa62583

fuse_core/src/fuse_echo.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/281/commits/c32719142e724537a9fe8060aa216d6d8fa62583

fuse_core/include/fuse_core/graph_deserializer.h
methylDragon: [`e5c0244` (#281)](https://github.com/locusrobotics/fuse/pull/281/commits/e5c024494feb1422b941d29a854654bf612a0ba0)

fuse_core/src/fuse_echo.cpp
methylDragon: https://github.com/locusrobotics/fuse/commit/c32719142e724537a9fe8060aa216d6d8fa62583

fuse_core/include/fuse_core/console.h
methylDragon: I ended up changing the implementation to use std::chrono instead (in the logging PR): https://github.com/locusrobotics/fuse/pull/279/files#diff-7252082b533722f0a0542d2c417f86dfc98c57db8df36e0d12b036eaebe68009
Using chrono in that case allows the filter to not rely on ROS also

My reasoning is that it's a console filter, so I'd be more concerned about throttling in real-time as opposed to sim-time, I think?

fuse_core/include/fuse_core/console.h
methylDragon: Changed in https://github.com/locusrobotics/fuse/pull/279/files#diff-7252082b533722f0a0542d2c417f86dfc98c57db8df36e0d12b036eaebe68009 (different PR)

fuse_core/include/fuse_core/message_buffer_impl.h
methylDragon: > `rclcpp::Time(0,0)` is tricky because [the default clock with this constructor is system time](https://github.com/ros2/rclcpp/blob/3aca271ef5d25c1b8c12e67ecf85166856a88000/rclcpp/include/rclcpp/time.hpp#L46), and that would cause a comparison failure if a different clock was used in a time in the `buffer_`. I'd recommend `rclcpp::Time(0, 0, ending_stamp.get_clock_type())` instead so the comparisons will work assuming all times in the buffer have the same clock type.

I ended up converging on this idea ahaha

(In a separate PR)
https://github.com/locusrobotics/fuse/pull/283/files#diff-0b68cb1be412632cb7ced5a187501a6214a908fd5c4d29b23af61f797ecb1440

fuse_core/include/fuse_core/throttled_callback.h
methylDragon: This so far has been deferred: https://github.com/locusrobotics/fuse/pull/283/files#diff-250689f3de589b83ad938a77e9a67784dd33d4321363870ba7c1cf742182c44c

The way it's implemented currently is that there's a boolean ROS parameter ([throttle_use_wall_time](https://github.com/locusrobotics/fuse/blob/devel/fuse_models/include/fuse_models/parameters/acceleration_2d_params.h#L75)) though (it then eventually gets passed as the flag).. If we are to pass in a clock type, does that parameter have to change as well? If so what parameter should it become? (String? Int?)

I'm inclined to keep the flag and leave the boolean ROS parameter alone so there's parameter parity between the ROS 1 and ported version of Fuse.

fuse_core/src/uuid.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/files#diff-4d7269f5ad306527f06d77e64f5230b27087a3be72516585a9e0dc9286db6116

Obsoleted!

methylDragon:
> We don't need an endianness lib, just an explicit byte packing order instead of the technically undefined behaviour of reinterpret_cast.
The compiler will produce the same behaviour and binary on little endian, but it'll have consistent output when compiled for other architectures. (Good habits from the embedded systems world)

> The uuid functions as a hash, so it only needs to use every meaningful bit in the time value

Apologies, I'm not too sure if I fully understand.

Do you mean iterating from the ref pointer (e.g. with an explicit for loop.. though we'd still end up needing to cast somehow, like say... `(unsigned char*)&nanoseconds;`, copying each time), or would moving away from reinterpret cast and into, say.. memcpy work?

```
    std::memcpy(
         iter, &nanoseconds, (buffer.size() - 1) * sizeof(unsigned char) + sizeof(std::array<unsigned char,1>)
    );
```

BrettRD:
All forms of pointer-based type-conversion magic cause undefined behaviour (including reinterpret_cast, memcpy, and Quake's fast inverse square root) 
Mask and Shift is the most the most straightforward way of getting consistent behaviour across all architectures

I've set up an example over on compiler explorer: https://godbolt.org/z/hdGh73395
At optimisation level -O2, two of the three functions will produce identical binaries
`foo_endian_unsafe` matches `foo_little_endian` on x86 and ARM
`foo_endian_unsafe` matches `foo_big_endian` on Power

<details><summary>example source</summary>
<p>

```
#include <stdint.h>
#include <stddef.h>
#include <string.h>

void foo_endian_unsafe(unsigned char* buffer, uint32_t payload) {
  memcpy(buffer, reinterpret_cast<const unsigned char*>(&payload), sizeof(payload));
}


void foo_little_endian(unsigned char* buffer, uint32_t payload) {
  buffer[0] = (unsigned char)((payload & 0x000000FF) >> 0);
  buffer[1] = (unsigned char)((payload & 0x0000FF00) >> 8);
  buffer[2] = (unsigned char)((payload & 0x00FF0000) >> 16);
  buffer[3] = (unsigned char)((payload & 0xFF000000) >> 24);
}


void foo_big_endian(unsigned char* buffer, uint32_t payload) {
  buffer[0] = (unsigned char)((payload & 0xFF000000) >> 24);
  buffer[1] = (unsigned char)((payload & 0x00FF0000) >> 16);
  buffer[2] = (unsigned char)((payload & 0x0000FF00) >> 8);
  buffer[3] = (unsigned char)((payload & 0x000000FF) >> 0);
}

```

</p>
</details>

methylDragon:
> All forms of pointer-based type-conversion magic cause undefined behaviour (including reinterpret_cast, memcpy, and Quake's fast inverse square root) Mask and Shift is the most the most straightforward way of getting consistent behaviour across all architectures
> 
> I've set up an example over on compiler explorer: https://godbolt.org/z/hdGh73395 At optimisation level -O2, two of the three functions will produce identical binaries `foo_endian_unsafe` matches `foo_little_endian` on x86 and ARM `foo_endian_unsafe` matches `foo_big_endian` on Power
> 
> example source
> ```
> #include <stdint.h>
> #include <stddef.h>
> #include <string.h>
> 
> void foo_endian_unsafe(unsigned char* buffer, uint32_t payload) {
>   memcpy(buffer, reinterpret_cast<const unsigned char*>(&payload), sizeof(payload));
> }
> 
> 
> void foo_little_endian(unsigned char* buffer, uint32_t payload) {
>   buffer[0] = (unsigned char)((payload & 0x000000FF) >> 0);
>   buffer[1] = (unsigned char)((payload & 0x0000FF00) >> 8);
>   buffer[2] = (unsigned char)((payload & 0x00FF0000) >> 16);
>   buffer[3] = (unsigned char)((payload & 0xFF000000) >> 24);
> }
> 
> 
> void foo_big_endian(unsigned char* buffer, uint32_t payload) {
>   buffer[0] = (unsigned char)((payload & 0xFF000000) >> 24);
>   buffer[1] = (unsigned char)((payload & 0x00FF0000) >> 16);
>   buffer[2] = (unsigned char)((payload & 0x0000FF00) >> 8);
>   buffer[3] = (unsigned char)((payload & 0x000000FF) >> 0);
> }
> ```

Ahh, thank you for this, this has been very insightful. I have not worked much with embedded systems, so considerations about packing bits are a little new to me.

I suppose I'll just adapt this for the `int64_t` we're working with and pick one of the packing orders.

So:
```
void foo_little_endian(unsigned char* buffer, int64_t payload) {
  buffer[0] = (unsigned char)((payload & 0x00000000'000000FF) >> 0);
  buffer[1] = (unsigned char)((payload & 0x00000000'0000FF00) >> 8);
  buffer[2] = (unsigned char)((payload & 0x00000000'00FF0000) >> 16);
  buffer[3] = (unsigned char)((payload & 0x00000000'FF000000) >> 24);
  buffer[4] = (unsigned char)((payload & 0x000000FF'00000000) >> 32);
  buffer[5] = (unsigned char)((payload & 0x0000FF00'00000000) >> 40);
  buffer[6] = (unsigned char)((payload & 0x00FF0000'00000000) >> 48);
  buffer[7] = (unsigned char)((payload & 0xFF000000'00000000) >> 56);
}
```

What I am confused a little bit about though, is, wouldn't the byte representation of the number be reversed depending on the endianness of the machine? Wouldn't that mean that, if we had an explicit byte packing order (in this case little endian), then wouldn't the byte buffer that eventually makes it to the uuid generation call differ? I feel like I'm missing something 🙇 

EDIT: I think I got it. I think I had a brain fart and forgot that the hex representation 
(that is, the hex **literal**) of the number is independent of the actual in-memory representation. The array should be the same regardless of system endianness. 🤦‍♂️

BrettRD:
Yep, endianness only affects types larger than 8b
UUID expects byte-array input, so it will produce consistent output based on array index order regardless of system arch.
unsafe casts like pointer casts will reverse the string order of an integer (and so change the hash output), depending on architecture

This is unlikely to pop up as an issue in the immediate future; very few people will need to synchronise a fuse_transaction across a mixed-arch robot fleet just yet, but it's a good habit to be in, especially with micro-ros gaining popularity.

You can run into a similar issue on a single machine too when doing pointer hacks over structs; certain compiler flags will include or exclude padding bytes between struct members for memory-alignment reasons, making the raw-memory representations of a struct incompatible between builds depending on build settings.

Explicit packing and parsing is almost always worth the effort, but you might like to back-log it until the CI testing comes online.

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,500 @@
...
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+install(DIRECTORY include/
+  DESTINATION include/${PROJECT_NAME}
 )
 
-#############
-## Testing ##
-#############
+ament_export_targets(${PROJECT_NAME}-export HAS_LIBRARY_TARGET)
+ament_export_dependencies(${ament_dependencies})
```
sloretz: It looks like this variable is undefined (context: https://github.com/locusrobotics/fuse/pull/281#discussion_r991624341 )

methylDragon:
Explicit byte packing for uuids are in (in the time PR):
 72f51e8290ab8a564d4ece417fd9fb7b54aa7868

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,500 @@
...
-    ${PROJECT_NAME}
-  CATKIN_DEPENDS
-    ${build_depends}
-  DEPENDS
-    Boost
-    CERES
-    EIGEN3
-    GLOG
-)
+find_package(glog REQUIRED)  # Used downstream
```
sloretz: What does downstream mean? Is it not a dependency of this package?

fuse_core/include/fuse_core/console.h
sloretz: I could see throttling on system time making sense. A downside is fewer messages logged when simulating at faster than real time.

fuse_core/src/fuse_echo.cpp
```diff
@@ -31,64 +31,75 @@
...
-    std::cout << "received at: " << ros::Time::now() << std::endl;
+    std::cout << "received at: " << this->now().seconds() << std::endl;
     auto transaction = transaction_deserializer_.deserialize(msg);
     transaction.print();
   }
 };
 
+} // namespace fuse_core
+
 int main(int argc, char **argv)
```
sloretz: I think `int main()` has to be in another file to use this as a component - the component manager already has a `main` fuction. It actually can be dropped completely because [rclcpp_components_register_node](https://github.com/ros2/rclcpp/blob/b9b1468d15c7ddc697c079e6934d54f183294280/rclcpp_components/cmake/rclcpp_components_register_node.cmake#L30) will create a standalone executable for a component.

fuse_core/CMakeLists.txt
methylDragon: It's a dependency of `fuse_constraints`, downstream from `fuse_core`. Core itself doesn't use glog, from what I can tell.

My understanding is that users can also define their own constraints (where they are likely to use glog), so finding glog here means that they won't have to explicitly find it.

sloretz: Ah. I would recommend against depending on something higher in the chain for convenience. It can unnecessarily increase the number of things that need to be installed. For example, say it was a build-only dependency downstream. Downstream could instead depend on it in their build process without requiring further downstream packages to install it.

In this specific case I see `glog/logging.h` is included in 3 `.cpp` files in `fuse_constraints`, but I don't see any `glog` API's used. Maybe it can be removed altogether.

methylDragon: I checked for glog API usage too!
Unintuitively, the stuff being used from the `glog` library are actually the `CHECK_XX` macros. :grimacing: 

e.g. 

```
  CHECK_GT(A_.rows(), 0);
  CHECK_EQ(A_.cols(), 3);
```

I'll move the find_package invocation to `fuse_constraints`

fuse_core/src/fuse_echo.cpp
methylDragon: Whoops!

Fixed: aa287f03c254134f577e887a6f9173a2d0b05619

fuse_core/CMakeLists.txt
methylDragon: Moved! aa287f03c254134f577e887a6f9173a2d0b05619

fuse_core/CMakeLists.txt
methylDragon: Oops, fixed: aa287f03c254134f577e887a6f9173a2d0b05619

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,513 @@
...
-    CXX_STANDARD_REQUIRED YES
-)
+## fuse_echo executable
+add_library(fuse_echo_component SHARED src/fuse_echo.cpp)
+target_link_libraries(fuse_echo_component PUBLIC ${PROJECT_NAME})
+target_link_libraries(fuse_echo_component PRIVATE rclcpp_components::component)
+
+rclcpp_components_register_node(fuse_echo_component
+  PLUGIN "fuse_core::FuseEcho"
+  EXECUTABLE fuse_echo_node
```
sloretz: nit, I'd recommend keeping the executable name as `fuse_echo`.

fuse_core/CMakeLists.txt
```diff
@@ -1,519 +1,513 @@
...
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
 )
 
 install(
-  TARGETS fuse_echo
-  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+  TARGETS fuse_echo_component
+  DESTINATION lib/${PROJECT_NAME}
```
sloretz: `fuse_echo_component` is a library, so it should have these destinations:

```CMake
install(TARGETS fuse_echo_component
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
  RUNTIME DESTINATION bin)
```

The original rule was installing the executable `fuse_echo`, but the `rclcpp_components_register_node` macro [already does that for us](https://github.com/ros2/rclcpp/blob/b9b1468d15c7ddc697c079e6934d54f183294280/rclcpp_components/cmake/rclcpp_components_register_node.cmake#L77-L79).

fuse_core/include/fuse_core/throttled_callback.h
sloretz: > I'm inclined to keep the flag and leave the boolean ROS parameter alone so there's parameter parity between the ROS 1 and ported version of Fuse.

I agree about keeping the parameter `throttle_use_wall_time` the same.

A clock will need to be passed in to use anything other than wall time because a ROS clock needs to be [attached to a time source](https://github.com/ros2/rclcpp/blob/b9b1468d15c7ddc697c079e6934d54f183294280/rclcpp/src/rclcpp/time_source.cpp#L92).  It can't be created inside of `ThrottledCallback()`. I'm not sure how to plumb that [here](https://github.com/locusrobotics/fuse/blob/07676a8a47ad00b605153d6d98aa58dce663a39e/fuse_models/src/twist_2d.cpp#L55) though. Maybe `fuse_core::AsyncSensorModel` can have a way for subclasses to get a ROS clock off of it?

methylDragon: That's a good point, I'll be adding node interfaces and node interface getters to the classes that are currently storing node handles. I'll probably use them to obtain the NodeClockInterface and then get the clock that way :thinking: 

I'll defer this and add a TODO pending that change (which should be the next thing to change for fuse_core anyway!)

sloretz (approved):
Looks good as a base for follow up PRs

fuse_core/CMakeLists.txt
svwilliams: If memory serves, the glog dependency was added to make Ceres Solver happy. Cmake support for Ceres had issues in Ubuntu Xenial. I'm actually a little surprised glogging is used directly within fuse at all. Those CHECK_XX macros can probably be switched out for asserts, and glogging dropped as direct dependency.

PR #283, opened 2022-10-14
fuse -> ROS 2 : Port Time

methylDragon:
See: #276

# Description
This PR ports time constructs (e.g. ros::Time, ros::Duration, timers, etc.) to ROS2 across the entire stack. I also "downgraded" to `std::chrono` when it made sense to.

~~It also introduces a new TimeStamp class (and uses it in most places) to allow for interoperability with std::chrono time points and rclcpp::Time (and to also introduce the explicit concept of uninitialized time (as distinguished from zero-initialized time)). The class is a wrapper around rclcpp::Time.~~ Edit: This is now obsoleted.

# PR Layout
I recommend looking at the individual commits when reviewing. I tried to have a commit for each type of construct.

# Notes
Note that the use of rclcpp::Time as the base class means that comparisons and arithmetic between different timestamps will need to have the same clock type. They'd otherwise result in runtime errors (which is a good thing.) This means that a user-written timestamp will need to be populated with the correct clock type. I've tried to do this for the stack, but we'll only find out once we can get the stack building and tests running. Note that this will be something that downstream users implementing their own plugins have to be mindful of.

Also, some changes here require changes in how nodes are getting passed around, so expect those in a future PR. I left some comments to remind myself to port them.


fuse_core/src/time.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+// UTILITIES =======================================================================================
+std::ostream& operator<<(std::ostream& os, const fuse_core::TimeStamp& timestamp)
+{
+    os << timestamp.nanoseconds();
+    return os;
+}
+
+
+bool isValid(rclcpp::Clock::SharedPtr clock)
```
sloretz: I'd recommend upstreaming this to the `rclcpp::Time` class.

fuse_core/src/time.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <fuse_core/time.h>
+#include <rclcpp/utilities.hpp>
+
+using chrono_ns_time_point_t =
+  std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds>;
+
+namespace fuse_core
+{
+TimeStamp::TimeStamp() :
+  rclcpp::Time(static_cast<int64_t>(0), RCL_CLOCK_UNINITIALIZED)
```
sloretz: This seems like something we could upstream, maybe as a static class member `rclcpp::Time::Zero`.

fuse_core/src/time.cpp
```diff
@@ -0,0 +1,136 @@
...
+}
+
+
+bool isValid(rclcpp::Clock::SharedPtr clock)
+{
+  switch (clock->get_clock_type()) {
+    case RCL_CLOCK_UNINITIALIZED:
+      return false;
+    case RCL_SYSTEM_TIME:
+      return true;
```
sloretz: The [design doc says](http://design.ros2.org/articles/clock_and_time.html) `A time value of zero should be considered an error meaning that time is uninitialized`, but it doesn't make it clear if that's for all clock types or just the ROS clock type.

`tf2` [considers zero on the system clock to be uninitialized](https://github.com/ros2/geometry2/blob/de08d177d424f0d73efd4ff5951fd1eaabc9638d/tf2/include/tf2/time.h#L45), so I'd recommend keeping that expectation here.

fuse_core/src/time.cpp
```diff
@@ -0,0 +1,136 @@
...
+  }
+}
+
+
+// Logic adapted from: http://docs.ros.org/en/noetic/api/rostime/html/src_2time_8cpp_source.html
+bool waitForValid(rclcpp::Clock::SharedPtr clock, rclcpp::Duration timeout)
+{
+   rclcpp::Time start = clock->now();
+   while (clock->now().nanoseconds() == 0 && rclcpp::ok()) {
+     rclcpp::sleep_for(std::chrono::nanoseconds(timeout.nanoseconds()));
```
sloretz: This looks like it will unconditionally sleep for the full timeout period. If a timeout is required and we can't use `clock->sleep_until()`, I'd recommend implementing this with a jump handler on the clock itself. That way this method will return as soon as time is valid.

fuse_core/src/time.cpp
```diff
@@ -0,0 +1,136 @@
...
+    case RCL_ROS_TIME:
+      return clock->now().nanoseconds() == 0;
+    default:
+      return false;
+  }
+}
+
+
+// Logic adapted from: http://docs.ros.org/en/noetic/api/rostime/html/src_2time_8cpp_source.html
+bool waitForValid(rclcpp::Clock::SharedPtr clock, rclcpp::Duration timeout)
```
sloretz: I'd recommend making this a method on the `rclcpp::Clock` object since it seems to be an API we're missing in ROS 2.

One way to implement it would be using `rclcpp::Clock::sleep_until()` the smallest valid time: `clock->sleep_until(rclcpp::Time(1, clock.get_clock_type(), context));`. The upside is it's simple and reuses some trick logic. The downside is it doesn't allow for a timeout. Not having a timeout seems like it would be fine though. I see it's used in tests here as an assertion that Time is initialized. In the unlikely case time is never initialized, the impact is minimal assuming the test-runner has a timeout for the whole test.

fuse_core/src/time.cpp
methylDragon: Regarding the design doc, I think that's concerned with **time abstractions** (i.e. the analog to the std::chrono::time_point<T>) as opposed to clocks, whereas these methods are very much related to clocks.

That said, if the method name is the same as from ROS 1, should we follow the logic from ROS 1 instead?
(Valid unless sim time is being used and is 0).
```
   bool Time::isValid()
   {
     return (!g_use_sim_time) || !g_sim_time.isZero();
   }
```

The isValid there checks for only time zero on simulation time (which I'd interpret as RCL_ROS_TIME and 0 (because it would be quite a roundabout route to trying to pull up the `use_sim_time` param..))

Edit: I just verified in source that `[ros_time_is_active()](https://docs.ros2.org/bouncy/api/rclcpp/classrclcpp_1_1_clock.html#a3bebc635619797c58c2da5dc9b6032a8) ()` gets set to true if an RCL_ROS_TIME source is using the time from the `/clock` topic.

fuse_core/src/time.cpp
methylDragon: I'll go prep a PR to rclcpp, but until that gets merged and released, let's get this in?
I'll add a TODO to use it when it gets upstreamed.

fuse_core/src/time.cpp
methylDragon: I didn't realize clock->sleep_until() was available :o
I'll use that instead (well, more precisely the convenience `clock->sleep_for()`.. which I suppose uses a jump handler anyway...

fuse_core/src/time.cpp
BrettRD: I thought time zero was explicitly removed from the ROS1 API?
It certainly makes little sense for ROS2; nodes have no fixed datum and there's no guarantee they won't cross zero at runtime. (I'm dealing with related issues in a GStreamer package)

I only built time zero into fuse timestamp to make it easier to remove it from the fuse_publisher logic, I intended to delete it during cleanup
`rclcpp::Time::Zero` could do more harm than good if it gets up-streamed.

fuse_core/src/time.cpp
BrettRD: My memory of ros1 time is pretty hazy, but I thought it ros1 started time at zero on all nodes when roscore started

ROS2 breaks some convenient assumptions by allowing nodes to start at any time, and typically wrapping various system clocks directly. if a wrapped clock produces a number, you can only assume that it's valid.

I think this switch statement is as sane as it gets

fuse_core/src/time.cpp
sloretz: > It certainly makes little sense for ROS2; nodes have no fixed datum and there's no guarantee they won't cross zero at runtime. (I'm dealing with related issues in a GStreamer package)

I don't think crossing zero is a concern for ROS 2. At the lowest level [`rcutils` forbids negative time from the system and steady clocks](https://github.com/ros2/rcutils/blob/98d624a455e52c8bdd150c380ca302cf6a8a750c/src/time_unix.c#L69-L72). The [`rclcpp::Time` class tries to forbid negative time too](https://github.com/ros2/rclcpp/blob/edbfe1404b24d0bc85ff88e8ff1f006670788e46/rclcpp/src/rclcpp/time.cpp#L52-L54), but I see at least one bug there that I'll go fix.

fuse_core/src/time.cpp
sloretz: ~~Given that rcutils and rclcpp forbid negative time regardless of clock type https://github.com/locusrobotics/fuse/pull/283/files#r1010608091 , I think `IsValid()` doesn't need to consider it. `IsValid` only needs to check time is greater than zero~~. **EDIT** wait, this doesn't forbid 0 with system and stead clocks. I'll bring this up in today's ROS 2 meeting.

fuse_core/src/time.cpp
methylDragon: Duplicated the implementation in https://github.com/ros2/rclcpp/pull/2040 and used them https://github.com/locusrobotics/fuse/pull/283/commits/96b055546697e9c28ed6208383c4379c08fe2367

fuse_core/src/time.cpp
methylDragon: Duplicated the implementation in https://github.com/ros2/rclcpp/pull/2040 and used them https://github.com/locusrobotics/fuse/pull/283/commits/96b055546697e9c28ed6208383c4379c08fe2367

fuse_core/src/time.cpp
methylDragon: Codified 0 time as invalid, duplicated the implementation in https://github.com/ros2/rclcpp/pull/2040 and used them https://github.com/locusrobotics/fuse/pull/283/commits/96b055546697e9c28ed6208383c4379c08fe2367

fuse_core/src/time.cpp
methylDragon: I didn't upstream this specifically because I didn't know what clock type to use as default. Well that and I don't think there should be a clock type that is used by default.

So making a Zero static class member with a particular clock type would bait a lot of users into getting runtime errors if they, for instance, used it for comparison.

fuse_core/src/time.cpp
methylDragon: Upstreamed, duplicated the implementation in https://github.com/ros2/rclcpp/pull/2040 and used them https://github.com/locusrobotics/fuse/pull/283/commits/96b055546697e9c28ed6208383c4379c08fe2367

fuse_core/include/fuse_core/time.h
```diff
@@ -0,0 +1,185 @@
...
+ * of the operator overloads and clock type tracking of the parent class. Including the consistency
+ * checks for clock types when doing time arithmetic.
+ *
+ * To support this, the internal representation of time is an rclcpp::Time object and conversion
+ * functions and constructors are written to create this object out of primitive, rclcpp::Time and
+ * std::chrono time objects.
+ *
+ * A significant deviation from rclcpp:Time is the introduction of the default constructor that
+ * explicitly sets the clock type to RCL_CLOCK_UNINITIALIZED. This is done to differentiate between
+ * zero-initializations of time, and an uninitialized time.
```
sloretz: Do we need `fuse_core::TimeStamp`? I don't think the `RCL_CLOCK_UNINITIALIZED` in the constructor is needed anymore since 0 on any clock type will be invalid (for better or worse).\

Any conversions to chrono types can be done in free functions, but I'm not sure converting to `std::chrono` types is a good idea. There's no guarantee that the clock used to get the ROS time is the same clock used by `std::chrono`. For example, for steady time `rcutils` is using [`CLOCK_MONOTONIC_RAW`](https://github.com/ros2/rcutils/blob/f69678ae6b9126895482b57b0cf6c47862385e03/src/time_unix.c#L90) while [gcc's chrono implentation is using `CLOCK_MONOTONIC`](https://github.com/gcc-mirror/gcc/blob/32a06ce38a38bf37db468f0e6c83520fcc221534/libstdc%2B%2B-v3/src/c%2B%2B11/chrono.cc#L85). `std::chrono::time_point<>` returned by both might give incorrect comparisons because [one is subject to NTP adjustments](https://linux.die.net/man/2/clock_gettime). The same could happen with system time depending on which C++ standard library implementation is used.

BrettRD: `fuse_core::TimeStamp` was mostly part of a plan to support backports to Fuse for ROS1
`std::chrono` isn't required, as long as there's enough feature parity that Fuse doesn't diverge substantially between ROS1 and ROS2 branches

I do worry about applications crashing when a (valid) time zero arrives

methylDragon: I think we're currently at a point where we can indeed use `rclcpp::Time` :thinking: 
The only other reason I can think of for wrapping the time class would be to add metadata, which isn't a concern at the moment.

For posterity though, I've saved the current state in a gist: https://gist.github.com/methylDragon/ba1a39f619d7f9ca22dfc9fd49e51b40

As for crashing when time zero arrives, the only time that could occur is if a system clock is reporting time 0 on its epoch (meaning it'd be in 1970, 1 Jan), or if we have time 0 in a simulation. The former is unlikely to happen, and the latter can be guarded against with [wait_for_valid()](https://github.com/methylDragon/fuse/blob/rolling-time/fuse_optimizers/src/optimizer.cpp#L108), so I tihnk it's fairly safe. In either case, those timestamps would be considered invalid.

I'll also update all Time instantiations of (0, 0) to (0, 1), to reflect the time validity. Change is pending.

methylDragon:
Note to self: I'm going to need to:
- **Eventually**: make sure this all runs by syncing up the appropriate clock types (no mismatches between RCL_ROS_TIME and RCL_SYSTEM_TIME should occur)
- Keep Time(0, 0) for when time is truly meant to be uninitialized
- Migrate to Time(0, 1) for when it's meant to signify "earliest valid time" (e.g. for comparisons)
	- What happens if Time(0, 1) gets compared to Time(0, 0)???? D:

fuse_core/include/fuse_core/time.h
BrettRD: Again, the point of `fuse_core::TimeStamp` was for ROS1 backports, and I still think that's an important usage for a wrapper class, even if it's just a typedef.
Time is currently the only ROS dependency in user-written `fuse_variables`. If you wrap time, migration of variable definitions between versions of ROS becomes trivial

I also want to emphatically dispute "unlikely" because ROS2 uses signed 32 bit integers on the system epoch, which overflows to negative values and stays there in a little less than 16 years from now.
This will never be a problem in ROS1 because nobody will accumulate 60 years of up-time on any ROS1 roscore process
The embedded systems world has conventions where we use overflow-tolerant arithmetic, but it strictly requires that the actual time values are never used as logical flags.

sloretz: > Again, the point of fuse_core::TimeStamp was for ROS1 backports, and I still think that's an important usage for a wrapper class, even if it's just a typedef.
Time is currently the only ROS dependency in user-written fuse_variables. If you wrap time, migration of variable definitions between versions of ROS becomes trivial

The typedef will need to be done to both ROS 1 and ROS 2 branches to be useful. Supporting ROS 1 is out of scope for our effort, but I don't see a reason against doing so as an enhancement to both branches later. In the meantime user-written `fuse_variables` can support both ROS versions with `#ifdef RCLCPP_VERSION_MAJOR` and `typedef`. That's not convenient, but I would guess downstream code probably already has some ROS-awareness at the build system level because of catkin.

> I also want to emphatically dispute "unlikely" because ROS2 uses signed 32 bit integers on the system epoch, which overflows to negative values and stays there in a little less than 16 years from now.

This is a larger issue than we can solve in the `fuse` repo. ROS 2 will probably follow whatever OMG DDS does https://issues.omg.org/issues/DDS15-13

methylDragon:
I've tried to make a best effort decision for when to use (0, 0) (uninitialized time) and (0, 1) (earliest time) from reading the code.
This is something to test when the whole stack eventually builds.

https://github.com/locusrobotics/fuse/pull/283/commits/cfd4d2164843f3b4670973cfceb6190cb3cc030b

It's complicated by the fact that uninitialized time is sometimes used (just like in tf), to grab the latest time. And other times used to signify non-initialization, or used for comparison. For most cases it's pretty clear, but there are some cases (e.g. for comparison for deletion) where I'm unsure if it would be favorable to allow (0, 0) to be deleted or not.

In those cases I erred on yes.

methylDragon:
Also in https://github.com/locusrobotics/fuse/pull/283/commits/cfd4d2164843f3b4670973cfceb6190cb3cc030b#diff-33b6e1985c2434b43effda68029aab772a2cf3efd8f7d74c91bdfa9c89dee37dR235, since we moved back to using rclcpp::Time, I needed to decide which clock type to use for the publisher synchronizer's internal time (for its starting time and subsequent time.)

I went with RCL_ROS_TIME since that made sense. But there is nothing stopping a user from dropping in a transaction into the graph with a different clock type.

fuse_core/include/fuse_core/message_buffer_impl.h
```diff
@@ -140,8 +140,14 @@ void MessageBuffer<Message>::purgeHistory()
 
   // Compute the expiration time carefully, as ROS can't handle negative times
   const auto& ending_stamp = buffer_.back().first;
-  auto expiration_time =
-      ending_stamp.toSec() > buffer_length_.toSec() ? ending_stamp - buffer_length_ : rclcpp::Time(0, 0);
+
+  if (ending_stamp.seconds() > buffer_length.seconds()) {
```
sloretz: `buffer_length` -> `buffer_length_`

fuse_core/include/fuse_core/time.h
```diff
@@ -0,0 +1,111 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+
+/*
+ * This file provides a sane configuration for a clock to manage order of events.
+ * This file exists to help with a migration from rclcpp::Time to rclcpp::time
+ * Priority is given to C++ std::chrono
+ * emphasis will be placed on interoperability with rclcpp::Time and rclcpp::time
```
sloretz: Is this comment still accurate? It seems like a left-over from the `fuse_core::Time` class.

fuse_core/test/test_timestamp_manager.cpp
```diff
@@ -60,33 +60,33 @@ class TimestampManagerTestFixture : public ::testing::Test
...
+    const rclcpp::Time& ending_stamp,
     std::vector<fuse_core::Constraint::SharedPtr>& /*constraints*/,
     std::vector<fuse_core::Variable::SharedPtr>& /*variables*/)
   {
     generated_time_spans.emplace_back(beginning_stamp, ending_stamp);
   }
 
-  fuse_core::TimestampManager manager;
-  std::vector<std::pair<ros::Time, ros::Time> > generated_time_spans;
+  rclcpp::TimeManager manager;
```
sloretz: `rclcpp::TimeManager` -> `fuse_core::TimestampManager`? Seems like an accidental find/replace.

fuse_graphs/src/hash_graph.cpp
```diff
@@ -426,18 +426,18 @@ ceres::Solver::Summary HashGraph::optimize(const ceres::Solver::Options& options
 }
 
 ceres::Solver::Summary HashGraph::optimizeFor(
-  const ros::Duration& max_optimization_time,
+  const std::chrono::nanoseconds& max_optimization_time,
```
sloretz: why `std::chrono` instead of `rclcpp::Time` here?

fuse_models/include/fuse_models/parameters/acceleration_2d_params.h
```diff
@@ -87,8 +87,8 @@ struct Acceleration2DParams : public ParameterBase
                                   //!< whatever the packet size. This reduces delay at the cost of network congestion,
                                   //!< specially if the payload of a packet is smaller than the TCP header data. This is
                                   //!< true for small ROS messages like geometry_msgs::AccelWithCovarianceStamped
-    ros::Duration tf_timeout { 0.0 };  //!< The maximum time to wait for a transform to become available
-    ros::Duration throttle_period { 0.0 };  //!< The throttle period duration in seconds
+    rclcpp::Duration tf_timeout { RCUTILS_S_TO_NS(0.0) };  //!< The maximum time to wait for a transform to become available
```
sloretz: nit, `RCUTILS_S_TO_NS` isn't needed here. `0` will be the same whether seconds or nanoseconds.

fuse_models/include/fuse_models/parameters/imu_2d_params.h
```diff
@@ -119,8 +119,8 @@ struct Imu2DParams : public ParameterBase
                                   //!< whatever the packet size. This reduces delay at the cost of network congestion,
                                   //!< specially if the payload of a packet is smaller than the TCP header data. This is
                                   //!< true for small ROS messages like geometry_msgs::AccelWithCovarianceStamped
-    ros::Duration tf_timeout { 0.0 };  //!< The maximum time to wait for a transform to become available
-    ros::Duration throttle_period { 0.0 };  //!< The throttle period duration in seconds
+    rclcpp::Duration tf_timeout { RCUTILS_S_TO_NS(0.0) };  //!< The maximum time to wait for a transform to become available
```
sloretz: Same comment about `RCUTILS_S_TO_NS` not being needed.

fuse_models/include/fuse_models/parameters/odometry_2d_params.h
```diff
@@ -117,8 +117,8 @@ struct Odometry2DParams : public ParameterBase
                                   //!< whatever the packet size. This reduces delay at the cost of network congestion,
                                   //!< specially if the payload of a packet is smaller than the TCP header data. This is
                                   //!< true for small ROS messages like geometry_msgs::AccelWithCovarianceStamped
-    ros::Duration tf_timeout { 0.0 };  //!< The maximum time to wait for a transform to become available
-    ros::Duration throttle_period { 0.0 };  //!< The throttle period duration in seconds
+    rclcpp::Duration tf_timeout { RCUTILS_S_TO_NS(0.0) };  //!< The maximum time to wait for a transform to become available
```
sloretz: Same comment about `RCUTILS_S_TO_NS` not being needed.

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
```diff
@@ -131,10 +131,10 @@ struct Odometry2DPublisherParams : public ParameterBase
   fuse_core::Matrix8d process_noise_covariance;   //!< Process noise covariance matrix
   bool scale_process_noise{ false };
   double velocity_norm_min{ 1e-3 };
-  ros::Duration covariance_throttle_period { 0.0 };  //!< The throttle period duration in seconds to compute the
+  rclcpp::Duration covariance_throttle_period { RCUTILS_S_TO_NS(0.0) };  //!< The throttle period duration in seconds to compute the
                                                      //!< covariance
-  ros::Duration tf_cache_time { 10.0 };
-  ros::Duration tf_timeout { 0.1 };
+  rclcpp::Duration tf_cache_time { RCUTILS_S_TO_NS(10.0) };
```
sloretz: This is fine, but in the future `rclcpp::Duration` also has a constructor taking two arguments: seconds and nanoseconds. I think this would work too.

```c++
rclcpp::Duration tf_cache_time{10, 0};
```

fuse_models/include/fuse_models/parameters/pose_2d_params.h
```diff
@@ -104,8 +104,8 @@ struct Pose2DParams : public ParameterBase
                                   //!< whatever the packet size. This reduces delay at the cost of network congestion,
                                   //!< specially if the payload of a packet is smaller than the TCP header data. This is
                                   //!< true for small ROS messages like geometry_msgs::AccelWithCovarianceStamped
-    ros::Duration tf_timeout { 0.0 };  //!< The maximum time to wait for a transform to become available
-    ros::Duration throttle_period { 0.0 };  //!< The throttle period duration in seconds
+    rclcpp::Duration tf_timeout { RCUTILS_S_TO_NS(0.0) };  //!< The maximum time to wait for a transform to become available
```
sloretz: Same comment about `RCUTILS_S_TO_NS` not being necessary here.

fuse_models/include/fuse_models/parameters/twist_2d_params.h
```diff
@@ -90,8 +90,8 @@ struct Twist2DParams : public ParameterBase
                                   //!< whatever the packet size. This reduces delay at the cost of network congestion,
                                   //!< specially if the payload of a packet is smaller than the TCP header data. This is
                                   //!< true for small ROS messages like geometry_msgs::AccelWithCovarianceStamped
-    ros::Duration tf_timeout { 0.0 };  //!< The maximum time to wait for a transform to become available
-    ros::Duration throttle_period { 0.0 };  //!< The throttle period duration in seconds
+    rclcpp::Duration tf_timeout { RCUTILS_S_TO_NS(0.0) };  //!< The maximum time to wait for a transform to become available
```
sloretz: Same comment about RCUTILS_S_TO_NS not being necessary here.

fuse_models/include/fuse_models/unicycle_2d.h
```diff
@@ -183,9 +183,9 @@ class Unicycle2D : public fuse_core::AsyncMotionModel
   static void validateMotionModel(const StateHistoryElement& state1, const StateHistoryElement& state2,
                                   const fuse_core::Matrix8d& process_noise_covariance);
 
-  ros::Duration buffer_length_;                    //!< The length of the state history
+  rclcpp::Duration buffer_length_;                    //!< The length of the state history
   fuse_core::UUID device_id_;                      //!< The UUID of the device to be published
-  fuse_core::TimestampManager timestamp_manager_;  //!< Tracks timestamps and previously created motion model segments
+  rclcpp::TimeManager timestamp_manager_;  //!< Tracks timestamps and previously created motion model segments
```
sloretz: `rclcpp::TimeManager` -> `fuse_core::TimestampManager`.

fuse_models/src/odometry_2d_publisher.cpp
```diff
@@ -104,7 +96,7 @@ void Odometry2DPublisher::notifyCallback(
 {
   // Find the most recent common timestamp
   const auto latest_stamp = synchronizer_.findLatestCommonStamp(*transaction, *graph);
-  if (latest_stamp == Synchronizer::TIME_ZERO)
+  if (latest_stamp == Synchronizer::ROS_TIME_ZERO)
```
sloretz: If `rclcpp::Time` had an `is_valid()` method it would be convenient here `if (latest_stamp.is_valid()) ...` so that `Synchronizer::ROS_TIME_ZERO` wasn't necessary.

fuse_optimizers/src/variable_stamp_index.cpp
```diff
@@ -58,7 +58,8 @@ ros::Time VariableStampIndex::currentStamp() const
   }
   else
   {
-    return ros::Time(0, 0);
+    // NOTE(CH3): Might be useful to figure out where to place uninitialized checks later on...
+    return rclcpp::Time(0, 0, default_clock_type);  // NOTE(CH3): Signals uninitialized
```
sloretz: `RCL_CLOCK_UNINITIALIZED` would make sense to me here if `rclcpp::Time::is_valid()` existed. Then there's no need for this parameter to get passed.

fuse_optimizers/src/batch_optimizer.cpp
```diff
@@ -206,12 +212,15 @@ void BatchOptimizer::transactionCallback(
...
+    rclcpp::Time purge_time(0, 0, transaction_clock_type);  // NOTE(CH3): Uninitialized
     if (started_)
     {
       purge_time = start_time_;
     }
-    else if (ros::Time(0, 0) + params_.transaction_timeout < last_pending_time)  // prevent a bad subtraction
+    // NOTE(CH3): In this case we're okay with uninitialized time since it's just used
+    //            for comparison
+    // prevent a bad subtraction
+    else if (rclcpp::Time(0, 0, transaction_clock_type) + params_.transaction_timeout < last_pending_time)
```
sloretz: I'm not sure why `rclcpp::Time(0,0)` is being added. Does `else if (params_.transaction_timeout < last_pending_time)` work?

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -190,6 +191,7 @@ void FixedLagSmoother::optimizationLoop()
       //         We do this to ensure state of the graph does not change between unlocking the pending_transactions
       //         queue and obtaining the lock for the graph. But we have now obtained two different locks. If we are
       //         not extremely careful, we could get a deadlock.
+      //  XXX make sure lag_expiration_ has been initialised
```
sloretz: `XXX` -> `TODO` for greping in the future?

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -279,13 +281,17 @@ void FixedLagSmoother::optimizerTimerCallback(const ros::TimerEvent& event)
     {
       std::lock_guard<std::mutex> lock(optimization_requested_mutex_);
       optimization_request_ = true;
-      optimization_deadline_ = event.current_expected + params_.optimization_period;
+      // TODO(CH3): ??????? No idea what this does...
+      // XXX event.current_expected refers to when ROS planned to execute the callback, not the current time.
+      //     wall-time and measurement-time should be decoupled further
+      // optimization_deadline_ = event.current_expected + params_.optimization_period;
+      optimization_deadline_ = get_clock()->now() + params_.optimization_period;
```
sloretz: The documentation says updates should happen at a fixed frequency.

https://github.com/locusrobotics/fuse/blob/1118d16856017a8b8430c0d24de66e1c993d5d43/fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother.h#L70-L7

I think using `now() + period` would shift when the next update will happen. Maybe `now()` + [`timer->time_until_trigger()`](https://github.com/ros2/rclcpp/blob/7c6785176a878cf16eb98995b732f405b9e776ed/rclcpp/include/rclcpp/timer.hpp#L121-L122) would have the same behavior as before?

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -523,9 +529,13 @@ void FixedLagSmoother::transactionCallback(
...
         // And purge out old transactions to limit the pending size while waiting for an ignition sensor
-        auto purge_time = ros::Time(0, 0);
+        auto purge_time = rclcpp::Time(0, 0, RCL_ROS_TIME);  // NOTE(CH3): Uninitialized
         auto last_pending_time = pending_transactions_.front().stamp();
-        if (ros::Time(0, 0) + params_.transaction_timeout < last_pending_time)  // ros::Time doesn't allow negatives
+
+        // rclcpp::Time doesn't allow negatives
+        // NOTE(CH3): In this case we're okay with uninitialized time since it's just used
+        //            for comparison
+        if (rclcpp::Time(0, 0, RCL_ROS_TIME) + params_.transaction_timeout < last_pending_time)
```
sloretz: Not sure what adding `0` time does here. Is it to convert a duration to a time for the comparison? If so, `rclcpp::Time(params_.transation_timeout.nanoseconds(0, last_pending_time.get_clock_type())` might be a better fit because it makes sure the clock types are comparable.

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -623,20 +633,23 @@ void FixedLagSmoother::setDiagnostics(diagnostic_updater::DiagnosticStatusWrappe
...
       const std::unique_lock<std::mutex> lock(optimization_requested_mutex_, std::try_to_lock);
       if (lock)
       {
         optimization_deadline = optimization_deadline_;
+        got_deadline_from_mutex = true;
       }
     }
 
-    if (!optimization_deadline.isZero())  // This is zero for the default-constructed optimization_deadline object
+    if (got_deadline_from_mutex)
```
sloretz: Seems like another spot where `rclcpp::Time::is_valid()` or mabye `rclcpp::Time::is_zero()` would be valuable.

fuse_publishers/include/fuse_publishers/stamped_variable_synchronizer.h
```diff
@@ -261,34 +261,44 @@ struct is_variable_in_pack<T, Ts...>
 
 }  // namespace detail
 
+// NOTE(CH3): ROS_TIME_ZERO should be uninitialized, it's used for getting latest time!
 template <typename ...Ts>
-const ros::Time StampedVariableSynchronizer<Ts...>::TIME_ZERO = ros::Time(0, 0);
+const rclcpp::Time StampedVariableSynchronizer<Ts...>::ROS_TIME_ZERO =
+  rclcpp::Time(0, 0, RCL_ROS_TIME);
```
sloretz: Seems like `rclcpp::Time::is_valid()` would eliminate the need for these constants since it would give downstream users a different way to check if time is valid or not.

fuse_publishers/src/pose_2d_publisher.cpp
```diff
@@ -190,7 +188,10 @@ void Pose2DPublisher::onStart()
   // Start the tf timer
   if (publish_to_tf_)
   {
-    tf_publish_timer_.start();
+    tf_publish_timer_ = node_.create_wall_timer(
```
sloretz: I'd recommend a ROS time timer by default, unless there's a specific need for system time. There's a [free function `rclcpp::create_timer()` to which a ROS time clock](https://github.com/ros2/rclcpp/blob/7c6785176a878cf16eb98995b732f405b9e776ed/rclcpp/include/rclcpp/create_timer.hpp#L107) can be passed.

fuse_publishers/src/pose_2d_publisher.cpp
```diff
@@ -199,7 +200,7 @@ void Pose2DPublisher::onStop()
   // Stop the tf timer
   if (publish_to_tf_)
   {
-    tf_publish_timer_.stop();
+    tf_publish_timer_.reset();
```
sloretz: I think `stop()` in ROS 1 translates to `cancel` in ROS 2 https://github.com/ros2/rclcpp/blob/7c6785176a878cf16eb98995b732f405b9e776ed/rclcpp/include/rclcpp/timer.hpp#L74

fuse_publishers/src/serialized_publisher.cpp
```diff
@@ -67,7 +67,7 @@ void SerializedPublisher::onInit()
   bool latch = false;
   private_node_handle_.getParam("latch", latch);
 
-  ros::Duration graph_throttle_period{ 0.0 };
+  rclcpp::Duration graph_throttle_period{ RCUTILS_S_TO_NS(0.0) };
```
sloretz: same comment about `RCUTILS_S_TO_NS` not being necessary here.

fuse_tutorials/src/range_sensor_simulator.cpp
```diff
@@ -117,10 +117,18 @@ std::vector<Beacon> createNoisyBeacons(const std::vector<Beacon>& beacons)
 /**
  * @brief Convert the set of beacons into a pointcloud for visualization purposes
  */
-sensor_msgs::PointCloud2::ConstPtr beaconsToPointcloud(const std::vector<Beacon>& beacons)
+sensor_msgs::PointCloud2::ConstPtr beaconsToPointcloud(
+  const std::vector<Beacon>& beacons,
+  rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface=nullptr)
```
sloretz: Since the `clock_interface` is used to get a temporary clock for one call, I'd recommend this method takes a `const rclcpp::Clock &` instead of the interface.

fuse_tutorials/src/range_sensor_simulator.cpp
```diff
@@ -190,7 +198,8 @@ nav_msgs::Odometry::ConstPtr robotToOdometry(const Robot& state)
  *
  * The state estimator will not run until it has been sent a starting pose.
  */
-void initializeStateEstimation(const Robot& state)
+void initializeStateEstimation(
+  const Robot& state, rclcpp::node_interfaces::NodeClockInterface::SharedPtr clock_interface=nullptr)
```
sloretz: Same comment about taking a clock instead of a clock interface here.

sloretz (changes requested):

fuse_graphs/src/hash_graph.cpp
methylDragon: Good catch! The Graph ABC and its implementations don't depend on ROS in any way, so I was hoping to keep them separate.

That said though, the other graph headers have been migrated to rclcpp::Duration, I'm thinking of changing them back to std::chrono :thinking: What do you think?

Edit: I changed them to std::chrono

fuse_optimizers/src/batch_optimizer.cpp
methylDragon: The timeout is a Duration, so the arithmetic is to get a Time object offset by the duration for comparison with last_pending_time, which is also a Time object.

fuse_optimizers/src/fixed_lag_smoother.cpp
methylDragon: Fair point, I'll update the one on top also

fuse_models/src/odometry_2d_publisher.cpp
methylDragon: Wouldn't it be more explicit that it's a comparison to Time(0, 0) here so readers of the code know that it's doing something like what TF2 is doing?

fuse_publishers/include/fuse_publishers/stamped_variable_synchronizer.h
methylDragon: From the earlier comment

> Wouldn't it be more explicit that it's a comparison to Time(0, 0) here so readers of the code know that it's doing something like what TF2 is doing?

I think we should keep it, but if we should remove it I can do that too!

fuse_tutorials/src/range_sensor_simulator.cpp
methylDragon: I'll take in an `rclcpp::Clock::SharedPtr`

fuse_publishers/src/pose_2d_publisher.cpp
methylDragon: I'll change the 4 instances to use the new [`create_timer`](https://docs.ros.org/en/rolling/p/rclcpp/generated/classrclcpp_1_1Node.html?highlight=node#_CPPv4I000EN6rclcpp4Node12create_timerEN6rclcpp12GenericTimerI9CallbackTE9SharedPtrENSt6chrono8durationI12DurationRepT9DurationTEE9CallbackTN6rclcpp13CallbackGroup9SharedPtrE) method

fuse_tutorials/src/range_sensor_simulator.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/59d10ddc448f2a21928d5104dabe494e0e37f8e4

fuse_tutorials/src/range_sensor_simulator.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/59d10ddc448f2a21928d5104dabe494e0e37f8e4

fuse_publishers/src/pose_2d_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/59d10ddc448f2a21928d5104dabe494e0e37f8e4

fuse_publishers/src/pose_2d_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/59d10ddc448f2a21928d5104dabe494e0e37f8e4

fuse_publishers/src/serialized_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9314b85ef282a2279c816c771a8d90f591f292d7

fuse_optimizers/src/fixed_lag_smoother.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9a1142173a60e2d50cb7bb40180b0ff85c8f486a

fuse_optimizers/src/fixed_lag_smoother.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/c16b02f15441790cdeea5bcd023940801ca89a01

fuse_optimizers/src/batch_optimizer.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/c16b02f15441790cdeea5bcd023940801ca89a01

fuse_optimizers/src/fixed_lag_smoother.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/59d10ddc448f2a21928d5104dabe494e0e37f8e4#diff-969711c8c46e1a4ced29c0acd11c4939bcfa965ef47be0d91d4a9449b0c5e079

fuse_optimizers/src/variable_stamp_index.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9a1142173a60e2d50cb7bb40180b0ff85c8f486a

fuse_models/include/fuse_models/unicycle_2d.h
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9314b85ef282a2279c816c771a8d90f591f292d7

fuse_models/include/fuse_models/parameters/twist_2d_params.h
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9314b85ef282a2279c816c771a8d90f591f292d7

fuse_models/include/fuse_models/parameters/pose_2d_params.h
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9314b85ef282a2279c816c771a8d90f591f292d7

fuse_models/include/fuse_models/parameters/odometry_2d_publisher_params.h
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9314b85ef282a2279c816c771a8d90f591f292d7

fuse_models/include/fuse_models/parameters/odometry_2d_params.h
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9314b85ef282a2279c816c771a8d90f591f292d7

fuse_models/include/fuse_models/parameters/imu_2d_params.h
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9314b85ef282a2279c816c771a8d90f591f292d7

fuse_models/include/fuse_models/parameters/acceleration_2d_params.h
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9314b85ef282a2279c816c771a8d90f591f292d7

fuse_graphs/src/hash_graph.cpp
methylDragon: Here: https://github.com/locusrobotics/fuse/pull/283/commits/9314b85ef282a2279c816c771a8d90f591f292d7#diff-d735274778d4bdbb320b2042641a7abe4f36e969e23bcfde0ab87d0f9df53474R382

fuse_core/test/test_timestamp_manager.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9314b85ef282a2279c816c771a8d90f591f292d7 my bad..

fuse_core/include/fuse_core/time.h
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/a2fe96edb0eade18241cae391ad5917bf3e4f765

fuse_core/include/fuse_core/message_buffer_impl.h
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/62296cf3f8903e65ee50aa8a89bb47337c208b85

fuse_optimizers/src/fixed_lag_smoother.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/59d10ddc448f2a21928d5104dabe494e0e37f8e4

fuse_graphs/src/hash_graph.cpp
sloretz: > The Graph ABC and its implementations don't depend on ROS in any way, so I was hoping to keep them separate.
> That said though, the other graph headers have been migrated to rclcpp::Duration, I'm thinking of changing them back to std::chrono thinking What do you think?

It looks like they use `ros::Duration` in ROS 1, so this is actually changing them to `std::chrono`. I'd recommend using ROS types throughout to avoid downstream users having to convert to/from `std::chrono` depending on the API in `fuse_core` they're using.

https://github.com/locusrobotics/fuse/blob/07676a8a47ad00b605153d6d98aa58dce663a39e/fuse_core/include/fuse_core/graph.h#L382

fuse_models/src/odometry_2d_publisher.cpp
sloretz: > Wouldn't it be more explicit that it's a comparison to Time(0, 0) here so readers of the code know that it's doing something like what TF2 is doing?

I'm not sure. Subjectively,`findLatestCommonStamp` returning an invalid `Time` object to indicate there is no common time stamp would seem more clear to me.

fuse_tutorials/src/range_sensor_simulator.cpp
sloretz: `rclcpp::Clock::SharedPtr` as a parameter signals [that the passed in clock is going to keep the Clock alive longer than just the function call](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-sharedptrparam-owner).

I see that the intent of making clock optional is so it can fallback to a system time clock; however, this function is in a `.cpp` file [and is not exposed in the header file](https://github.com/methylDragon/fuse/blob/rolling-time/fuse_tutorials/include/fuse_tutorials/range_sensor_model.h). All calls to it [do pass in a clock](https://github.com/methylDragon/fuse/blob/ad6056c7f890ec51e65664cef7ebad02a8c3bca3/fuse_tutorials/src/range_sensor_simulator.cpp#L368-L372), so the the fallback behavior of using system time isn't needed. I think [`const rclcpp::Clock &` is a better fit](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const).

fuse_tutorials/src/range_sensor_simulator.cpp
sloretz: Same comment as https://github.com/locusrobotics/fuse/pull/283#discussion_r1018249517

methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/c18049be8cda5cf7ee3dd4eac7f3e2f1da50578e

fuse_tutorials/src/range_sensor_simulator.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/c18049be8cda5cf7ee3dd4eac7f3e2f1da50578e

I actually wanted to ask about this one, does this mean that the `is_valid` and `wait_for_valid` methods [in my time patch](https://github.com/locusrobotics/fuse/pull/283/files#diff-5e9d38b7ea0ad9239343157fb2e49683467e170e6df1471fce6a0d2c1a97a798R62) should also take in const refs?

I was thinking to use shared_ptrs because the node interface returns a shared ptr to the node's clock, so it'd avoid needing to do a dereference, and also because, in my view, the node's clock would continue to exist.

fuse_publishers/include/fuse_publishers/stamped_variable_synchronizer.h
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/bdedc95997d9e5fe34614bdeace81f21fabb654f

fuse_models/src/odometry_2d_publisher.cpp
methylDragon: okay https://github.com/locusrobotics/fuse/pull/283/commits/bdedc95997d9e5fe34614bdeace81f21fabb654f

fuse_graphs/src/hash_graph.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/283/commits/9156f257216c8bed9287df8d23707db59779a674

fuse_tutorials/src/range_sensor_simulator.cpp
sloretz: > I actually wanted to ask about this one, does this mean that the is_valid and wait_for_valid methods [in my time patch](https://github.com/locusrobotics/fuse/pull/283/files#diff-5e9d38b7ea0ad9239343157fb2e49683467e170e6df1471fce6a0d2c1a97a798R62) should also take in const refs?

Probably, yeah :no_mouth: 

> I was thinking to use shared_ptrs because the node interface returns a shared ptr to the node's clock, so it'd avoid needing to do a dereference, and also because, in my view, the node's clock would continue to exist.

Fair point, a `Clock::SharedPtr` is more convenient given other APIs return that.

> and also because, in my view, the node's clock would continue to exist.

From the perspective of `beaconsToPointcloud()`, the clock only has to exist while the it's being run.

fuse_graphs/src/hash_graph.cpp
```diff
@@ -426,21 +427,26 @@ ceres::Solver::Summary HashGraph::optimize(const ceres::Solver::Options& options
 }
 
 ceres::Solver::Summary HashGraph::optimizeFor(
-  const ros::Duration& max_optimization_time,
+  const rclcpp::Duration& max_optimization_time,
   const ceres::Solver::Options& options)
 {
-  auto start = ros::Time::now();
+  auto clock = rclcpp::Clock::Clock(RCL_STEADY_TIME);
```
sloretz: I'd recommend passing a clock so that a caller may use a ROS_TIME clock if they want.

fuse_tutorials/src/range_sensor_simulator.cpp
```diff
@@ -190,7 +193,8 @@ nav_msgs::Odometry::ConstPtr robotToOdometry(const Robot& state)
  *
  * The state estimator will not run until it has been sent a starting pose.
  */
-void initializeStateEstimation(const Robot& state)
+void initializeStateEstimation(
+  const Robot& state, rclcpp::Clock::SharedPtr clock=nullptr)
```
sloretz: Since this is an internal function, same comment as `beaconsToPointcloud` about unconditionally taking a clock. I see that where it's used [it's called without a clock](https://github.com/locusrobotics/fuse/blob/c18049be8cda5cf7ee3dd4eac7f3e2f1da50578e/fuse_tutorials/src/range_sensor_simulator.cpp#L380), but it could be given `*node->get_clock()` there.

sloretz (approved):

PR #284, opened 2022-11-07
fuse -> ROS 2 fuse_core : Nodes and Waitables

methylDragon:
This depends on: https://github.com/locusrobotics/fuse/pull/283
- When merging this, @methylDragon should check that it doesn't lead to Time regressions...

# Description
This PR ports the ROS 1 nodes to ROS 2 node pointers, and also uses Brett's solution for CallbackWrappers and Waitables to support async behavior.

Whether it works or not remains to be seen, we'll likely need to circle back if we see deadlocks happening.

Some node handles have been replaced with node pointers or node interfaces. A future PR will introduce the concept of a [NodeInterfaceHandle](https://github.com/ros2/rclcpp/pull/2041) and change the signatures to support it.

# PR Layout
Because I was cherry-picking Brett's commits here, I'd recommend reviewing the PR by looking at the merged changes instead of per commit.

# Additional Notes
All subordinate plugins (e.g. sensors, motion models, etc.) now start an internal node using the default global context. We need to first make sure that configuration works first before we try to improve it, so it's a place that's likely to be circled back to.. Hopefully...

### DIFF
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -29,21 +29,22 @@ include(boost-extras.cmake)
 
 ## fuse_core library
 add_library(${PROJECT_NAME} SHARED
-#    src/async_motion_model.cpp
-#    src/async_publisher.cpp
-#    src/async_sensor_model.cpp
-#    src/ceres_options.cpp
-    src/constraint.cpp
-    src/graph.cpp
-    src/graph_deserializer.cpp
-    src/loss.cpp
-    src/serialization.cpp
-    src/time.cpp
-    src/timestamp_manager.cpp
-    src/transaction.cpp
-    src/transaction_deserializer.cpp
-    src/uuid.cpp
-    src/variable.cpp
+  src/async_motion_model.cpp
+  src/async_publisher.cpp
+  src/async_sensor_model.cpp
+  src/callback_wrapper.cpp
+  # src/ceres_options.cpp
+  src/constraint.cpp
+  src/graph.cpp
+  src/graph_deserializer.cpp
+  src/loss.cpp
+  src/serialization.cpp
+  src/time.cpp
+  src/timestamp_manager.cpp
+  src/transaction.cpp
+  src/transaction_deserializer.cpp
+  src/uuid.cpp
+  src/variable.cpp
 )
 target_include_directories(${PROJECT_NAME} PUBLIC
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -82,8 +82,58 @@ if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
-  # find_package(ament_cmake_gtest REQUIRED)
-  # add_subdirectory(test)
+  ament_add_gtest(test_constraint
+    test/test_constraint.cpp)
+  target_link_libraries(test_constraint ${PROJECT_NAME})
+  target_include_directories(test_constraint PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+  ament_add_gtest(test_eigen
+    test/test_eigen.cpp)
+  target_link_libraries(test_eigen ${PROJECT_NAME})
+  target_include_directories(test_eigen PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+  # TODO(sloretz) fails to build: can't include ros/node_handle.h
+  # ament_add_gtest(test_local_parameterization
+  #   test/test_local_parameterization.cpp)
+  # target_link_libraries(test_local_parameterization ${PROJECT_NAME})
+  # target_include_directories(test_local_parameterization PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+  ament_add_gtest(test_loss
+    test/test_loss.cpp)
+  target_link_libraries(test_loss ${PROJECT_NAME})
+  target_include_directories(test_loss PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+  # TODO(sloretz) fails to build: invalid use of rclcpp::Time::get_clock_type()
+  # ament_add_gtest(test_message_buffer
+  #   test/test_message_buffer.cpp)
+  # target_link_libraries(test_message_buffer ${PROJECT_NAME})
+  # target_include_directories(test_message_buffer PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+  ament_add_gtest(test_timestamp_manager
+    test/test_timestamp_manager.cpp)
+  target_link_libraries(test_timestamp_manager ${PROJECT_NAME})
+  target_include_directories(test_timestamp_manager PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+  ament_add_gtest(test_transaction
+    test/test_transaction.cpp)
+  target_link_libraries(test_transaction ${PROJECT_NAME})
+  target_include_directories(test_transaction PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+  # TODO(sloretz) fails to build: can't include ros/node_handle.h
+  # ament_add_gtest(test_util
+  #   test/test_util.cpp)
+  # target_link_libraries(test_util ${PROJECT_NAME})
+  # target_include_directories(test_util PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+  ament_add_gtest(test_uuid
+    test/test_uuid.cpp)
+  target_link_libraries(test_uuid ${PROJECT_NAME})
+  target_include_directories(test_uuid PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+  ament_add_gtest(test_variable
+    test/test_variable.cpp)
+  target_link_libraries(test_variable ${PROJECT_NAME})
+  target_include_directories(test_variable PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
 endif()
 
 # TODO(CH3): Move this to the test directory and port to ROS 2
@@ -199,173 +249,6 @@ endif()
 #       CXX_STANDARD_REQUIRED YES
 #   )
 
-#   # Constraint tests
-#   catkin_add_gtest(test_constraint
-#     test/test_constraint.cpp
-#   )
-#   add_dependencies(test_constraint
-#     ${catkin_EXPORTED_TARGETS}
-#   )
-#   target_include_directories(test_constraint
-#     PRIVATE
-#       include
-#       ${Boost_INCLUDE_DIRS}
-#       ${catkin_INCLUDE_DIRS}
-#       ${CERES_INCLUDE_DIRS}
-#       ${CMAKE_CURRENT_SOURCE_DIR}
-#       ${EIGEN3_INCLUDE_DIRS}
-#   )
-#   target_link_libraries(test_constraint
-#     ${PROJECT_NAME}
-#   )
-#   set_target_properties(test_constraint
-#     PROPERTIES
-#       CXX_STANDARD 14
-#       CXX_STANDARD_REQUIRED YES
-#   )
-
-#   # Eigen tests
-#   catkin_add_gtest(test_eigen
-#     test/test_eigen.cpp
-#   )
-#   add_dependencies(test_eigen
-#     ${catkin_EXPORTED_TARGETS}
-#   )
-#   target_include_directories(test_eigen
-#     PRIVATE
-#       include
-#       ${Boost_INCLUDE_DIRS}
-#       ${catkin_INCLUDE_DIRS}
-#       ${CERES_INCLUDE_DIRS}
-#       ${CMAKE_CURRENT_SOURCE_DIR}
-#       ${EIGEN3_INCLUDE_DIRS}
-#   )
-#   target_link_libraries(test_eigen
-#     ${PROJECT_NAME}
-#   )
-#   set_target_properties(test_eigen
-#     PROPERTIES
-#       CXX_STANDARD 14
-#       CXX_STANDARD_REQUIRED YES
-#   )
-
-#   # Local Parameterization tests
-#   catkin_add_gtest(test_local_parameterization
-#     test/test_local_parameterization.cpp
-#   )
-#   target_include_directories(test_local_parameterization
-#     PRIVATE
-#       include
-#   )
-#   target_link_libraries(test_local_parameterization
-#     ${PROJECT_NAME}
-#   )
-#   set_target_properties(test_local_parameterization
-#     PROPERTIES
-#       CXX_STANDARD 14
-#       CXX_STANDARD_REQUIRED YES
-#   )
-
-#   # Loss tests
-#   catkin_add_gtest(test_loss
-#     test/test_loss.cpp
-#   )
-#   add_dependencies(test_loss
-#     ${catkin_EXPORTED_TARGETS}
-#   )
-#   target_include_directories(test_loss
-#     PRIVATE
-#       include
-#       ${Boost_INCLUDE_DIRS}
-#       ${catkin_INCLUDE_DIRS}
-#       ${CERES_INCLUDE_DIRS}
-#       ${CMAKE_CURRENT_SOURCE_DIR}
-#       ${EIGEN3_INCLUDE_DIRS}
-#   )
-#   target_link_libraries(test_loss
-#     ${PROJECT_NAME}
-#   )
-#   set_target_properties(test_loss
-#     PROPERTIES
-#       CXX_STANDARD 14
-#       CXX_STANDARD_REQUIRED YES
-#   )
-
-#   # Message Buffer Tests
-#   catkin_add_gtest(test_message_buffer
-#     test/test_message_buffer.cpp
-#   )
-#   add_dependencies(test_message_buffer
-#     ${catkin_EXPORTED_TARGETS}
-#   )
-#   target_include_directories(test_message_buffer
-#     PRIVATE
-#       include
-#       ${Boost_INCLUDE_DIRS}
-#       ${catkin_INCLUDE_DIRS}
-#       ${CERES_INCLUDE_DIRS}
-#       ${EIGEN3_INCLUDE_DIRS}
-#   )
-#   target_link_libraries(test_message_buffer
-#     ${catkin_LIBRARIES}
-#   )
-#   set_target_properties(test_message_buffer
-#     PROPERTIES
-#       CXX_STANDARD 14
-#       CXX_STANDARD_REQUIRED YES
-#   )
-
-#   # Timestamp Manager Tests
-#   catkin_add_gtest(test_timestamp_manager
-#     test/test_timestamp_manager.cpp
-#   )
-#   add_dependencies(test_timestamp_manager
-#     ${catkin_EXPORTED_TARGETS}
-#   )
-#   target_include_directories(test_timestamp_manager
-#     PRIVATE
-#       include
-#       ${Boost_INCLUDE_DIRS}
-#       ${catkin_INCLUDE_DIRS}
-#       ${CERES_INCLUDE_DIRS}
-#       ${EIGEN3_INCLUDE_DIRS}
-#   )
-#   target_link_libraries(test_timestamp_manager
-#     ${PROJECT_NAME}
-#     ${catkin_LIBRARIES}
-#   )
-#   set_target_properties(test_timestamp_manager
-#     PROPERTIES
-#       CXX_STANDARD 14
-#       CXX_STANDARD_REQUIRED YES
-#   )
-
-#   # Transaction tests
-#   catkin_add_gtest(test_transaction
-#     test/test_transaction.cpp
-#   )
-#   add_dependencies(test_transaction
-#     ${catkin_EXPORTED_TARGETS}
-#   )
-#   target_include_directories(test_transaction
-#     PRIVATE
-#       include
-#       ${Boost_INCLUDE_DIRS}
-#       ${catkin_INCLUDE_DIRS}
-#       ${CERES_INCLUDE_DIRS}
-#       ${CMAKE_CURRENT_SOURCE_DIR}
-#       ${EIGEN3_INCLUDE_DIRS}
-#   )
-#   target_link_libraries(test_transaction
-#     ${PROJECT_NAME}
-#     ${catkin_LIBRARIES}
-#   )
-#   set_target_properties(test_transaction
-#     PROPERTIES
-#       CXX_STANDARD 14
-#       CXX_STANDARD_REQUIRED YES
-#   )
-
 #   # Parameter tests
 #   add_rostest_gtest(test_parameter
 #     test/parameter.test
@@ -402,80 +285,6 @@ endif()
 #     ${catkin_LIBRARIES}
 #   )
 #   set_target_properties(test_throttled_callback
-#     PROPERTIES
-#       CXX_STANDARD 14
-#       CXX_STANDARD_REQUIRED YES
-#   )
-
-#   # Util tests
-#   catkin_add_gtest(test_util
-#     test/test_util.cpp
-#   )
-#   add_dependencies(test_util
-#     ${catkin_EXPORTED_TARGETS}
-#   )
-#   target_include_directories(test_util
-#     PRIVATE
-#       include
-#       ${Boost_INCLUDE_DIRS}
-#       ${catkin_INCLUDE_DIRS}
-#       ${CERES_INCLUDE_DIRS}
-#       ${EIGEN3_INCLUDE_DIRS}
-#   )
-#   target_link_libraries(test_util
-#     ${catkin_LIBRARIES}
-#   )
-#   set_target_properties(test_util
-#     PROPERTIES
-#       CXX_STANDARD 14
-#       CXX_STANDARD_REQUIRED YES
-#   )
-
-#   # UUID tests
-#   catkin_add_gtest(test_uuid
-#     test/test_uuid.cpp
-#   )
-#   add_dependencies(test_uuid
-#     ${catkin_EXPORTED_TARGETS}
-#   )
-#   target_include_directories(test_uuid
-#     PRIVATE
-#       include
-#       ${Boost_INCLUDE_DIRS}
-#       ${catkin_INCLUDE_DIRS}
-#       ${CERES_INCLUDE_DIRS}
-#       ${EIGEN3_INCLUDE_DIRS}
-#   )
-#   target_link_libraries(test_uuid
-#     ${PROJECT_NAME}
-#     ${catkin_LIBRARIES}
-#   )
-#   set_target_properties(test_uuid
-#     PROPERTIES
-#       CXX_STANDARD 14
-#       CXX_STANDARD_REQUIRED YES
-#   )
-
-#   # Variable tests
-#   catkin_add_gtest(test_variable
-#     test/test_variable.cpp
-#   )
-#   add_dependencies(test_variable
-#     ${catkin_EXPORTED_TARGETS}
-#   )
-#   target_include_directories(test_variable
-#     PRIVATE
-#       include
-#       ${Boost_INCLUDE_DIRS}
-#       ${catkin_INCLUDE_DIRS}
-#       ${CERES_INCLUDE_DIRS}
-#       ${CMAKE_CURRENT_SOURCE_DIR}
-#       ${EIGEN3_INCLUDE_DIRS}
-#   )
-#   target_link_libraries(test_variable
-#     ${PROJECT_NAME}
-#   )
-#   set_target_properties(test_variable
 #     PROPERTIES
 #       CXX_STANDARD 14
 #       CXX_STANDARD_REQUIRED YES
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -32,6 +32,7 @@ add_library(${PROJECT_NAME} SHARED
 #    src/async_motion_model.cpp
 #    src/async_publisher.cpp
 #    src/async_sensor_model.cpp
+  src/callback_wrapper.cpp
 #    src/ceres_options.cpp
     src/constraint.cpp
     src/graph.cpp
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -29,11 +29,11 @@ include(boost-extras.cmake)
 
 ## fuse_core library
 add_library(${PROJECT_NAME} SHARED
-#    src/async_motion_model.cpp
-#    src/async_publisher.cpp
-#    src/async_sensor_model.cpp
-  src/callback_wrapper.cpp
-#    src/ceres_options.cpp
+    src/async_motion_model.cpp
+#   src/async_publisher.cpp
+#   src/async_sensor_model.cpp
+    src/callback_wrapper.cpp
+#   src/ceres_options.cpp
     src/constraint.cpp
     src/graph.cpp
     src/graph_deserializer.cpp
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -33,7 +33,7 @@ add_library(${PROJECT_NAME} SHARED
   src/async_publisher.cpp
   src/async_sensor_model.cpp
   src/callback_wrapper.cpp
-  src/ceres_options.cpp
+  # src/ceres_options.cpp
   src/constraint.cpp
   src/graph.cpp
   src/graph_deserializer.cpp
--- a/fuse_core/CMakeLists.txt
+++ b/fuse_core/CMakeLists.txt
@@ -29,21 +29,22 @@ include(boost-extras.cmake)
 
 ## fuse_core library
 add_library(${PROJECT_NAME} SHARED
-    src/async_motion_model.cpp
-#   src/async_publisher.cpp
-#   src/async_sensor_model.cpp
-    src/callback_wrapper.cpp
-#   src/ceres_options.cpp
-    src/constraint.cpp
-    src/graph.cpp
-    src/graph_deserializer.cpp
-    src/loss.cpp
-    src/serialization.cpp
-    src/timestamp_manager.cpp
-    src/transaction.cpp
-    src/transaction_deserializer.cpp
-    src/uuid.cpp
-    src/variable.cpp
+  src/async_motion_model.cpp
+  src/async_publisher.cpp
+  src/async_sensor_model.cpp
+  src/callback_wrapper.cpp
+  src/ceres_options.cpp
+  src/constraint.cpp
+  src/graph.cpp
+  src/graph_deserializer.cpp
+  src/loss.cpp
+  src/serialization.cpp
+  src/time.cpp
+  src/timestamp_manager.cpp
+  src/transaction.cpp
+  src/transaction_deserializer.cpp
+  src/uuid.cpp
+  src/variable.cpp
 )
 target_include_directories(${PROJECT_NAME} PUBLIC
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
--- a/fuse_core/include/fuse_core/async_motion_model.h
+++ b/fuse_core/include/fuse_core/async_motion_model.h
@@ -34,14 +34,15 @@
 #ifndef FUSE_CORE_ASYNC_MOTION_MODEL_H
 #define FUSE_CORE_ASYNC_MOTION_MODEL_H
 
-#include <fuse_core/graph.h>
-#include <fuse_core/fuse_macros.h>
 #include <fuse_core/motion_model.h>
+
 #include <fuse_core/transaction.h>
-// #include <rclcpp/rclcpp.hpp>  TODO(CH3): Uncomment when ready
-#include <ros/callback_queue.h>
-#include <ros/node_handle.h>
-#include <ros/spinner.h>
+#include <fuse_core/graph.h>
+#include <fuse_core/callback_wrapper.h>
+
+//#include <fuse_core/fuse_macros.h>
+
+#include <rclcpp/rclcpp.hpp>
 
 #include <string>
 
@@ -171,13 +172,12 @@ public:
   void stop() override;
 
 protected:
-  ros::CallbackQueue callback_queue_;  //!< The local callback queue used for all subscriptions
+  std::shared_ptr<fuse_core::CallbackAdapter> callback_queue_; //!< The callback queue used for fuse internal callbacks
   std::string name_;  //!< The unique name for this motion model instance
-  // rclcpp::Node::SharedPtr node_;  //!< The node for this motion model (TODO(CH3): Uncomment when it's time)
-  ros::NodeHandle node_handle_;  //!< A node handle in the global namespace using the local callback queue
-  ros::NodeHandle private_node_handle_;  //!< A node handle in the private namespace using the local callback queue
-  ros::AsyncSpinner spinner_;  //!< A single/multi-threaded spinner assigned to the local callback queue
-
+  rclcpp::Node::SharedPtr node_;  //!< The node for this motion model
+  rclcpp::executors::MultiThreadedExecutor::SharedPtr executor_;  //!< A single/multi-threaded spinner assigned to the local callback queue
+  rclcpp::node_interfaces::NodeWaitablesInterface::SharedPtr waitables_interface_;
+  size_t executor_thread_count_;
   /**
    * @brief Constructor
    *
--- a/fuse_core/include/fuse_core/async_motion_model.h
+++ b/fuse_core/include/fuse_core/async_motion_model.h
@@ -38,6 +38,8 @@
 #include <fuse_core/fuse_macros.h>
 #include <fuse_core/motion_model.h>
 #include <fuse_core/transaction.h>
+#include <fuse_core/callback_wrapper.h>
+
 #include <rclcpp/rclcpp.hpp>
 
 #include <string>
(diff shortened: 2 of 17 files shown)


methylDragon:
fuse_core builds if https://github.com/locusrobotics/fuse/pull/283 is merged and this is rebased (with conflicts resolved)
I'm going to mark this PR as ready.

sloretz:
> fuse_core builds if https://github.com/locusrobotics/fuse/pull/283 is merged and this is rebased (with conflicts resolved)

:tada: That's great! That means it's a great time to start re-enabling tests. I opened a PR against the `ros-params` branch to enable the gtest ones in https://github.com/methylDragon/fuse/pull/1 . Once #283 merges, mind cherry-picking that commit to this branch?

Is enough of fuse_core ported with this PR to enable all of the tests? (the rest of the tests will need to be migrated to `launch_pytest`).

fuse_core/CMakeLists.txt
```diff
@@ -29,20 +29,22 @@ include(boost-extras.cmake)
...
-    src/timestamp_manager.cpp
-    src/transaction.cpp
-    src/transaction_deserializer.cpp
-    src/uuid.cpp
-    src/variable.cpp
+  src/async_motion_model.cpp
+  src/async_publisher.cpp
+  src/async_sensor_model.cpp
+  src/callback_wrapper.cpp
+  # src/ceres_options.cpp
```
sloretz, line 36: Is this one going to be ported in this branch, or a future one?

fuse_core/include/fuse_core/async_motion_model.h
```diff
@@ -34,14 +34,15 @@
...
+
 #include <fuse_core/transaction.h>
-// #include <rclcpp/rclcpp.hpp>  TODO(CH3): Uncomment when ready
-#include <ros/callback_queue.h>
-#include <ros/node_handle.h>
-#include <ros/spinner.h>
+#include <fuse_core/graph.h>
+#include <fuse_core/callback_wrapper.h>
+
+//#include <fuse_core/fuse_macros.h>
```
sloretz: nit, can this be deleted?

fuse_core/include/fuse_core/async_motion_model.h
```diff
@@ -171,13 +172,12 @@ class AsyncMotionModel : public MotionModel
...
 protected:
-  ros::CallbackQueue callback_queue_;  //!< The local callback queue used for all subscriptions
+  std::shared_ptr<fuse_core::CallbackAdapter> callback_queue_; //!< The callback queue used for fuse internal callbacks
   std::string name_;  //!< The unique name for this motion model instance
-  // rclcpp::Node::SharedPtr node_;  //!< The node for this motion model (TODO(CH3): Uncomment when it's time)
-  ros::NodeHandle node_handle_;  //!< A node handle in the global namespace using the local callback queue
-  ros::NodeHandle private_node_handle_;  //!< A node handle in the private namespace using the local callback queue
-  ros::AsyncSpinner spinner_;  //!< A single/multi-threaded spinner assigned to the local callback queue
-
+  rclcpp::Node::SharedPtr node_;  //!< The node for this motion model
```
sloretz, line 175: It looks like the node is unused except for adding the waitable to the executor.

Maybe we could pass the node or node interfaces [via `MotionModel::Initialize()`](https://github.com/methylDragon/fuse/blob/ee39c45cb7a704e0c646daf53993cfb5085f0d7c/fuse_core/include/fuse_core/motion_model.h#L97)? That eliminates the extra context and node, and it looks doable higher in the stack where [`MotionModel::initialize()` is called by an `Optimizer`](https://github.com/methylDragon/fuse/blob/ee39c45cb7a704e0c646daf53993cfb5085f0d7c/fuse_optimizers/src/optimizer.cpp#L182) which [does have a Node](https://github.com/methylDragon/fuse/blob/ee39c45cb7a704e0c646daf53993cfb5085f0d7c/fuse_optimizers/src/optimizer.cpp#L83).

fuse_core/include/fuse_core/callback_wrapper.h
```diff
@@ -127,13 +141,59 @@ class CallbackWrapper : public ros::CallbackInterface
...
+class CallbackAdapter : public rclcpp::Waitable
+{
+public:
+
+  CallbackAdapter(std::shared_ptr<rclcpp::Context> context_ptr);
+
+  /**
+   * @brief tell the CallbackGroup how many guard conditions are ready in this waitable
+   */
+  size_t get_number_of_ready_guard_conditions();
```
sloretz: nit, `override` here and the methods below.

methylDragon:
> > fuse_core builds if #283 is merged and this is rebased (with conflicts resolved)
> 
> tada That's great! That means it's a great time to start re-enabling tests. I opened a PR against the `ros-params` branch to enable the gtest ones in [methylDragon#1](https://github.com/methylDragon/fuse/pull/1) . Once #283 merges, mind cherry-picking that commit to this branch?
> 
> Is enough of fuse_core ported with this PR to enable all of the tests? (the rest of the tests will need to be migrated to `launch_pytest`).

This PR should migrate everything **except** ceres options (and its interactions with ROS 2 parameters)
I was gonna get the params first, but that might take awhile, so enabling tests first and fixing them probably makes more sense

fuse_core/include/fuse_core/callback_wrapper.h
```diff
@@ -127,13 +141,59 @@ class CallbackWrapper : public ros::CallbackInterface
...
+
+  void addCallback(const std::shared_ptr<CallbackWrapperBase> &callback);
+
+  void addCallback(std::shared_ptr<CallbackWrapperBase> && callback);
+
+  void removeAllCallbacks();
+
+
+private:
+  std::recursive_mutex reentrant_mutex_;  //!< mutex to allow this callback to be added to multiple callback groups simultaneously
```
sloretz: Does "multiple callback groups simultaneously" mean adding the same Waitable instance to multiple executors? That sounds undesirable. If two executors are trying to execute the same thing at the same time then one is going to be blocked, when it could have been executing something else.

Can this be removed?

fuse_core/src/async_motion_model.cpp
```diff
@@ -62,56 +62,79 @@ bool AsyncMotionModel::apply(Transaction& transaction)
...
   auto result = callback->getFuture();
-  callback_queue_.addCallback(callback, reinterpret_cast<uint64_t>(this));
+  callback_queue_->addCallback(callback);
   result.wait();
 }
 
 void AsyncMotionModel::stop()
 {
-  if (ros::ok())
+  if (rclcpp::ok())
```
sloretz: This should probably  `context->is_valid()` taken from the owner of the instance, which seems to be the `Optimizer` class downstream. Usually there should only be one `rclcpp::Context` per process as that's the barrier for inter-process communication.

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+    rcl_guard_condition_options_t guard_condition_options =
+        rcl_guard_condition_get_default_options();
+
+    // Guard condition is used by the wait set to handle execute-or-not logic
+    gc_ = rcl_get_zero_initialized_guard_condition();
+    if (RCL_RET_OK != rcl_guard_condition_init(
+      &gc_, context_ptr->get_rcl_context().get(), guard_condition_options)
+    ) {
+      RCLCPP_WARN(rclcpp::get_logger("fuse"),
```
sloretz: I'd recommend throwing `std::runtime_error` or another exception here to stop construction. There's no use for this class instance if we can't create the guard condition.

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+
+  /**
+   * @brief tell the CallbackGroup that this waitable is ready to execute anything
+   */
+  bool CallbackAdapter::is_ready(rcl_wait_set_t * wait_set) {
+    (void) wait_set;
+    // std::lock_guard<std::recursive_mutex> lock(queue_mutex_);
+    // a thread glitch isn't a disaster here and a mutex wouldn't stop it anyway
+    return !callback_queue_.empty();
```
sloretz, line 64: I think this could cause a multithreaded executor to wake multiple times (until the callback starts executing). I think `callback_queue_` should be protected by a mutex, but it doesn't need to be a `recursive_mutex`, just a regular one should do.

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,145 @@
...
+    if (RCL_RET_OK != rcl_wait_set_add_guard_condition(wait_set, &gc_, NULL)) {
+    }
+  }
+
+  /**
+   * @brief unused part of the rclcpp::waitables interface
+   *
+   */
+  std::shared_ptr< void > CallbackAdapter::take_data(){
+    return nullptr;
```
sloretz: The idea behind `take_data` is the executor asks `rcl_wait` if there's any data to execute, finds out that the waitable is ready, takes the data from the lower level, and then passes it to the thing that's going to execute it.

Imagine there are multiple callbacks in the queue owned by `CallbackAdapter`, and a multithreaded executor is used. With the current implementation the executor discovers one callback is ready, queues it to execute, but doesn't learn that there's another callback until the first callback starts executing. This increases the time it takes for all callbacks after the first one to start, and the delay increases for each additional callback.

The fix is to make this the method where `callback_queue` is popped from. It should return the callback as a `std::shared_ptr<void>` that was created using [the `aliasing constructor`](https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr). This pointer will later be fed as the parameter to `execute()`. This allows the executor to start executing all callbacks rapidly without waiting for the first one to start executing.

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,145 @@
...
+   *
+   */
+  void CallbackAdapter::execute(std::shared_ptr<void> & /*data*/){
+    std::shared_ptr<CallbackWrapperBase> cb_wrapper = nullptr;
+    // fetch the callback ptr and release the lock without spending time in the callback
+    {
+      std::lock_guard<std::recursive_mutex> lock(queue_mutex_);
+      if(!callback_queue_.empty()){
+        cb_wrapper = callback_queue_.front();
+        callback_queue_.pop_front();
```
sloretz, line 90: Move this to `take_data()`

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,145 @@
...
+   */
+  std::shared_ptr< void > CallbackAdapter::take_data(){
+    return nullptr;
+  }
+
+  /**
+   * @brief hook that allows the rclcpp::waitables interface to run the next callback
+   *
+   */
+  void CallbackAdapter::execute(std::shared_ptr<void> & /*data*/){
```
sloretz: use the `data` argument.

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,145 @@
...
+    // fetch the callback ptr and release the lock without spending time in the callback
+    {
+      std::lock_guard<std::recursive_mutex> lock(queue_mutex_);
+      if(!callback_queue_.empty()){
+        cb_wrapper = callback_queue_.front();
+        callback_queue_.pop_front();
+      }
+    }
+    //the lock is released and the callback is no longer associated with the queue, run it.
+    if(cb_wrapper) {
```
sloretz: This would become `data`, and if `data` is `NULL` something has been programmed wrong. I would make this an `assert`.

fuse_core/CMakeLists.txt
sloretz: Answered by https://github.com/locusrobotics/fuse/pull/284#issuecomment-1309519542

fuse_core/include/fuse_core/async_motion_model.h
BrettRD: The node here is mostly used to structure parameters in the config files
It's reasonable to run the graph with dozens of named motion_models, and each one needs a unique name and parameters, each motion model can be associated with a different physical component that may or may not be included in any given deployment.

The separated node also allows complex motion_models (usually associated with a single robot chassis) to be targeted by ros cli tools as a separate entity in the ROS graph

fuse_core/include/fuse_core/callback_wrapper.h
BrettRD: the comment is mistaken, (oops)
reentrant_mutex is called  to protect `rcl_wait_set_add_guard_condition`, which inserts a pointer into a list of pointers and is not thread-safe.  This allows multiple threads to insert callbacks into one queue without trashing RCL's internal data structures

queue_mutex is the one you're worried about, it allows multiple executors to pull jobs from the queue.
If someone managed to put the wrapper into multiple executors without the mutex, it would be a nightmare to debug.

The lock is released before the callback is run. An executor will only be blocked by the time it takes for another executor to pull a pointer from a queue.
https://github.com/locusrobotics/fuse/blob/ee39c45cb7a704e0c646daf53993cfb5085f0d7c/fuse_core/src/callback_wrapper.cpp#L102-L112
Executors won't try to execute the same code twice, and they won't wait for a callback to finish either.

edit: grammar

fuse_core/src/callback_wrapper.cpp
BrettRD: I think I was reading from the executor pattern in rclc, it's good to hear the additional context

fuse_core/include/fuse_core/async_motion_model.h
methylDragon: @BrettRD would you say that it's preferable then to keep the nodes separated? This was one of the places I planned to circle back on to try to combine it all into a single node (with the iffy bit being what happens when all the callback queues are merged), but the point you raised about CLI tools is pretty valid..

I'm not sure of the implications on lifecycle nodes and node components though.

fuse_core/src/callback_wrapper.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/fb5254343a1098ead1608a3d65580615e611f1d6

fuse_core/src/callback_wrapper.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/fb5254343a1098ead1608a3d65580615e611f1d6

fuse_core/src/callback_wrapper.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/fb5254343a1098ead1608a3d65580615e611f1d6

(GitHub isn't marking this as outdated for some reason)

fuse_core/src/callback_wrapper.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/fb5254343a1098ead1608a3d65580615e611f1d6

fuse_core/src/callback_wrapper.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/fb5254343a1098ead1608a3d65580615e611f1d6

Github isn't marking this as outdated for some reason..

fuse_core/src/callback_wrapper.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/fb5254343a1098ead1608a3d65580615e611f1d6

fuse_core/include/fuse_core/callback_wrapper.h
methylDragon: Updated the comment https://github.com/locusrobotics/fuse/pull/284/commits/fb5254343a1098ead1608a3d65580615e611f1d6

fuse_core/include/fuse_core/callback_wrapper.h
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/fb5254343a1098ead1608a3d65580615e611f1d6

fuse_core/include/fuse_core/async_motion_model.h
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/fb5254343a1098ead1608a3d65580615e611f1d6

fuse_core/src/async_motion_model.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/ba64b63714b8acaa5409946b44d84fc7b5ee1d58

Note that the optimizer becomes a child class of rclcpp::Node in the time PR.

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+
+  /**
+   * @brief tell the CallbackGroup that this waitable is ready to execute anything
+   */
+  bool CallbackAdapter::is_ready(rcl_wait_set_t * wait_set) {
+    (void) wait_set;
+
+    // Prevent multiple simultaneous calls of is_ready (causing the executor to wake multiple times)
+    std::lock_guard<std::mutex> lock(ready_mutex_);
```
sloretz: My earlier comment about the executor waking multiple times in `is_ready()` doesn't make sense to me today :no_mouth: 

This should be `queue_mutex_`, instead of `ready_mutex_`, but maybe no-mutex is OK after all, assuming the executor won't call `is_ready()` again until after it calls `take_data()`.

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+
+  /**
+   * @brief add_to_wait_set is called by rclcpp during NodeWaitables::add_waitable() and CallbackGroup::add_waitable()
+    waitable_ptr = std::make_shared<CallbackAdapter>();
+    node->get_node_waitables_interface()->add_waitable(waitable_ptr, (rclcpp::CallbackGroup::SharedPtr) nullptr);
+   */
+  void CallbackAdapter::add_to_wait_set(rcl_wait_set_t * wait_set)
+  {
+    std::lock_guard<std::recursive_mutex> lock(reentrant_mutex_);
```
sloretz: I think a `std::mutex` would do. I don't see anywhere that `reentrant_mutex_` would be aquired by a thread that already holds it.

> reentrant_mutex is called to protect rcl_wait_set_add_guard_condition, which inserts a pointer into a list of pointers and is not thread-safe. This allows multiple threads to insert callbacks into one queue without trashing RCL's internal data structures

I notice the [rclcpp_action Client doesn't bother to lock](https://github.com/ros2/rclcpp/blob/91bc312190e61c8915b0152e5c927c987fc22d6c/rclcpp_action/src/client.cpp#L258). It seems like the action client is implicitly assuming an executor won't try to modify the waitset in parallel. It seems like we could make the same assumption here and remove this lock.

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * @brief check the callback queue and return the next callback to run
+   *
+   */
+  std::shared_ptr< void > CallbackAdapter::take_data(){
+    std::shared_ptr<CallbackWrapperBase> cb_wrapper = nullptr;
+    // fetch the callback ptr and release the lock without spending time in the callback
+    {
+      std::lock_guard<std::recursive_mutex> lock(queue_mutex_);
```
sloretz: It looks like `std::mutex` would do.

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,148 @@
...
+      if(!callback_queue_.empty()){
+        cb_wrapper = callback_queue_.front();
+        callback_queue_.pop_front();
+      }
+    }
+    if (cb_wrapper) {
```
sloretz: Ah maybe the aliasing constructor isn't as convenient. Can this `if(){return ...} return;` be eliminated with:

```c++
return std::static_pointer_cast<void>(cb_wrapper);
```

sloretz (commented):
In-progress review. Github is now showing all the Time changes in this PR. I'll try an old trick of changing the base and back to see if that fixes it.

fuse_core/include/fuse_core/async_motion_model.h
BrettRD: I don't know what's better. (edit: it's a pretty huge API divergence to change the shape of Fuse on the node graph)

Have a think about sensor_models and fuse_publishers too, they will need to interact with sensor nodes and planning tools that will definitely appear under ros namespaces, so it would be nice for a plugin to exist in the node graph under the relevant namespace

One node wouldn't be a disaster.
You can still isolate device-specific configs into separate conf-only packages and load multiple config files into the optimiser node at launch.
However, if it's one node, it's going to make a blob in the node graph so dense it'll become a running joke. edit: (I have this exact problem exposing gstreamer properties on a pipeline node, it's not unmanageable)

I was expecting to run separate nodes and have to do some fancy footwork to allow fuse_optimiser to side-load its plugins into its host composable node container, and then have a lot of trouble getting access to the parameters.
Either that, or have fuse_optimiser inherit from container, and cause a headache when any other package tries to do the same
I haven't studied lifecycle nodes enough to comment there

I don't think the callback queues will be a problem; executors are built to cope with multiple queues, and most of the fuse plugins are sharing the one optimiser queue

sloretz: > The node here is mostly used to structure parameters in the config files
It's reasonable to run the graph with dozens of named motion_models, and each one needs a unique name and parameters, each motion model can be associated with a different physical component that may or may not be included in any given deployment.

I think we can still structure parameters in the config file by prepending `name_ + '.param_name'` to each parameter name.

> The separated node also allows complex motion_models (usually associated with a single robot chassis) to be targeted by ros cli tools as a separate entity in the ROS graph

I can think of a few reasons to avoid multiple nodes, but I'm not familiar enough with fuse to judge how useful it would be to a node per async motion model/publisher/sensor model. @svwilliams what are your thoughts?

Reasons to avoid multiple nodes:
* Nodes come with extra overhead like a graph guard condition, a rosout publisher, a parameter events publisher, and services for setting and getting parameters.
* Atomically changing parameters is scoped to a single node. A single node would allow someone to change parameters to two motion models atomically, such that if one failed (maybe from an invalid value) none of the changes would go through
* Building the nodes internally prevents choosing whether they use regular or lifecycle nodes.

fuse_core/src/callback_wrapper.cpp
BrettRD: `add_to_wait_set` is part of a call chain from user code (during init in optimiser and fuse_core/src/async_*.cpp), but I think you're right that there's no reason to expect two threads to ever call that concurrently

fuse_core/src/callback_wrapper.cpp
BrettRD: `std::mutex` is indeed enough, reentrant/recursive is a hangover from when I expected a need for concurrent reads from a multi-threaded executor

fuse_core/src/callback_wrapper.cpp
BrettRD: habitually locking data structures (queue) that might be modified by another thread is good practices,
but in this case it can still report work is ready to be done when another thread has previously checked and is blocked waiting to pop the last job; so it didn't seem worth locking.

fuse_core/src/callback_wrapper.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/cdeca18dee79f9a50a3788c844308a84f192b717

fuse_core/src/callback_wrapper.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/cdeca18dee79f9a50a3788c844308a84f192b717

fuse_core/src/callback_wrapper.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/cdeca18dee79f9a50a3788c844308a84f192b717

fuse_core/src/callback_wrapper.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/cdeca18dee79f9a50a3788c844308a84f192b717

fuse_core/src/async_motion_model.cpp
```diff
@@ -62,56 +62,77 @@ bool AsyncMotionModel::apply(Transaction& transaction)
...
 void AsyncMotionModel::initialize(const std::string& name)
 {
   // Initialize internal state
   name_ = name;
-  node_handle_.setCallbackQueue(&callback_queue_);
-  private_node_handle_ = ros::NodeHandle("~/" + name_);
-  private_node_handle_.setCallbackQueue(&callback_queue_);
+  std::string node_namespace = "";
+
+  rclcpp::Context::SharedPtr ros_context = rclcpp::contexts::get_global_default_context();
```
sloretz, line 81: Definitely add a TODO to pass either the context or the node in.

fuse_core/src/async_motion_model.cpp
```diff
@@ -62,56 +62,77 @@ bool AsyncMotionModel::apply(Transaction& transaction)
...
+}
+
+void AsyncMotionModel::graphCallback(Graph::ConstSharedPtr graph)
+{
+  auto callback = std::make_shared<CallbackWrapper<void>>(
+    std::bind(&AsyncMotionModel::onGraphUpdate, this, std::move(graph))
+  );
+  auto result = callback->getFuture();
+  callback_queue_->addCallback(callback);
+  result.wait();
```
sloretz: I think the intent of the separate queue and executor is to not wait for the result here.

fuse_core/src/async_publisher.cpp
```diff
@@ -33,71 +33,76 @@
...
 void AsyncPublisher::initialize(const std::string& name)
 {
   // Initialize internal state
   name_ = name;
-  node_handle_.setCallbackQueue(&callback_queue_);
-  private_node_handle_ = ros::NodeHandle("~/" + name_);
-  private_node_handle_.setCallbackQueue(&callback_queue_);
+  std::string node_namespace = "";
+
+  rclcpp::Context::SharedPtr ros_context = rclcpp::contexts::get_global_default_context();
```
sloretz, line 55: same comment about TODO to pass in either a context or a node here, pending resolution of https://github.com/locusrobotics/fuse/pull/284/files#r1018501020

fuse_core/src/async_publisher.cpp
```diff
@@ -33,71 +33,76 @@
...
+  callback_queue_ = std::make_shared<CallbackAdapter>(ros_context);
+  node_->get_node_waitables_interface()->add_waitable(
+    callback_queue_, (rclcpp::CallbackGroup::SharedPtr) nullptr);
 
   // Call the derived onInit() function to perform implementation-specific initialization
   onInit();
 
   // Start the async spinner to service the local callback queue
-  spinner_.start();
+  // XXX spinner_.start();
```
sloretz: `executor_->add_node(...)`

fuse_core/src/async_sensor_model.cpp
```diff
@@ -36,47 +36,60 @@
...
   TransactionCallback transaction_callback)
 {
   // Initialize internal state
   name_ = name;
-  node_handle_.setCallbackQueue(&callback_queue_);
-  private_node_handle_ = ros::NodeHandle("~/" + name_);
-  private_node_handle_.setCallbackQueue(&callback_queue_);
+  std::string node_namespace = "";
+
+  rclcpp::Context::SharedPtr ros_context = rclcpp::contexts::get_global_default_context();
```
sloretz, line 65: Same comment about TODO to pass in either context or node

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,135 @@
...
+      RCLCPP_WARN(rclcpp::get_logger("fuse"),
+                  "Could not trigger guard condition for callback, pulling callback off the queue.");
+      callback_queue_.pop_back();  // Undo
+    }
+  }
+
+  void CallbackAdapter::removeAllCallbacks(){
+    std::lock_guard<std::mutex> lock(queue_mutex_);
+    while(!callback_queue_.empty()){
+      callback_queue_.pop_front();
```
sloretz: nit, `callback_queue_.clear()`: https://en.cppreference.com/w/cpp/container/deque/clear

fuse_core/src/async_motion_model.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/7f463a9ba85b274a53f618159fac00ae32b29f05

fuse_core/src/async_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/7f463a9ba85b274a53f618159fac00ae32b29f05

fuse_core/src/async_sensor_model.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/7f463a9ba85b274a53f618159fac00ae32b29f05

fuse_core/src/callback_wrapper.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/b7480f866847a6801144837ca13dd8ef6c561f0a

fuse_core/src/async_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/284/commits/b7480f866847a6801144837ca13dd8ef6c561f0a

fuse_core/src/async_motion_model.cpp
methylDragon: I removed it everywhere where it made sense to https://github.com/locusrobotics/fuse/pull/284/commits/d3b6c944272008f0ad4f62582c73deb350f3b9e1

There is only one wait() remaining, in [`bool AsyncMotionModel::apply(Transaction& transaction)`](https://github.com/methylDragon/fuse/blob/rolling-waitables/fuse_core/src/async_motion_model.cpp#L71)

I commented them out and added a TODO though, just in case we need to wait due to race conditions/etc. when everything compiles

fuse_core/include/fuse_core/callback_wrapper.h
```diff
@@ -113,10 +128,9 @@ class CallbackWrapper : public ros::CallbackInterface
   /**
    * @brief Call this function. This is used by the callback queue.
    */
-  CallResult call() override
+  void call()
```
sloretz: nit, `override`

fuse_core/src/async_motion_model.cpp
```diff
@@ -62,56 +62,79 @@ bool AsyncMotionModel::apply(Transaction& transaction)
...
-  auto callback = boost::make_shared<CallbackWrapper<void>>(std::bind(&AsyncMotionModel::onStart, this));
-  auto result = callback->getFuture();
-  callback_queue_.addCallback(callback, reinterpret_cast<uint64_t>(this));
-  result.wait();
+  auto callback = std::make_shared<CallbackWrapper<void>>(
+    std::bind(&AsyncMotionModel::onStart, this)
+  );
+  // auto result = callback->getFuture();  // TODO(CH3): Circle back if we need this again
+  callback_queue_->addCallback(callback);
+  // result.wait();
```
sloretz: I think it's needed in `start` - at least that's what the ROS 1 code is doing. It looks like starting/stopping is synchronous and graphCallback is asynchronous in the ROS 1 code.

fuse_core/src/async_motion_model.cpp
```diff
@@ -62,56 +62,79 @@ bool AsyncMotionModel::apply(Transaction& transaction)
...
-    auto callback = boost::make_shared<CallbackWrapper<void>>(std::bind(&AsyncMotionModel::onStop, this));
-    auto result = callback->getFuture();
-    callback_queue_.addCallback(callback, reinterpret_cast<uint64_t>(this));
-    result.wait();
+    auto callback = std::make_shared<CallbackWrapper<void>>(
+      std::bind(&AsyncMotionModel::onStop, this)
+    );
+    // auto result = callback->getFuture();  // TODO(CH3): Circle back if we need this again
+    callback_queue_->addCallback(callback);
+    // result.wait();
```
sloretz: Same comment about the ROS 1 version being synchronous here.

fuse_core/src/async_publisher.cpp
```diff
@@ -33,71 +33,77 @@
...
 void AsyncPublisher::start()
 {
-  auto callback = boost::make_shared<CallbackWrapper<void>>(std::bind(&AsyncPublisher::onStart, this));
-  auto result = callback->getFuture();
-  callback_queue_.addCallback(callback, reinterpret_cast<uint64_t>(this));
-  result.wait();
+  auto callback = std::make_shared<CallbackWrapper<void>>(std::bind(&AsyncPublisher::onStart, this));
+  // auto result = callback->getFuture();  // TODO(CH3): Circle back if we need this again
+  callback_queue_->addCallback(callback);
+  // result.wait();
```
sloretz: Same comment about `start/stop` being synchronous in the ROS 1 version

fuse_core/src/async_publisher.cpp
```diff
@@ -33,71 +33,77 @@
...
+  if (node_->get_node_base_interface()->get_context()->is_valid())
   {
-    auto callback = boost::make_shared<CallbackWrapper<void>>(std::bind(&AsyncPublisher::onStop, this));
-    auto result = callback->getFuture();
-    callback_queue_.addCallback(callback, reinterpret_cast<uint64_t>(this));
-    result.wait();
+    auto callback = std::make_shared<CallbackWrapper<void>>(std::bind(&AsyncPublisher::onStop, this));
+    // auto result = callback->getFuture();  // TODO(CH3): Circle back if we need this again
+    callback_queue_->addCallback(callback);
+    // result.wait();
```
sloretz: Same here

fuse_core/src/async_sensor_model.cpp
```diff
@@ -86,24 +101,30 @@ void AsyncSensorModel::sendTransaction(Transaction::SharedPtr transaction)
...
-  auto callback = boost::make_shared<CallbackWrapper<void>>(std::bind(&AsyncSensorModel::onStart, this));
-  auto result = callback->getFuture();
-  callback_queue_.addCallback(callback, reinterpret_cast<uint64_t>(this));
-  result.wait();
+  auto callback = std::make_shared<CallbackWrapper<void>>(
+    std::bind(&AsyncSensorModel::onStart, this)
+  );
+  // auto result = callback->getFuture();  // TODO(CH3): Circle back if we need this again
+  callback_queue_->addCallback(callback);
+  // result.wait();
```
sloretz: Same comment here

fuse_core/src/async_sensor_model.cpp
```diff
@@ -86,24 +101,30 @@ void AsyncSensorModel::sendTransaction(Transaction::SharedPtr transaction)
...
-    auto callback = boost::make_shared<CallbackWrapper<void>>(std::bind(&AsyncSensorModel::onStop, this));
-    auto result = callback->getFuture();
-    callback_queue_.addCallback(callback, reinterpret_cast<uint64_t>(this));
-    result.wait();
+    auto callback = std::make_shared<CallbackWrapper<void>>(
+      std::bind(&AsyncSensorModel::onStop, this)
+    );
+    // auto result = callback->getFuture();  // TODO(CH3): Circle back if we need this again
+    callback_queue_->addCallback(callback);
+    // result.wait();
```
sloretz: Same comment here

fuse_core/src/callback_wrapper.cpp
```diff
@@ -0,0 +1,135 @@
...
+    if (RCL_RET_OK != rcl_trigger_guard_condition(&gc_)) {
+      RCLCPP_WARN(rclcpp::get_logger("fuse"),
+                  "Could not trigger guard condition for callback, pulling callback off the queue.");
+      callback_queue_.pop_back();  // Undo
+    }
+  }
+
+  void CallbackAdapter::removeAllCallbacks(){
+    std::lock_guard<std::mutex> lock(queue_mutex_);
+    while(!callback_queue_.empty()){
```
sloretz: nit, no need to check if it's emtpy before calling `clear()`, nor loop over clear.

methylDragon: https://github.com/locusrobotics/fuse/compare/d3b6c944272008f0ad4f62582c73deb350f3b9e1..3e38f8d4c103da15b673464842e504c4d7a8f799

fuse_core/src/async_sensor_model.cpp
methylDragon: https://github.com/locusrobotics/fuse/compare/d3b6c944272008f0ad4f62582c73deb350f3b9e1..3e38f8d4c103da15b673464842e504c4d7a8f799

fuse_core/src/async_sensor_model.cpp
methylDragon: https://github.com/locusrobotics/fuse/compare/d3b6c944272008f0ad4f62582c73deb350f3b9e1..3e38f8d4c103da15b673464842e504c4d7a8f799

fuse_core/src/async_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/compare/d3b6c944272008f0ad4f62582c73deb350f3b9e1..3e38f8d4c103da15b673464842e504c4d7a8f799

fuse_core/src/async_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/compare/d3b6c944272008f0ad4f62582c73deb350f3b9e1..3e38f8d4c103da15b673464842e504c4d7a8f799

fuse_core/src/async_motion_model.cpp
methylDragon: https://github.com/locusrobotics/fuse/compare/d3b6c944272008f0ad4f62582c73deb350f3b9e1..3e38f8d4c103da15b673464842e504c4d7a8f799

fuse_core/src/async_motion_model.cpp
methylDragon: https://github.com/locusrobotics/fuse/compare/d3b6c944272008f0ad4f62582c73deb350f3b9e1..3e38f8d4c103da15b673464842e504c4d7a8f799

fuse_core/include/fuse_core/callback_wrapper.h
methylDragon: https://github.com/locusrobotics/fuse/compare/d3b6c944272008f0ad4f62582c73deb350f3b9e1..3e38f8d4c103da15b673464842e504c4d7a8f799

sloretz (approved):
Changes LGTM (noting TODO to decide on multiple nodes vs single node).

Builds locally and all tests pass except for the linters.

fuse_core/include/fuse_core/async_motion_model.h
svwilliams: This is where my lack of experience with ROS2 is going to hinder me offering sound advice.

I can discuss the design intent in ROS1 and see if that helps.
* The motion models and sensors models act very node-like. They subscribe to sensor topics, process the data, and "publish" the results for consumption by the optimizer.
* The reason the sensor models and motion models are not stand-alone nodes or nodelets in ROS1 is the each model "publishes" a _derived class object_ instead of a message. Each model creates derived Constraints and Variables for use by the Optimizer. New user-derived models may create Constraints and Variables the Optimizer node cannot know about at compile-time. I could not think of an elegant solution to transmit derived objects using the standard pub-sub mechanisms in ROS1.
* Using ROS1 plugins, however, I can define an interface to the Optimizer that sends derived object pointers around, so that was what I implemented.
* But there is a downside to the plugin system. The Optimizer class creates the sensor and motion model instances as class variables. The Optimizer can call methods of the loaded plugins, so it would be easy to poll each loaded sensor and motion model for new data. But the sensor models and motion models are _producing_ data for the Optimizer. In an ideal world, the models would decide for themselves when they had new data and would push that into the Optimizer. It is difficult for the plugins to call methods of the Optimizer.
* To solve the polling/pushing issues, there are a series callbacks registered between the Optimizer and the Models. This results in rather complex threading implications.
* To hide all all of that complexity, the AsyncModels each run their own thread(s) and callback spinner. This makes them act a lot like a conventional ROS node or nodelet. Each derived Model can define the number of threads used to service its callbacks, it gets a private node handle in its namespace for parameter reading, etc.

Does any of that help?

fuse_core/src/async_motion_model.cpp
svwilliams: The futures and waits are all about thread management.

The AsyncMotionModel is running its own callback spinner thread. So if a derived implementation created a Timer or subscribed to a sensor topic, those callbacks will run in the AsyncMotionModel spinner thread. However, the `apply()` method here is going to be called by the Optimizer, and hence it will run in the Optimizer's thread. This means any derived motion model must be very careful about thread synchronization. And it is not very obvious what callbacks will run in what threads. To make it easier to write derived motion model classes, I implemented the following:

The `apply()` method will be called by the Optimizer thread
```
bool AsyncMotionModel::apply(Transaction& transaction)
{
```

Wrap the user-defined implementation, `applyCallback()`, inside a promise.
```
  auto callback = boost::make_shared<CallbackWrapper<bool>>(
    std::bind(&AsyncMotionModel::applyCallback, this, std::ref(transaction)));
```

Create a future from the wrapper so that the `apply()` function may receive the results of the `applyCallback()` function.
```
auto result = callback->getFuture();
```

Insert the `applyCallback()` call _into the callback queue of the AsyncMotionModel_. This means that the `applyCallback()` method will be executed by the AsyncMotionModel spinner thread.
```
callback_queue_.addCallback(callback, reinterpret_cast<uint64_t>(this));
```

Wait for the AsyncMotionModel spinner thread to complete the `applyCallback()` call.
```
result.wait();
```

Return the results of the `applyCallback()` method out of the `apply()` method.
```
return result.get();
```

This was done so that the user-implementation of the `apply()` business logic, implemented in the `applyCallback()` method, will execute synchronously with any other callbacks (timers, subscriptions, services, etc) defined in the motion model. Typically the author of a derived AsyncMotionModel does not need to implement any mutexes or other thread synchronization mechanisms even though the execution of a motion model involves multiple threads created in different parts of the fuse ecosystem.

PR #285, opened 2022-11-12
fuse -> ROS 2 fuse_core : Messages and Services

methylDragon:

# Description
Just a small PR migrating messages and services.

Namespaces and headers were hit.

The only catch is the migration away from ConstPtrs. I tried to make a judgement call for what types to migrate to, and I left notes in the source for when I thought it was iffy.

I'm also kinda suspicious of the number of ConstSharedPtrs that I threw around.

See https://github.com/locusrobotics/fuse/pull/285/commits/43a59d76c98d1aec1126032547317d967c98d9a5 for those changes

EDIT: This will just feature changes for fuse_core, the changes from this will be sprinkled around the other packages as we get to them

### DIFF
--- a/fuse_constraints/include/fuse_constraints/absolute_orientation_3d_stamped_constraint.h
+++ b/fuse_constraints/include/fuse_constraints/absolute_orientation_3d_stamped_constraint.h
@@ -111,7 +111,7 @@ public:
   AbsoluteOrientation3DStampedConstraint(
     const std::string& source,
     const fuse_variables::Orientation3DStamped& orientation,
-    const geometry_msgs::Quaternion& mean,
+    const geometry_msgs::msg::Quaternion& mean,
     const std::array<double, 9>& covariance);
 
   /**
@@ -171,7 +171,7 @@ protected:
    * @param[in] quaternion - The input ROS quaternion message
    * @return The \p quaternion, converted to an Eigen Vector4d
    */
-  static fuse_core::Vector4d toEigen(const geometry_msgs::Quaternion& quaternion);
+  static fuse_core::Vector4d toEigen(const geometry_msgs::msg::Quaternion& quaternion);
 
   /**
    * @brief Utility method to convert a flat 1D array to a 3x3 Eigen matrix
--- a/fuse_constraints/include/fuse_constraints/absolute_orientation_3d_stamped_constraint.h
+++ b/fuse_constraints/include/fuse_constraints/absolute_orientation_3d_stamped_constraint.h
@@ -40,8 +40,8 @@
 #include <fuse_core/serialization.hpp>
 #include <fuse_core/uuid.hpp>
 #include <fuse_variables/orientation_3d_stamped.h>
-#include <geometry_msgs/PoseWithCovariance.h>
-#include <geometry_msgs/Quaternion.h>
+#include <geometry_msgs/msg/PoseWithCovariance.hpp>
+#include <geometry_msgs/msg/Quaternion.hpp>
 
 #include <boost/serialization/access.hpp>
 #include <boost/serialization/base_object.hpp>
--- a/fuse_constraints/include/fuse_constraints/absolute_orientation_3d_stamped_constraint.h
+++ b/fuse_constraints/include/fuse_constraints/absolute_orientation_3d_stamped_constraint.h
@@ -40,7 +40,8 @@
 #include <fuse_core/serialization.hpp>
 #include <fuse_core/uuid.hpp>
 #include <fuse_variables/orientation_3d_stamped.h>
-#include <geometry_msgs/msg/quaternion.hpp>
+#include <geometry_msgs/PoseWithCovariance.h>
+#include <geometry_msgs/Quaternion.h>
 
 #include <boost/serialization/access.hpp>
 #include <boost/serialization/base_object.hpp>
@@ -110,7 +111,7 @@ public:
   AbsoluteOrientation3DStampedConstraint(
     const std::string& source,
     const fuse_variables::Orientation3DStamped& orientation,
-    const geometry_msgs::msg::Quaternion& mean,
+    const geometry_msgs::Quaternion& mean,
     const std::array<double, 9>& covariance);
 
   /**
@@ -170,7 +171,7 @@ protected:
    * @param[in] quaternion - The input ROS quaternion message
    * @return The \p quaternion, converted to an Eigen Vector4d
    */
-  static fuse_core::Vector4d toEigen(const geometry_msgs::msg::Quaternion& quaternion);
+  static fuse_core::Vector4d toEigen(const geometry_msgs::Quaternion& quaternion);
 
   /**
    * @brief Utility method to convert a flat 1D array to a 3x3 Eigen matrix
--- a/fuse_constraints/include/fuse_constraints/absolute_orientation_3d_stamped_constraint.h
+++ b/fuse_constraints/include/fuse_constraints/absolute_orientation_3d_stamped_constraint.h
@@ -40,8 +40,7 @@
 #include <fuse_core/serialization.hpp>
 #include <fuse_core/uuid.hpp>
 #include <fuse_variables/orientation_3d_stamped.h>
-#include <geometry_msgs/msg/PoseWithCovariance.hpp>
-#include <geometry_msgs/msg/Quaternion.hpp>
+#include <geometry_msgs/msg/quaternion.hpp>
 
 #include <boost/serialization/access.hpp>
 #include <boost/serialization/base_object.hpp>
--- a/fuse_constraints/test/test_absolute_orientation_3d_stamped_euler_constraint.cpp
+++ b/fuse_constraints/test/test_absolute_orientation_3d_stamped_euler_constraint.cpp
@@ -36,7 +36,7 @@
 #include <fuse_core/eigen_gtest.hpp>
 #include <fuse_core/uuid.hpp>
 #include <fuse_variables/orientation_3d_stamped.h>
-#include <geometry_msgs/Quaternion.h>
+#include <geometry_msgs/msg/Quaternion.hpp>
 
 #include <ceres/covariance.h>
 #include <ceres/problem.h>
--- a/fuse_constraints/test/test_absolute_orientation_3d_stamped_euler_constraint.cpp
+++ b/fuse_constraints/test/test_absolute_orientation_3d_stamped_euler_constraint.cpp
@@ -36,6 +36,7 @@
 #include <fuse_core/eigen_gtest.hpp>
 #include <fuse_core/uuid.hpp>
 #include <fuse_variables/orientation_3d_stamped.h>
+#include <geometry_msgs/Quaternion.h>
 
 #include <ceres/covariance.h>
 #include <ceres/problem.h>
--- a/fuse_constraints/test/test_absolute_orientation_3d_stamped_euler_constraint.cpp
+++ b/fuse_constraints/test/test_absolute_orientation_3d_stamped_euler_constraint.cpp
@@ -36,7 +36,6 @@
 #include <fuse_core/eigen_gtest.hpp>
 #include <fuse_core/uuid.hpp>
 #include <fuse_variables/orientation_3d_stamped.h>
-#include <geometry_msgs/msg/Quaternion.hpp>
 
 #include <ceres/covariance.h>
 #include <ceres/problem.h>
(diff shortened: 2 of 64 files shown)


fuse_constraints/test/test_absolute_orientation_3d_stamped_euler_constraint.cpp
```diff
@@ -36,7 +36,7 @@
 #include <fuse_core/eigen_gtest.h>
 #include <fuse_core/uuid.h>
 #include <fuse_variables/orientation_3d_stamped.h>
-#include <geometry_msgs/Quaternion.h>
+#include <geometry_msgs/msg/Quaternion.hpp>
```
svwilliams: If the message isn't used here, feel free to drop the include

fuse_models/include/fuse_models/pose_2d.h
```diff
@@ -119,20 +119,20 @@ class Pose2D : public fuse_core::AsyncSensorModel
...
    * @param[out] transaction - The generated variables and constraints are added to this transaction
    */
-  void processDifferential(const geometry_msgs::PoseWithCovarianceStamped& pose, const bool validate,
+  void processDifferential(const geometry_msgs::msg::PoseWithCovarianceStamped& pose, const bool validate,
                            fuse_core::Transaction& transaction);
 
   ParameterType params_;
 
-  geometry_msgs::PoseWithCovarianceStamped::ConstPtr previous_pose_msg_;
+  geometry_msgs::msg::PoseWithCovarianceStamped::ConstSharedPtr previous_pose_msg_;
```
svwilliams: If `process()` is not receiving a shared ptr, there's no reason for the previous pose to be a shared ptr either.

fuse_optimizers/test/test_fixed_lag_ignition.cpp
```diff
@@ -113,11 +113,16 @@ TEST(FixedLagIgnition, SetInitialState)
...
   // Wait for the optimizer to process all queued transactions
   rclcpp::Time result_timeout = node->now() + rclcpp::Duration::from_seconds(3.0);
-  auto odom_msg = nav_msgs::Odometry::ConstPtr();
+  auto odom_msg = nav_msgs::msg::Odometry::ConstSharedPtr();
   while ((!odom_msg || odom_msg->header.stamp != rclcpp::Time(3, 0)) &&
          (node->now() < result_timeout))
   {
-    odom_msg = ros::topic::waitForMessage<nav_msgs::Odometry>("/odom", rclcpp::Duration::from_seconds(1.0));
+    // TODO(CH3): Oh no, there's no equivalent rclcpp method I think...
+    //            Probably just wait a bit?? It's in a test anyway
```
svwilliams: You can make a subscriber with a callback that populates a std::future. You can then wait on the future here. It is more work, but it means you are not blocking the unit test longer than is needed.

fuse_tutorials/src/range_sensor_simulator.cpp
```diff
@@ -117,26 +117,26 @@ std::vector<Beacon> createNoisyBeacons(const std::vector<Beacon>& beacons)
 /**
  * @brief Convert the set of beacons into a pointcloud for visualization purposes
  */
-sensor_msgs::PointCloud2::ConstPtr beaconsToPointcloud(
+sensor_msgs::msg::PointCloud2::ConstSharedPtr beaconsToPointcloud(
```
svwilliams: These methods are constructing ROS messages to be published. In ROS1 I always publish messages as shared ptrs. But it seems you have been publishing and subscribing with bare objects in ROS2. If that is the case, you can drop the shared pointer complexity here entirely.

fuse_core/test/test_throttled_callback.cpp
```diff
@@ -151,33 +151,34 @@ class PointSensorModel
   /**
    * @brief Keep callback, that counts the number of times it has been called and caches the last message received
    *
-   * @param[in] msg A geometry_msgs::Point message
+   * @param[in] msg A geometry_msgs::msg::Point message
    */
-  void keepCallback(const geometry_msgs::Point::ConstPtr& msg)
+  // TODO(CH3): I'm iffy on this...
+  void keepCallback(const geometry_msgs::msg::Point::ConstSharedPtr msg)
```
svwilliams: The callback signature for the ThrottledMessageCallback is `void(const typename M&)`. So shouldn't this be `void keepCallback(const geometry_msgs::msg::Point& msg)`?

svwilliams (commented):
In ROS1, I reflexively publish and subscribe using shared ptrs to messages. I notice that has been changed to const refs to messages. I don't know enough about ROS2 yet. Is that the standard practice in ROS2?

sloretz:
> In ROS1, I reflexively publish and subscribe using shared ptrs to messages. I notice that has been changed to const refs to messages. I don't know enough about ROS2 yet. Is that the standard practice in ROS2?

@svwilliams Yes, most of the time. It has to do with how intra-process communication is handled. There's more information here: https://docs.ros.org/en/rolling/Tutorials/Demos/Intra-Process-Communication.html

If I understand correctly, in most cases const references are the most efficient. If the code is going to modify the message and re-publish it (like in a sensor processing pipeline), then non-const `std::unique_ptr` may be a better option as `rclcpp` can choose to `std::move()` the message from the publisher to the subscriber.

fuse_models/include/fuse_models/pose_2d.h
methylDragon: It would make sense to leave it a ptr type to prevent copy assignment in the implementation of `processDifferential` I think?
It's also necessary for the conditional

```
void Pose2D::processDifferential(const geometry_msgs::msg::PoseWithCovarianceStamped& pose, const bool validate,
                                 fuse_core::Transaction& transaction)
{
  auto transformed_pose = std::make_unique<geometry_msgs::msg::PoseWithCovarianceStamped>();
  transformed_pose->header.frame_id = params_.target_frame.empty() ? pose.header.frame_id : params_.target_frame;

  if (!common::transformMessage(tf_buffer_, pose, *transformed_pose))
  {
    RCLCPP_WARN_STREAM_THROTTLE(node_->get_logger(), *node_->get_clock(), 5.0 * 1000,
                                "Cannot transform pose message with stamp " << pose.header.stamp
                                << " to target frame " << params_.target_frame);
    return;
  }

  if (previous_pose_msg_)
  {
    common::processDifferentialPoseWithCovariance(
      name(),
      device_id_,
      *previous_pose_msg_,
      *transformed_pose,
      params_.independent,
      params_.minimum_pose_relative_covariance,
      params_.loss,
      params_.position_indices,
      params_.orientation_indices,
      validate,
      transaction);
  }

  previous_pose_msg_ = std::move(transformed_pose);  <----------------- here
}
```

methylDragon: (I will convert the constsharedptr to unique_ptr, though)

fuse_optimizers/test/test_fixed_lag_ignition.cpp
methylDragon: Seems that there actually is a [wait_for_message](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/wait_for_message.hpp) method in rclcpp. I'll use that when it's time to get around to fuse_optimizers. And if that fails I'll go the condition variable route (instead of futures) :grimacing:

sloretz:
@ros-pull-request-builder retest this please!

fuse_constraints/include/fuse_constraints/absolute_orientation_3d_stamped_constraint.h
```diff
@@ -40,8 +40,8 @@
 #include <fuse_core/serialization.hpp>
 #include <fuse_core/uuid.hpp>
 #include <fuse_variables/orientation_3d_stamped.h>
-#include <geometry_msgs/PoseWithCovariance.h>
-#include <geometry_msgs/Quaternion.h>
+#include <geometry_msgs/msg/PoseWithCovariance.hpp>
```
sloretz: I think this is actually `geometry_msgs/msg/pose_with_covariance.hpp`. As far as I know all generated headers in ROS 2 are lowercase with underscores, so same comment throughout the PR

sloretz: It also looks like `PoseWithCovariance` is unused in this file, so this include could be removed.

sloretz (changes requested):
I think I'd recommend making these changes on a per-package basis as each package is ported rather than all at once. That will allow using the compiler to check the paths and types are correct.

methylDragon:
> I think I'd recommend making these changes on a per-package basis as each package is ported rather than all at once. That will allow using the compiler to check the paths and types are correct.

Could we just get this in first and adjust as we go up the stack? Otherwise it'll be a guaranteed extra step for each package..

methylDragon:

fuse_core/test/test_throttled_callback.cpp
```diff
@@ -203,7 +210,9 @@ class PointSensorModel : public rclcpp::Node
 
   size_t kept_messages_{0};                           //!< Messages kept
   size_t dropped_messages_{0};                        //!< Messages dropped
-  geometry_msgs::msg::Point::ConstSharedPtr last_kept_message_;  //!< The last message kept
+
+  // We use a SharedPtr to check for nullptr just for this test
+  geometry_msgs::msg::Point::SharedPtr last_kept_message_;  //!< The last message kept
```
sloretz, line 208: Since the `keepCallback` is going to copy the fields anyways, I think `geometry_msgs::Point last_kept_message;` (no `shared_ptr`) is a better fit here. Then we can use the assignment operator on the whole message in that callback.

fuse_core/include/fuse_core/graph_deserializer.hpp
```diff
@@ -74,7 +74,7 @@ class GraphDeserializer
    * @return          A unique_ptr to a derived Graph object
    */
   inline fuse_core::Graph::UniquePtr deserialize(
-    const fuse_msgs::msg::SerializedGraph::ConstSharedPtr msg) const
+    const fuse_msgs::msg::SerializedGraph::SharedPtr msg) const
```
sloretz: Since this is just passing through to a method that treats the value as const, I think `ConstSharedPtr` makes more sense as it would allow both `std::shared_ptr<const MessageType>` and `std::shared_ptr<MessageType>` to be pasesd in.

<details><summary> Small example in C++</summary>

```c++
#include <iostream>
#include <memory>

void const_func(std::shared_ptr<const double> dp)
{
  std::cout << "Const Got some double: " << *dp << "\n";
}

void non_const_func(std::shared_ptr<double> dp)
{
  std::cout << "non Const Got some double: " << *dp << "\n";
}


int main()
{
  auto const_dp = std::make_shared<const double>(42.0);
  auto non_const_dp = std::make_shared<double>(13.0);

  const_func(const_dp);
  const_func(non_const_dp);


  // Can't do this
  // non_const_func(const_dp);
  //
  // because:
  //    $ g++ -std=c++17 a.cpp 
  //    a.cpp: In function ‘int main()’:
  //    a.cpp:23:18: error: could not convert ‘const_dp’ from ‘shared_ptr<const double>’ to ‘shared_ptr<double>’
  //       23 |   non_const_func(const_dp);
  //          |                  ^~~~~~~~
  //          |                  |
  //          |                  shared_ptr<const double>
  //

  non_const_func(non_const_dp);

  return 0;
}
```
</details>

fuse_core/include/fuse_core/transaction_deserializer.hpp
```diff
@@ -75,8 +75,11 @@ class TransactionDeserializer
    * @param[IN]  msg  The SerializedTransaction message to be deserialized
    * @return          A fuse Transaction object
    */
-  fuse_core::Transaction deserialize(const fuse_msgs::msg::SerializedTransaction::SharedPtr msg)
-  const;
+  inline fuse_core::Transaction::UniquePtr deserialize(
+    const fuse_msgs::msg::SerializedTransaction::SharedPtr msg) const
```
sloretz: same comment about `ConstSharedPtr`

fuse_core/test/test_throttled_callback.cpp
sloretz: Talked offline, resolving since this is using `nullptr` to check that a message was received.

sloretz:

sloretz (approved):
LGTM if the Rpr job comes back green

methylDragon:

methylDragon:

methylDragon:
Rpr is unstable due to deadlocking issues.
https://github.com/locusrobotics/fuse/pull/294 should be merged first to fix it.

methylDragon:

PR #286, opened 2022-11-17
fuse -> ROS 2 fuse_core : Parameters and Tests

methylDragon:

# Description

With this PR, tests now run (and pass) on `fuse_core`! (Sans linting)

This PR incorporates all remaining changes @BrettRD made to fuse_core, and then applies patches on top of them to fix parameters and the tests.

Additionally, it brings in changes from https://github.com/locusrobotics/fuse/pull/273 to allow for Ceres support in the tests. (That PR's been merged but not on the rolling branch. I've cherry-picked the changes.)

Also, this PR adds the `NodeInterfaces` class proposed in https://github.com/ros2/rclcpp/pull/2041 to `fuse_core`, to port passing of ROS node handles around.

# Notes
- The launch testing is [a little hacky](https://github.com/locusrobotics/fuse/pull/286/files#diff-287770b642463999ec892472e3077010823b0ae052308290ccee439585d6b360) (I have to assert stdout for a gtest).
- I had to [change the behavior of two tests](https://github.com/locusrobotics/fuse/pull/286/files#diff-a50addea383a0573f2962b8b227a33c4645e8118c518bc6866ce56406bf5f232R212) due to changes in how ROS 2 handles parameters (strongly typed), I think what I did made sense, but I'd like a confirmation


fuse_core/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.5)
 project(fuse_core)
 
+set(CMAKE_BUILD_TYPE RelWithDebInfo)
```
sloretz: This is normally left unset, and chosen by the person building it. Was this for debugging, or is it being set for a specific reason?

fuse_core/include/fuse_core/throttled_callback.h
```diff
@@ -161,9 +169,16 @@ class ThrottledCallback
     // (a) This is the first call, i.e. the last called time is still invalid because it has not been set yet
     // (b) The throttle period is zero, so we should always keep the callbacks
     // (c) The elpased time between now and the last called time is greater than the throttle period
-    #warn "using a valid time value as a flag"
-    const rclcpp::Time now = use_wall_time_ ? rclcpp::Clock::Clock(RCL_SYSTEM_TIME).now() : node->now();
-    if ((last_called_time_.nanoseconds() == 0) || (throttle_period_.nanoseconds() == 0) || now - last_called_time_ > throttle_period_)
+    rclcpp::Time now;
+    if (use_wall_time_) {
+      now = rclcpp::Clock(RCL_SYSTEM_TIME).now();
```
sloretz: This seems to be constructing a `Clock` every callback. I recommend storing the clock on the `ThrottledCallback` class instead since it seems to be needed for the life of the instance.

fuse_core/include/fuse_core/throttled_callback.h
```diff
@@ -52,28 +56,32 @@ template <class Callback>
...
+   * @throws if the use_wall_time is false and no NodeInterfaces is passed in
    */
   ThrottledCallback(Callback&& keep_callback = nullptr,  // NOLINT(whitespace/operators)
                     Callback&& drop_callback = nullptr,  // NOLINT(whitespace/operators)
                     const rclcpp::Duration& throttle_period = rclcpp::Duration(0,0),
-                    const bool use_wall_time = false)
+                    const bool use_wall_time = false,
+                    node_interfaces::NodeInterfaces<
+                      node_interfaces::Clock
+                    >::SharedPtr interfaces = nullptr)
```
sloretz: Looking at the implementation, the `use_wall_time` and `interfaces` arguments seem indirect. The `ThrottledCallback` class needs a persistent clock; I don't think it needs to concern itself with where the clock comes from. I would recommend replacing both of these arguments with a single argument `rclcpp::Clock::SharedPtr clock`.

That both simplifies the implementation, and would allow things like passing in a steady time clock.

fuse_core/src/async_motion_model.cpp
```diff
@@ -97,6 +105,15 @@ void AsyncMotionModel::initialize(const std::string& name)
   onInit();
 
   executor_->add_node(node_);
+
+  // TODO(CH3): Remove this if the internal node is removed
+  spinning_ = true;
+  spinner_ = std::thread([&](){
+    auto context = node_->get_node_base_interface()->get_context();
+    while(context->is_valid() && spinning_) {
```
sloretz: Executors are already supposed to stop spinning when the context becomes invalid (see [multithreaded executor](https://github.com/ros2/rclcpp/blob/91bc312190e61c8915b0152e5c927c987fc22d6c/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L89)).

I don't think the lambda or loop is needed. The thread could run `executor->spin()`, and `spinning_` being set to `False` could be replaced with `executor->cancel()`.

sloretz (commented):
Quick feedback - will come back to review parameters use and tests

fuse_core/src/async_motion_model.cpp
methylDragon: We need the lambda for starting a separate thread to host the blocking spin call (because otherwise the optimizer thread that constructs the motion models and calls initialize() will get blocked).

I'll make use of `cancel()` and remove the `spinning_` flag though!

fuse_core/CMakeLists.txt
methylDragon: I'm not sure how that slipped in, I'll remove it :grimacing:

fuse_core/src/async_motion_model.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/286/commits/329a3dc486673b87ced0c3576a4d61894bf8798e

fuse_core/include/fuse_core/throttled_callback.h
methylDragon: https://github.com/locusrobotics/fuse/pull/286/commits/329a3dc486673b87ced0c3576a4d61894bf8798e

Signature changed! I am still keeping usage of `use_wall_time` on the caller side though, at which point an appropriate clock will be set.

fuse_core/include/fuse_core/throttled_callback.h
methylDragon: https://github.com/locusrobotics/fuse/pull/286/commits/329a3dc486673b87ced0c3576a4d61894bf8798e

fuse_core/CMakeLists.txt
methylDragon: https://github.com/locusrobotics/fuse/pull/286/commits/329a3dc486673b87ced0c3576a4d61894bf8798e

methylDragon:
Wait a sec.. that caused a test regression :X

Edit: Fixed https://github.com/locusrobotics/fuse/compare/329a3dc486673b87ced0c3576a4d61894bf8798e..67c6fd9a7f97eb49dc851a119ea0940d2c7706ff

With these changes merged into the downstream PRs, no non-lint tests are broken.

fuse_core/include/fuse_core/async_publisher.h
```diff
@@ -142,8 +142,9 @@ class AsyncPublisher : public Publisher
   std::string name_;  //!< The unique name for this publisher instance
   rclcpp::Node::SharedPtr node_;  //!< The node for this publisher
   rclcpp::executors::MultiThreadedExecutor::SharedPtr executor_;  //!< A single/multi-threaded spinner assigned to the local callback queue
-  rclcpp::node_interfaces::NodeWaitablesInterface::SharedPtr waitables_interface_;
   size_t executor_thread_count_;
+  std::thread spinner_;  //!< Internal thread for spinning the executor
+  std::atomic<bool> spinning_;  //!< Flag for spinning the spin thread
```
sloretz: nit, no need for `spinning_`

fuse_core/test/test_async_publisher.cpp
```diff
@@ -91,23 +105,18 @@ TEST(AsyncPublisher, notifyCallback)
   fuse_core::Graph::ConstSharedPtr graph;  // nullptr...which is fine because we do not actually use it
   publisher.notify(transaction, graph);
   EXPECT_FALSE(publisher.callback_processed);
-  rclcpp::Time wait_time_elapsed = rclcpp::Clock(RCL_SYSTEM_TIME).now() + rclcpp::Duration::from_seconds(10.0);
-  while (!publisher.callback_processed && rclcpp::Clock(RCL_SYSTEM_TIME).now() < wait_time_elapsed)
+  rclcpp::Time wait_time_elapsed =
+    rclcpp::Clock(RCL_SYSTEM_TIME).now() + rclcpp::Duration::from_seconds(10.0);
```
sloretz: nit, I'd recommend constructing the clock once and then using it.

fuse_core/test/test_async_publisher.cpp
```diff
@@ -91,23 +105,18 @@ TEST(AsyncPublisher, notifyCallback)
...
+  while (!publisher.callback_processed
+         && rclcpp::Clock(RCL_SYSTEM_TIME).now() < wait_time_elapsed)
   {
-    rclcpp::sleep_for(rclcpp::Duration::from_seconds(0.1));
+    rclcpp::sleep_for(std::chrono::milliseconds(100));
   }
   EXPECT_TRUE(publisher.callback_processed);
 }
 
 int main(int argc, char** argv)
```
sloretz: Nit, i think the whole `int main()` function can be deleted since `ament_add_gtest` [links to gtest_main](https://github.com/ament/ament_cmake/blob/4d56683ae48c985e715207bfd8a0880900beb2f2/ament_cmake_gtest/cmake/ament_add_gtest_executable.cmake#L53).

fuse_core/test/test_async_sensor_model.cpp
```diff
@@ -121,12 +136,5 @@ TEST(AsyncSensorModel, SendTransaction)
 int main(int argc, char** argv)
```
sloretz: nit, same comment about `int main()` can be deleted.

fuse_core/test/test_async_motion_model.cpp
```diff
@@ -120,12 +135,5 @@ TEST(AsyncMotionModel, ApplyCallback)
 int main(int argc, char** argv)
```
sloretz: nit same comment about `int main()`

fuse_core/test/test_async_motion_model.cpp
```diff
@@ -53,14 +53,14 @@ class MyMotionModel : public fuse_core::AsyncMotionModel
...
+    rclcpp::sleep_for(std::chrono::milliseconds(1000));
     transaction_received = true;
     return true;
   }
 
   void onGraphUpdate(fuse_core::Graph::ConstSharedPtr /*graph*/) override
   {
-    rclcpp::sleep_for(rclcpp::Duration::from_seconds(1.0));
+    rclcpp::sleep_for(std::chrono::milliseconds(1000));
     graph_received = true;
```
sloretz, line 64: I can't comment on the right line because this PR didn't change it, but it looks to me like `transaction_received` and `graph_received` are used uninitialized. I'd recommend setting them to `false` either in the constructor or with a [default member initializer](https://en.cppreference.com/w/cpp/language/data_members#Member_initialization).

fuse_core/test/test_async_sensor_model.cpp
```diff
@@ -72,22 +72,36 @@ class MySensor : public fuse_core::AsyncSensorModel
 
   void onGraphUpdate(fuse_core::Graph::ConstSharedPtr /*graph*/) override
   {
-    rclcpp::sleep_for(rclcpp::Duration::from_seconds(1.0));
+    rclcpp::sleep_for(std::chrono::milliseconds(1000));
     graph_received = true;
```
sloretz, line 76: same comment about `graph_received` being used uninitialized

fuse_core/test/test_callback_wrapper.cpp
```diff
@@ -57,42 +56,69 @@ class MyClass
...
 {
   MyClass my_object;
   std::vector<double> data = {1.0, 2.0, 3.0, 4.0, 5.0};
-  auto callback = boost::make_shared<fuse_core::CallbackWrapper<double>>(
+
+  auto node = rclcpp::Node::make_shared("callback_wrapper_double_test_node");
+  auto callback_queue =
+    std::make_shared<fuse_core::CallbackAdapter>(node->get_node_base_interface()->get_context());
+  node->get_node_waitables_interface()->add_waitable(
+    callback_queue, (rclcpp::CallbackGroup::SharedPtr) nullptr);
```
sloretz, line 82: This seems a bit verbose. Maybe `node_base_interface->call_soon(callback)` might be a nice feature to upstream. I'll open a ticket.

fuse_core/test/test_callback_wrapper.cpp
```diff
@@ -57,42 +56,69 @@ class MyClass
...
 
+  callback_queue->addCallback(callback);
+  rclcpp::spin_until_future_complete(node, result);
+
+  // This is technically redundant but this is here to mimic how the callbacks are actually used
+  EXPECT_EQ(std::future_status::ready, result.wait_for(std::chrono::seconds(10)));
   EXPECT_EQ(15.0, output);
 }
 
 int main(int argc, char** argv)
```
sloretz: nit, same comment about `int main()`

fuse_core/test/test_throttled_callback.cpp
```diff
@@ -162,112 +187,139 @@ class PointSensorModel
...
+  std::thread spinner_;  //!< Internal thread for spinning the executor
+  rclcpp::executors::SingleThreadedExecutor::SharedPtr executor_;
+};
+
+TEST_F(TestThrottledCallback, NoDroppedMessagesIfThrottlePeriodIsZero)
 {
-  // Time should be valid after ros::init() returns in main(). But it doesn't hurt to verify.
-  // TODO(CH3): Refactor this with a passed in node
-  // ASSERT_TRUE(fuse_core::wait_for_valid(<NODE_HERE>->get_clock(), rclcpp::Duration::from_seconds(1.0)));
+  rclcpp::executors::SingleThreadedExecutor exec;
```
sloretz: nit, I think this is unused.

fuse_core/test/test_throttled_callback.cpp
```diff
@@ -162,112 +187,139 @@ class PointSensorModel
...
+  EXPECT_EQ(num_messages - 1u, sensor_model->getDroppedMessages());
 
   // Check the message kept was the first message:
-  const auto last_kept_message = sensor_model.getLastKeptMessage();
+  const auto last_kept_message = sensor_model->getLastKeptMessage();
   ASSERT_NE(nullptr, last_kept_message);
   EXPECT_EQ(0.0, last_kept_message->x);
 }
 
 int main(int argc, char** argv)
```
sloretz: nit, same comment about `int main()`

fuse_core/test/launch_tests/test_parameters.py
```diff
@@ -0,0 +1,68 @@
...
+import launch_testing.util
+import launch_testing.markers
+import launch_testing_ros
+from launch_ros.substitutions import FindPackageShare
+
+import pytest
+
+@pytest.mark.launch_test
+@launch_testing.markers.keep_alive
+def generate_test_description():
```
sloretz: Mind using `launch_pytest` instead of `launch_testing`? I think that's the latest recommendation 

There's a quick start example in the readme https://index.ros.org/p/launch_pytest/#rolling

fuse_core/test/launch_tests/test_parameters.cpp
```diff
@@ -32,50 +32,75 @@
...
+  {
+    executor_->cancel();
+    if (spinner_.joinable()) {
+      spinner_.join();
+    }
+    executor_.reset();
+  }
+
+  std::thread spinner_;  //!< Internal thread for spinning the executor
+  std::atomic<bool> spinning_;  //!< Flag for spinning the spin thread
```
sloretz: nit, unused

fuse_core/test/launch_tests/test_parameters.cpp
```diff
@@ -135,100 +160,110 @@ TEST(Parameter, GetCovarianceDiagonalParam)
...
-      FAIL() << "Failed to get " << node_handle.resolveName(parameter_name) << ": " << ex.what();
-    }
+    ASSERT_FALSE(node->has_parameter(parameter_name));
+    EXPECT_THROW(fuse_core::getCovarianceDiagonalParam<Size>(node, parameter_name, default_variance),
+                 rclcpp::exceptions::InvalidParameterTypeException);
+
+    // NOTE(CH3): This test used to work this way, but doesn't anymore because of strongly typed
+    //            params in ROS2
+    //
+    // A covariance diagonal with invalid element type does not throw, but nothing it is loaded, so we should get the
```
sloretz: I don't think including the old source code is necessary. Just a comment about it how it is different from ROS 1 is fine.

Could we make it behave the same way in `fuse_core::GetCovarianceDiagonalParam` by catching the exception and returning the default instead? If not, we should check with @svwilliams that the new behavior is ok.

fuse_core/test/launch_tests/test_parameters.cpp
```diff
@@ -135,100 +160,110 @@ TEST(Parameter, GetCovarianceDiagonalParam)
...
```
sloretz: Same comment about old source code and checking this behavior with @svwilliams

fuse_core/test/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+
+
+# ROS TESTS (WITH LAUNCH) ==========================================================================
+find_package(ament_cmake_pytest REQUIRED)
+
+ament_add_gtest_executable(test_parameters launch_tests/test_parameters.cpp)
+target_link_libraries(test_parameters ${PROJECT_NAME})
+
+ament_add_pytest_test(test_parameters_py "launch_tests/test_parameters.py"
+  PYTHON_EXECUTABLE "${_PYTHON_EXECUTABLE}"
```
sloretz: Where is `_PYTHON_EXECUTABLE` set?  Sometimes we do odd things in CMake with the python executable because of Window Debug, but I'm not sure that's necessary here. [I don't see this parameter being used by `rclpy`](https://github.com/ros2/rclpy/blob/c7feb4b70c33aaf22e8ad467605c0a58451299c9/rclpy/CMakeLists.txt#L204-L209).

fuse_core/test/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+  WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}"
+)
+
+install(
+  DIRECTORY launch_tests
```
sloretz: I don't think tests should get installed

sloretz (commented):
A bunch of small comments! I still have 6 more files to review.

fuse_core/test/test_callback_wrapper.cpp
methylDragon: Incidentally, I was also wondering why the node API didn't include a `get_context()` method

fuse_core/include/fuse_core/parameter.h
```diff
@@ -46,45 +48,141 @@
...
+    node_interfaces::Logging,
+    node_interfaces::Parameters
+  > interfaces,
+  const std::string& key,
+  std::string& value
+){
+  std::string default_value = "";
+  value = getParam(interfaces, key, default_value);
+
+  if (value == default_value)
```
sloretz, line 150: What if the value was explicitly set to empty string?

This might have to use one of the `decalare_parameter()` overrides that doesn't accept a default value. It looks like the second `getParam()` above already implements that.

fuse_core/src/ceres_options.cpp
```diff
@@ -46,156 +46,518 @@
...
+    node_interfaces::Base,
+    node_interfaces::Logging,
+    node_interfaces::Parameters
+  > interfaces,
+  ceres::Covariance::Options& covariance_options,
+  const std::string& namespace_string)
 {
+  rcl_interfaces::msg::ParameterDescriptor tmp_descr;
+
+  std::string namespace_;
```
sloretz: nit, `namespace_` with a trailing underscore means "member variable" (in ROS 2 style via Google Style guide).

fuse_core/src/ceres_options.cpp
```diff
@@ -46,156 +46,518 @@
...
+  if (namespace_string.empty()) {
+    namespace_ = namespace_string;
+  } else {
+    namespace_ = namespace_string + ".";
+  }
+
 #if CERES_VERSION_AT_LEAST(1, 13, 0)
```
sloretz: [libceres-dev is version 1.13.0 in Bionic](https://packages.ubuntu.com/bionic/libceres-dev) - since we're targeting Rolling I think it's safe to remove this check.

fuse_core/src/ceres_options.cpp
```diff
@@ -46,156 +46,518 @@
...
-  nh.param("enable_fast_removal", problem_options.enable_fast_removal, problem_options.enable_fast_removal);
-  nh.param("disable_all_safety_checks", problem_options.disable_all_safety_checks,
-           problem_options.disable_all_safety_checks);
+  rcl_interfaces::msg::ParameterDescriptor tmp_descr;
+
+  std::string namespace_;
+  if (namespace_string.empty()) {
+    namespace_ = namespace_string;
+  } else {
+    namespace_ = namespace_string + ".";
```
sloretz: Two nits: check `namespace_string` does not already end in `.`, and make a function for this logic since it's reused a couple times in this file.

fuse_core/src/ceres_options.cpp
```diff
@@ -46,156 +46,518 @@
...
+  }
+
+  tmp_descr.description = "trades memory for faster Problem::RemoveResidualBlock()";
+  problem_options.enable_fast_removal = fuse_core::getParam(
+    interfaces,
+    namespace_ + "enable_fast_removal", problem_options.enable_fast_removal, tmp_descr
+  );
+
+  tmp_descr.description = "If true, trades memory for faster Problem::RemoveResidualBlock()";
```
sloretz: It looks like this description was copied from the above, and it seems different from upstream's description:

https://github.com/ceres-solver/ceres-solver/blob/352b320ab1b5438a0838aea09cbbf07fa4ff5d71/include/ceres/problem.h#L149-L158

fuse_core/src/ceres_options.cpp
```diff
@@ -46,156 +46,518 @@
...
+    fuse_core::declareCeresParam(interfaces, namespace_ + "minimizer_type", solver_options.minimizer_type);
+  solver_options.line_search_direction_type = fuse_core::declareCeresParam(
+    interfaces, namespace_ + "line_search_direction_type", solver_options.line_search_direction_type);
+  solver_options.line_search_type =
+    fuse_core::declareCeresParam(interfaces, namespace_ + "line_search_type", solver_options.line_search_type);
+  solver_options.nonlinear_conjugate_gradient_type = fuse_core::declareCeresParam(
+    interfaces, namespace_ + "nonlinear_conjugate_gradient_type",
+    solver_options.nonlinear_conjugate_gradient_type);
+
+  tmp_descr.description = "";
```
sloretz: Can get description from upstream source code: https://github.com/ceres-solver/ceres-solver/blob/352b320ab1b5438a0838aea09cbbf07fa4ff5d71/include/ceres/solver.h#L99-L123

But also, there weren't any descriptions before so I think it's fine to leave them blank for now.

fuse_core/src/ceres_options.cpp
```diff
@@ -46,156 +46,518 @@
...
+
+  tmp_descr.description = "Maximum amount of time for which the solver should run";
+  solver_options.max_solver_time_in_seconds = fuse_core::getParam(
+    interfaces,
+    namespace_ + "max_solver_time_in_seconds",
+    solver_options.max_solver_time_in_seconds,
+    tmp_descr
+  );
+
+  tmp_descr.description = "Maximum number of iterations for which the solver should run";
```
sloretz: seems incorrect for a parameter called "num_threads"

fuse_core/test/launch_tests/test_parameters.cpp
methylDragon: Small second ping to @svwilliams for comment on this. If there's no comment though, I'm for catching the exception and returning the default, I'll implement that.

fuse_core/include/fuse_core/parameter.h
methylDragon: Actually all of the declare_parameter signatures require a default value... The one with `value.get_type()` just uses the default constructed value of that type (which is an empty string)... I'm not sure how you'd distinguish between the two in this case :/

You have to declare a param before you can query it, and declaring it requires the input of a default value.

fuse_core/src/ceres_options.cpp
methylDragon: Ah, I was trying to get around the keyword `namespace`, I'll rename it to `ns` instead then.

fuse_core/test/launch_tests/test_parameters.cpp
svwilliams: `getCovarianceDiagonalParam()` would throw an exception if the number of elements was wrong. Throwing an exception if the element type is wrong seems like the correct action to me. The "ignore and load default" was really a limitation of ROS1 and not the desires behavior. I like the current implementation, throwing an invalid parameter type exception.

methylDragon: Alright, reverted!

fuse_core/src/ceres_options.cpp
methylDragon: I ended up filling some of them in (not all though)

methylDragon: Edit: I filled all of them except the ones with getCeresParam (because most of them don't have comments in the upstream source code)

fuse_core/test/test_throttled_callback.cpp
```diff
@@ -110,11 +117,29 @@ class PointSensorModel
...
+        std::bind(&PointSensorModel::dropCallback, this, std::placeholders::_1),
+        throttle_period
+      )
   {
-    subscriber_ = node_handle_.subscribe<geometry_msgs::Point>(
-        "point", 10, &PointThrottledCallback::callback, &throttled_callback_);
+    subscription_ = this->create_subscription<geometry_msgs::msg::Point>(
+      "point", 10,
+      [&](const geometry_msgs::msg::Point::ConstSharedPtr& msg) {
+        this->throttled_callback_.callback(msg);
```
sloretz: nit, I'd recommend calling [via `operator()`](https://github.com/locusrobotics/fuse/blob/9f9395e5146e8bf0bbeedd47dc953ec974ab4187/fuse_core/include/fuse_core/throttled_callback.h#L194-L197) since that's what the test [is using in ROS 1](https://github.com/locusrobotics/fuse/blob/0461f779f9ac4defca6ecfd102a675add92bb3a9/fuse_core/test/test_throttled_callback.cpp#L116).

sloretz (commented):
Most of the way through re-reviewing, only one comment so far!

fuse_core/test/test_throttled_callback.cpp
methylDragon: I had to use a lambda because ROS 2's subscribe function is missing a signature for conveniently passing method pointers of objects. (I remember that bind doesn't work nicely here, and there's a whole rabbit hole of arcane wizardry used to get it working in ROS 1)

ROS 1 [has such a function signature](http://docs.ros.org/en/indigo/api/roscpp/html/classros_1_1NodeHandle.html#a317fe4c05919e0bf3fb5162ccb2f7c28),

With the rabbit hole encompassing:
- [the original call](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/include/ros/node_handle.h#L464)
- [init](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/include/ros/subscribe_options.h#L103)

and inner factory functions.

I'm not too familiar with how ROS 1 does it to replicate it here (or upstream it for that matter.. :/)

methylDragon: Here's the log from me trying to bind it instead

<details>

```
/home/methyldragon/Desktop/Projects/2022_10__locus_fuse/fuse_ws/src/fuse/fuse_core/test/test_throttled_callback.cpp: In constructor ‘PointSensorModel::PointSensorModel(const rclcpp::Duration&)’:
/home/methyldragon/Desktop/Projects/2022_10__locus_fuse/fuse_ws/src/fuse/fuse_core/test/test_throttled_callback.cpp:128:29: error: no matching function for call to ‘bind(<unresolved overloaded function type>, PointSensorModel::PointThrottledCallback*, const std::_Placeholder<1>&)’
  128 |       "point", 10, std::bind(&PointThrottledCallback::callback, &throttled_callback_, std::placeholders::_1));
      |                    ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/methyldragon/Desktop/Projects/2022_10__locus_fuse/fuse_ws/src/fuse/fuse_core/include/fuse_core/throttled_callback.h:37,
                 from /home/methyldragon/Desktop/Projects/2022_10__locus_fuse/fuse_ws/src/fuse/fuse_core/test/test_throttled_callback.cpp:34:
/usr/include/c++/11/functional:789:5: note: candidate: ‘template<class _Func, class ... _BoundArgs> typename std::_Bind_helper<std::__is_socketlike<_Func>::value, _Func, _BoundArgs ...>::type std::bind(_Func&&, _BoundArgs&& ...)’
  789 |     bind(_Func&& __f, _BoundArgs&&... __args)
      |     ^~~~
/usr/include/c++/11/functional:789:5: note:   template argument deduction/substitution failed:
/home/methyldragon/Desktop/Projects/2022_10__locus_fuse/fuse_ws/src/fuse/fuse_core/test/test_throttled_callback.cpp:128:29: note:   couldn’t deduce template parameter ‘_Func’
  128 |       "point", 10, std::bind(&PointThrottledCallback::callback, &throttled_callback_, std::placeholders::_1));
      |                    ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/methyldragon/Desktop/Projects/2022_10__locus_fuse/fuse_ws/src/fuse/fuse_core/include/fuse_core/throttled_callback.h:37,
                 from /home/methyldragon/Desktop/Projects/2022_10__locus_fuse/fuse_ws/src/fuse/fuse_core/test/test_throttled_callback.cpp:34:
/usr/include/c++/11/functional:813:5: note: candidate: ‘template<class _Result, class _Func, class ... _BoundArgs> typename std::_Bindres_helper<_Result, _Func, _BoundArgs>::type std::bind(_Func&&, _BoundArgs&& ...)’
  813 |     bind(_Func&& __f, _BoundArgs&&... __args)
      |     ^~~~
/usr/include/c++/11/functional:813:5: note:   template argument deduction/substitution failed:
/home/methyldragon/Desktop/Projects/2022_10__locus_fuse/fuse_ws/src/fuse/fuse_core/test/test_throttled_callback.cpp:128:29: note:   couldn’t deduce template parameter ‘_Result’
  128 |       "point", 10, std::bind(&PointThrottledCallback::callback, &throttled_callback_, std::placeholders::_1));
      |                    ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
gmake[2]: *** [test/CMakeFiles/test_throttled_callback.dir/build.make:76: test/CMakeFiles/test_throttled_callback.dir/test_throttled_callback.cpp.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:678: test/CMakeFiles/test_throttled_callback.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
```
</details>

methylDragon: Edit: Seems like looking at it with fresh eyes worked :face_exhaling: 

https://github.com/locusrobotics/fuse/compare/d126a1e1b67585f04ae2727fde188f2ed5cd44e9..536743131a2705e900eec50eee3c8fa15fc5dd1e

fuse_core/test/launch_tests/test_parameters.py
```diff
@@ -0,0 +1,59 @@
...
+from launch_pytest.tools import process as process_tools
+
+from launch_ros.substitutions import FindPackageShare
+
+import pytest
+
+@pytest.fixture
+def test_proc():
+    test_root = os.path.join(
+        get_package_prefix('fuse_core'), '..', '..', 'build', 'fuse_core', 'test'
```
sloretz: It looks like this assumes the test is being run after being built with colcon, with the default name for the build directory. It also assumes the software has been installed. I would recommend not using `get_package_prefix()` at all. There are at least a couple good alternatives.

One alternative would be to set the [`WORKING_DIRECTORY` argument to `ament_add_pytest_test`](https://github.com/locusrobotics/fuse/pull/286/files#diff-8f0dbf48cc6fdbc4ec0aa1ed045b1a616e2d5d337bffadb59d75c2901173e74eR58) to `${CMAKE_CURRENT_SOURCE_DIR}` and use a relative path.

Another alternative would be to use [the `ENV` arugment to pass in the path to the parameters file](https://github.com/ament/ament_cmake/blob/d38014a000241c145904c4a16851cb5033d0286f/ament_cmake_pytest/cmake/ament_add_pytest_test.cmake#L41) as an environment variable and use that as the path here.

fuse_core/test/launch_tests/test_parameters.py
```diff
@@ -0,0 +1,59 @@
...
+
+@launch_pytest.fixture
+def generate_test_description(test_proc):
+    return LaunchDescription([test_proc, ReadyToTest()])
+
+@pytest.mark.launch(fixture=generate_test_description)
+def test_no_failed_gtests(test_proc, launch_context):
+    def validate_output(output):
+        assert '[  PASSED  ] 2 tests.' in output.splitlines(), (
+            'process never printed required passes')
```
sloretz: It looks like the test is returning `RUN_ALL_TESTS();`. I'd bet the exit code would be non-zero on failure, and so that could be the check here.

sloretz (changes requested):
Last two comments!

fuse_core/test/launch_tests/test_parameters.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+import unittest
```
sloretz: Nit unused import

fuse_core/test/launch_tests/test_parameters.py
```diff
@@ -0,0 +1,52 @@
...
+
+from ament_index_python.packages import get_package_prefix
```
sloretz: nit, unused import

fuse_core/test/launch_tests/test_parameters.py
```diff
@@ -0,0 +1,52 @@
...
+
+from launch import LaunchContext, LaunchDescription, LaunchService
```
sloretz: nit, only `LaunchDescription` is used

fuse_core/test/launch_tests/test_parameters.py
```diff
@@ -0,0 +1,52 @@
...
+from launch.actions import EmitEvent, ExecuteProcess
```
sloretz: nit, EmitEvent is unused

fuse_core/test/launch_tests/test_parameters.py
```diff
@@ -0,0 +1,52 @@
...
+from launch.events import Shutdown
```
sloretz: nit, Shutdown is unused.

fuse_core/test/launch_tests/test_parameters.py
```diff
@@ -0,0 +1,52 @@
...
+
+import launch_pytest
+from launch_pytest.actions import ReadyToTest
+from launch_pytest.tools import process as process_tools
+
+from launch_ros.substitutions import FindPackageShare
```
sloretz: Nit, unused

sloretz (commented):
The substance looks good to me, just nits about unused imports

sloretz (changes requested):
One last important change - the `package.xml` needs a `<test_depend>launch_pytest</test_depend>`

sloretz (approved):
LGTM!

All tests pass except cpplint, uncrustify, and xmllint. Approving since linting is to be fixed in the next PR.

PR #287, opened 2022-11-17
fuse -> ROS 2 fuse_loss: Port fuse_loss

methylDragon:

# Description

This PR ports the entirety of fuse_loss, including:
- CMakeLists and packages
- Source code
- Tests

It makes minor changes to fuse_core to support the loss changes.
Also note that some changes to fuse_model were made (that won't allow it to get built, but are put in for posterity.)

Other stuff like docs are in other PRs

# Notes
- This PR is branched off the branch from https://github.com/locusrobotics/fuse/pull/286
	- i.e. https://github.com/locusrobotics/fuse/pull/286 should be merged FIRST (otherwise merging this will merge 286 as well)

### DIFF
--- a/fuse_loss/CMakeLists.txt
+++ b/fuse_loss/CMakeLists.txt
@@ -1,36 +1,28 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.16)
 project(fuse_loss)
 
-set(build_depends
-  fuse_core
-  pluginlib
-  roscpp
-)
+# Default to C++17
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 17)
+  set(CXX_STANDARD_REQUIRED YES)
+endif()
 
-find_package(catkin REQUIRED COMPONENTS
-  ${build_depends}
-)
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Werror -Wpedantic)
+endif()
+
+find_package(ament_cmake REQUIRED)
+find_package(fuse_core REQUIRED)
+find_package(pluginlib REQUIRED)
 
 find_package(Ceres REQUIRED)
-
-catkin_package(
-  INCLUDE_DIRS
-    include
-  LIBRARIES
-    ${PROJECT_NAME}
-  CATKIN_DEPENDS
-    ${build_depends}
-  DEPENDS
-    CERES
-)
 
 ###########
 ## Build ##
 ###########
 
-add_compile_options(-Wall -Werror)
-
-add_library(${PROJECT_NAME}
+## fuse_loss library
+add_library(${PROJECT_NAME} SHARED
   src/arctan_loss.cpp
   src/cauchy_loss.cpp
   src/composed_loss.cpp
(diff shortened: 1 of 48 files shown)


fuse_loss/CMakeLists.txt
```diff
@@ -46,218 +38,52 @@ add_library(${PROJECT_NAME}
...
-install(DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+install(DIRECTORY include/
+  DESTINATION include/${PROJECT_NAME}
 )
 
 install(
   FILES fuse_plugins.xml
-  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
+  DESTINATION share/${PROJECT_NAME}
```
sloretz: Use `pluginlib_export_plugin_description_file()` https://github.com/ros/pluginlib/blob/rolling/pluginlib/cmake/pluginlib_export_plugin_description_file.cmake - it will both do this install and setup the right stuff in the `ament_index` for pluginlib to find the plugins.

fuse_loss/CMakeLists.txt
```diff
@@ -1,36 +1,28 @@
...
 
 ###########
 ## Build ##
 ###########
 
-add_compile_options(-Wall -Werror)
-
-add_library(${PROJECT_NAME}
+## fuse_loss library
+add_library(${PROJECT_NAME} SHARED
```
sloretz: If you intentionally leave off `SHARED` then this library can be built statically by setting `-DBUILD_SHARED_LIBS=OFF` in CMake.

This is one of those weird CMake features/conventions, but  It looks like it was being used here before since `SHARED` wasn't specified. `ament_cmake_ros` documents that option when it's `find_package()'d`, so change `find_package(ament_cmake REQUIRED)` to `find_package(ament_cmake_ros REQUIRED)` above too.

fuse_loss/cmake/FindQWT.cmake
```diff
@@ -38,75 +38,75 @@
...
 # handle the QUIETLY and REQUIRED arguments
-include ( FindPackageHandleStandardArgs )
-if ( CMAKE_VERSION LESS 2.8.3 )
-  find_package_handle_standard_args( Qwt DEFAULT_MSG QWT_LIBRARY QWT_INCLUDE_DIR _QWT_VERSION_MATCH )
-else ()
-  find_package_handle_standard_args( Qwt REQUIRED_VARS QWT_LIBRARY QWT_INCLUDE_DIR _QWT_VERSION_MATCH VERSION_VAR QWT_VERSION_STRING )
-endif ()
+include( FindPackageHandleStandardArgs )
+if( CMAKE_VERSION LESS 2.8.3 )
+  find_package_handle_standard_args( QWT DEFAULT_MSG QWT_LIBRARY QWT_INCLUDE_DIR _QWT_VERSION_MATCH )
```
sloretz, line 102: The capitalization might be important here. I think it would be the difference between `QWT_FOUND` vs `Qwt_FOUND` being set.

fuse_loss/test/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+target_link_libraries(test_scaled_loss ${PROJECT_NAME})
+
+# Qwt Loss Plot Tests
+option(BUILD_WITH_PLOT_TESTS "Build with plot tests. These test might fail to run in headless systems." OFF)
+if(BUILD_WITH_PLOT_TESTS)
+  find_package(Qt5 REQUIRED COMPONENTS Core Widgets)
+
+  # Find Qwt using FindQwt.cmake copied from:
+  # https://gitlab.kitware.com/cmake/community/-/wikis/contrib/modules/FindQwt
+  list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
```
sloretz: This path is incorrect since it was moved to the `test` directory.

sloretz (commented):
Mind fixing this compiler warning?

```
In file included from /home/osrf/ws/rolling/src/fuse/fuse_loss/test/test_qwt_loss_plot.cpp:46:
/home/osrf/ws/rolling/src/fuse/fuse_loss/include/fuse_loss/qwt_loss_plot.h: In constructor ‘fuse_loss::QwtLossPlot::QwtLossPlot(const std::vector<double, std::allocator<double> >&, const fuse_loss::HSVColormap&)’:
/home/osrf/ws/rolling/src/fuse/fuse_loss/include/fuse_loss/qwt_loss_plot.h:221:35: warning: ‘static QVector<T> QVector<T>::fromStdVector(const std::vector<BufferT>&) [with T = double]’ is deprecated: Use QVector<T>(vector.begin(), vector.end()) instead. [-Wdeprecated-declarations]
  221 |     : residuals_(QVector<double>::fromStdVector(residuals))
```


```
colcon build --packages-select fuse_loss --cmake-args -DBUILD_WITH_PLOT_TESTS=ON -DBUILD_WITH_INTERACTIVE_TESTS=ON
```

fuse_loss/cmake/FindQWT.cmake
methylDragon: I changed the capitalization so ament_lint_cmake would stop throwing warnings, is there a way to silence the linting for this? :o

fuse_loss/CMakeLists.txt
methylDragon: I'll also remove the SHARED on `fuse_core` in this PR too. It was added by me in a previous PR.

fuse_loss/cmake/FindQWT.cmake
methylDragon: According to the file itself, the variable that's being set is
```
#  QWT_FOUND - the system has Qwt
```

So I think we can change it? The error's coming from the CMAKE_MODULE_PATH call.
I've updated it to:
`  list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../cmake")`

methylDragon:
Addressed!
I also tried to roll in the same sort of changes requested in fuse_core: https://github.com/locusrobotics/fuse/pull/287/commits/a211ec68c28af288348d8b75a6d837fc9d64d07a

sloretz:

fuse_loss/package.xml
```diff
@@ -11,19 +11,26 @@
   <author email="swilliams@locusrobotics.com">Stephen Williams</author>
   <license>BSD</license>
 
-  <buildtool_depend>catkin</buildtool_depend>
+  <buildtool_depend>ament_cmake</buildtool_depend>
```
sloretz: nit [`ament_cmake_ros` instead of `ament_cmake`](https://github.com/locusrobotics/fuse/pull/287/files#diff-c18b20aa084503537399b799cf05fbe3443eff62b48cff23c3c17ecec57fc117R14).

sloretz (commented):
LGTM with one small fix

All tests pass locally except cpplint, uncrustify, and copyright, but that's expected to be fixed in the next PR.

fuse_loss/package.xml
methylDragon: https://github.com/locusrobotics/fuse/compare/25e6567ae3925ff56f5db18a4ea8d3dbb8e320f1..7e4da3d38f24cb4f39f0536c33531388fb297692

sloretz (approved):

PR #292, opened 2022-12-01
fuse -> ROS 2 fuse_core: Linting

methylDragon:
See: #276

# Description

- Fixes linting. Nuff said.
- Also migrates `.h` files to `.hpp` files!

Lots of files got changed!

It's a wrap for fuse_core (at least until we need to circle back for waitables)


fuse_doc/doc/.templates/full_globaltoc.html
```diff
@@ -6,7 +6,7 @@ <h3><a href="{{ pathto(master_doc) }}">{{ _('Further Documentation') }}</a></h3>
   <li class="toctree-l1">API Documentation</li>
   <ul>
     <li class="toctree-l1"><a href="../../fuse_constraints/api/index.html"/>fuse_constraints</a></li>
-    <li class="toctree-l1"><a href="../../fuse_core/api/index.html"/>fuse_core</a></li>
```
sloretz: This one looks like an accidental find/replace

fuse_core/include/fuse_core/node_interfaces/node_interfaces.h
```diff
@@ -0,0 +1,43 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef FUSE_CORE__NODE_INTERFACES__NODE_INTERFACES_H_
+#define FUSE_CORE__NODE_INTERFACES__NODE_INTERFACES_H_
+
+#warning \
+  This header is obsolete, please include fuse_core/node_interfaces/node_interfaces.hpp instead
+
+#include <fuse_core/node_interfaces/node_interfaces.hpp>
```
sloretz: I don't think we need a `.h` migration header for this one since we just added the file in the last PR. There isn't any downstream code including it that would benefit.

fuse_core/include/fuse_core/node_interfaces/node_interfaces_helpers.h
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+#ifndef FUSE_CORE__NODE_INTERFACES__NODE_INTERFACES_HELPERS_H_
+#define FUSE_CORE__NODE_INTERFACES__NODE_INTERFACES_HELPERS_H_
+
+#warning \
+  This header is obsolete, please include fuse_core/node_interfaces/node_interfaces_helpers.hpp \
+  instead
+
+#include <fuse_core/node_interfaces/node_interfaces_helpers.hpp>
```
sloretz: Same comment about not needing a `.h` migration header for this one

fuse_core/include/fuse_core/time.h
```diff
@@ -32,85 +32,11 @@
...
- *
- * @return valid  true if clock was or became valid
- */
-// TOOD(CH3): Replace with rclcpp's implementation when https://github.com/ros2/rclcpp/pull/2040 is in
-bool wait_for_valid(
-  rclcpp::Clock::SharedPtr clock,
-  const rclcpp::Duration & timeout,
-  rclcpp::Context::SharedPtr context = rclcpp::contexts::get_global_default_context(),
-  const rclcpp::Duration & wait_tick_ns = rclcpp::Duration(0, static_cast<uint32_t>(1e7)));
+#warning This header is obsolete, please include fuse_core/time.hpp instead
```
sloretz: Same comment about not needing a migration header for this one since we just added it.

fuse_core/test/test_timestamp_manager.cpp
```diff
@@ -77,25 +76,23 @@ class TimestampManagerTestFixture : public ::testing::Test
...
 
 
 TEST_F(TimestampManagerTestFixture, Empty)
 {
-  // Test:
-  // Existing: |-------------------------------------> t
-  // Adding:   |------********-----------------------> t
-  // Expected: |------********-----------------------> t
+  // Test: Existing: |-------------------------------------> t Adding:
+  // |------********-----------------------> t Expected: |------********-----------------------> t
```
sloretz: I think these comments got a little messed up in the reformatting. The newlines seem to be intentional for showing a timeline

sloretz: There are a bunch more places like this in this file below.

fuse_core/test/test_local_parameterization.cpp
```diff
@@ -89,8 +88,8 @@ TEST(LocalParameterization, PlusJacobian)
 
   fuse_core::MatrixXd expected(3, 2);
   expected << 2.0, 0.0,
-              0.0, 5.0,
-              0.0, 0.0;
+    0.0, 5.0,
+    0.0, 0.0;
```
sloretz: I think the formatting was intentional before to line up the numbers in the matrix

fuse_core/test/test_local_parameterization.cpp
```diff
@@ -120,7 +119,7 @@ TEST(LocalParameterization, MinusJacobian)
 
   fuse_core::MatrixXd expected(2, 3);
   expected << 0.5, 0.0, 0.0,
-              0.0, 0.2, 0.0;
+    0.0, 0.2, 0.0;
```
sloretz: Same comment about the formatting being intentional here

fuse_core/src/graph_deserializer.cpp
```diff
@@ -31,61 +31,58 @@
...
   constraint_loader_("fuse_core", "fuse_core::Constraint"),
   loss_loader_("fuse_core", "fuse_core::Loss"),
   graph_loader_("fuse_core", "fuse_core::Graph")
 {
-  // Load all known plugin libraries
-  // I believe the library containing a given Variable or Constraint type must be loaded in order to deserialize
-  // an object of that type. But I haven't actually tested that theory.
-  for (const auto& class_name : variable_loader_.getDeclaredClasses())
-  {
+  // Load all known plugin libraries I believe the library containing a given Variable or Constraint
```
sloretz: Maybe add a period after `Load all known plugin libraries`

fuse_core/src/transaction_deserializer.cpp
```diff
@@ -31,60 +31,60 @@
...
+: variable_loader_("fuse_core", "fuse_core::Variable"),
   constraint_loader_("fuse_core", "fuse_core::Constraint"),
   loss_loader_("fuse_core", "fuse_core::Loss")
 {
-  // Load all known plugin libraries
-  // I believe the library containing a given Variable or Constraint must be loaded in order to deserialize
-  // an object of that type. But I haven't actually tested that theory.
-  for (const auto& class_name : variable_loader_.getDeclaredClasses())
-  {
+  // Load all known plugin libraries I believe the library containing a given Variable or Constraint
```
sloretz: same comment about adding a period here.

fuse_core/include/fuse_core/autodiff_local_parameterization.hpp
```diff
@@ -0,0 +1,242 @@
...
+/**
+ * @brief Create a local parameterization with the Jacobians computed via automatic
+ *        differentiation.
+ *
+ * To get an auto differentiated local parameterization, you must define two classes with a
+ * templated operator() (a.k.a. a functor).
+ *
+ * The first functor should compute:
+ *
+ * Plus(x, delta) -> x_plus_delta
```
sloretz: I think this line being indented before was intentional

fuse_core/include/fuse_core/autodiff_local_parameterization.hpp
```diff
@@ -0,0 +1,242 @@
...
+ *
+ * And the second functor should compute the inverse operation:
+ *
+ * Minus(x1, x2) -> delta
```
sloretz: Same comment about how the indentation here before was intentional

fuse_core/include/fuse_core/ceres_macros.hpp
```diff
@@ -0,0 +1,46 @@
...
+#ifndef FUSE_CORE__CERES_MACROS_HPP_
+#define FUSE_CORE__CERES_MACROS_HPP_
+
+#include <ceres/version.h>
+
+/**
+ * Check for at least Ceres Solver version x.y.z, where: x = major, y = minor and z = revision.
+ */
+#define CERES_VERSION_AT_LEAST(x, y, z) (CERES_VERSION_MAJOR > x || (CERES_VERSION_MAJOR >= x && \
+  (CERES_VERSION_MINOR > y || (CERES_VERSION_MINOR >= y && \
```
sloretz: Indentation here before looked intentional. Would the linter be unhappy if we restored it?

fuse_core/include/fuse_core/console.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+    return false;
+  }
+
+  /**
+   * @brief Rests the last time the filter condition was hit
+   */
+  void reset()
+  {
+    last_hit_ = time_point_cast<milliseconds>(std::chrono::system_clock::from_time_t(-1));
```
sloretz: Is there a test covering this code? I would have expected an error because of `milliseconds` instead of `std::chrono::milliseconds` (same above).

fuse_core/include/fuse_core/graph.hpp
```diff
@@ -0,0 +1,594 @@
...
+   * @brief Evalute the values of the current set of variables, given the current set of
+   *        constraints.
+   *
+   * The values in the graph do not change after the call.
+   *
+   * If any of the output arguments is nullptr, it will not be evaluated. This mimics the
+   * ceres::Problem::Evaluate method API. Here all output arguments default to nullptr except
+   * for the cost.
+   *
+   * TODO(efernandez) support jacobian output argument The jacobian output argument is not
```
sloretz: I think a period should be added here too

fuse_core/include/fuse_core/message_buffer_impl.hpp
```diff
@@ -150,23 +150,23 @@ void MessageBuffer<Message>::purgeHistory()
...
-  auto is_greater = [](const auto& stamp, const auto& element) -> bool
-  {
-    return (element.first > stamp);
-  };
-  auto expiration_iter = std::upper_bound(buffer_.begin(), buffer_.end(), expiration_time, is_greater);
-  if (expiration_iter != buffer_.begin())
-  {
-    // expiration_iter points to the first element > expiration_time.
-    // Back up one entry, to a point that is <= expiration_time
+  // Remove buffer elements before the expiration time. Be careful to ensure that:  - at least two
```
sloretz: I think the newlines for the list were intentional here

fuse_core/include/fuse_core/parameter.hpp
```diff
@@ -0,0 +1,289 @@
...
+ *
+ * Declare a parameter if not declared, otherwise, get its value.
+ *
+ * This is needed because the node parameters interface does not do the type conversions to and
+ * from ParameterValue for us.
+ *
+ * @param[in] interfaces - The node interfaces used to load the parameter
+ * @param[in] parameter_name - The ROS parameter name
+ * @param[in] parameter_descriptor - An optional, custom description for the parameter.
+ * @param[in] When `true`, the parameter override is ignored. Default to `false`. \return The
```
sloretz: I think this line is missing the parameter name `ignore_override`

fuse_core/include/fuse_core/parameter.hpp
```diff
@@ -0,0 +1,289 @@
...
+ *                  value of the parameter. @throws if the parameter has already been declared
```
sloretz: `@throws` should probably be a separate line

fuse_core/include/fuse_core/parameter.hpp
```diff
@@ -0,0 +1,289 @@
...
+    throw rclcpp::exceptions::InvalidParameterTypeException(parameter_name, ex.what());
+  }
+}
+
+/**
+ * @brief Utility method for handling required ROS params
+ *
+ * @param[in] interfaces - The node interfaces used to load the parameter
+ * @param[in] key - The ROS parameter key for the required parameter
+ * @param[out] value - The ROS parameter value for the \p key @throws std::runtime_error if the
```
sloretz: same comment about `@throws` - it might be worth grepping for `@throws` to double check them if a script was used to change this one

fuse_core/include/fuse_core/transaction.hpp
```diff
@@ -0,0 +1,331 @@
...
+ */
+#ifndef FUSE_CORE__TRANSACTION_HPP_
+#define FUSE_CORE__TRANSACTION_HPP_
+
+#include <iostream>
+#include <ostream>
+#include <set>
+#include <vector>
+
+#include <boost/archive/binary_oarchive.hpp>
```
sloretz, line 42: Is this include unused? It seems to be new in this PR but none of the `boost::` things in this file look like they come from it.

fuse_core/include/fuse_core/util.hpp
```diff
@@ -0,0 +1,154 @@
...
+ * @param[in/out] angle Input angle to be wrapped to the [-Pi, +Pi) range. Angle is updated by
+ *                      this function.
+ */
+template<typename T>
+void wrapAngle2D(T & angle)
+{
+  // Define some necessary variations of PI with the correct type (double or Jet)
+  static const T PI = T(M_PI);
+  static const T TAU = T(2 * M_PI);
+  // Handle the 1*Tau roll-over (https://tauday.com/tau-manifesto) Use ceres::floor because it is
```
sloretz: Recommend period or newline here

sloretz (changes requested):

fuse_core/include/fuse_core/parameter.hpp
methylDragon: Ah, an edge case I did not account for in the script, now accounted for

fuse_core/include/fuse_core/transaction.hpp
methylDragon: Because of the include reordering, I needed to add this in, otherwise one of the boost serialization libraries complains about a missing method.

(The issue used to be masked by fuse_core/serialization.hpp being included first, but that can't happen without making cpplint unhappy.)

fuse_doc/doc/.templates/full_globaltoc.html
methylDragon: Whoa

fuse_core/test/test_local_parameterization.cpp
methylDragon: Hmm, I think this edge case is going to have to stay (for manual fixing...)
I'm not sure how to implement logic to parse this case

I'll manually fix the issues as they come up in reviews (and in my own quick review passes before pushing also)

methylDragon: Oh actually this was done by uncrustify reformat :OOOOOOOO

methylDragon: I'm going to need to use a bypassing directive for uncrustify. I'll add it in and fix the formatting

sloretz: +1 to manually fixing instead of trying to make automation do 100%

fuse_core/include/fuse_core/transaction.hpp
sloretz: Ah got it. Sounds like a bug in Boost then.

fuse_core/src/graph_deserializer.cpp
methylDragon: This is an edge case of my script with greedy wrapping. I'll edit it and have the script only start wrapping once it encounters a block that's long (which means short lines prior to those won't get smashed together)

There were originally newlines, I'll preserve them (instead of adding a period AND the original newline back)

fuse_core/include/fuse_core/autodiff_local_parameterization.hpp
methylDragon: Script has been updated to now catch this edge-case. But I'll have to manually fix these in this package.

fuse_core/include/fuse_core/ceres_macros.hpp
methylDragon: This is uncrustify automatically doing it. I'll add a bypass directive.

fuse_core/include/fuse_core/console.hpp
methylDragon: That's strange. Something like this shouldn't compile, yet it compiles fine locally.. Maybe a dirty workspace?

Ah, it never gets used in fuse_core, only in fuse_publishers (currently `COLCON_IGNORED`, so the error would only come up then.

Nice catch!

sloretz (approved):
LGTM and all tests in fuse_core pass locally using Rolling deps from the testing repo on Jammy

fuse_core/include/fuse_core/transaction.hpp
svwilliams: Sorry. I'm behind on my email.

I just want to note that the boost serialization system can be particular about when/where includes happen. Specifically the input/output archives must be included before any BOOST_CLASS_EXPORT macros are used.

https://www.boost.org/doc/libs/1_71_0/libs/serialization/doc/special.html#export
> BOOST_CLASS_EXPORT in the same source module that includes any of the archive class headers will instantiate code required to serialize polymorphic pointers of the indicated type to the all those archive classes. If no archive class headers are included, then no code will be instantiated.
>
> Note that the implemenation of this functionality requires that the BOOST_CLASS_EXPORT macro appear after the inclusion of any archive class headers for which code is to be instantiated. So, code that uses BOOST_CLASS_EXPORT will look like the following:
>
>```
> #include <boost/archive/text_oarchive.hpp>
> #include <boost/archive/text_oarchive.hpp>
> ... // other archives
> 
> #include "a.hpp" // header declaration for class a
> BOOST_CLASS_EXPORT(a)
> ... // other class headers and exports
> ```

PR #293, opened 2022-12-03
fuse -> ROS 2 fuse_core: Fix Async

methylDragon:

# Description
There are several outstanding issues with the async model used in fuse_core:
1. There's a chance for the destructor to get called before the internal spinner thread reaches the spin() call, causing the executor to mask the `executor->cancel()` call and spin forever.
2. If a new callback is added before the internal spinner attaches the waitable, the executor only wakes up once when there are actually two waitables to execute. That is, despite two guard condition triggers for two waitables, the executor only services one waitable.
3. For some reason, using an internal executor with 2 or more threads causes the waitables to never get picked up. I suspect this is because of (2). That is, because of (2) the waitable that adds the callback queue never gets served (the guard condition never gets added to the waitset).

I'm a little iffy on if that's what's happening with (2), but regardless, the solutions implemented in this PR fix both issues.

The solution is to keep triggering the guard condition on hooking of the callback queues until the callback queues are serviced.

I also edited the tests to be a ~~little bit~~ a lot stricter and now test for multiple callback additions and initializations.
The tests now feature async classes with executor thread count of 0 (which means the number of threads will equal the number of CPU cores), as troublesome a case as possible to test, to ensure my solutions are as robust as possible.

I had to do this because a lot of the issues seem to be race condition related and come up extremely rarely.

# Notes
There was an alternative solution that I explored (which also worked), which was to use the [timeout argument](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/src/rclcpp/executors/multi_threaded_executor.cpp#L33) for the multi-threaded executor. This removes any chance of the multi-threaded executor deadlocking on waitables, [but at the cost of free spinning at the rate of the timeout](https://github.com/ros2/rclcpp/blob/d119157948720f5888d47898e1c59b56fe1f86c5/rclcpp/src/rclcpp/executor.cpp#L912).

If we encounter deadlock or threading issues down the line, we might have to use that solution instead.

- Sometimes the initialize still deadlocks (and we're stuck in the waiting for the `initialized_` atomic to become true. I can't reproduce it though... Is there a chance for callbacks to get dropped??
- I'm also kinda iffy about all the random sleep_for dotted around the init, but it's the only way to avoid the warning printouts most of the time

# Update
EDIT: I pulled out all the stops and used all the solutions I found, all at once. I do not think I understand the problem enough to come up with an elegant solution, as for some reason all the "good practices" (like locking on threads that notify condition variables), or relying on condition variable waits instead of sleeps all result in deadlocks or weird behavior like atomic flags not getting set when the block that sets them ostensibly runs (???).

The solution I've devised combines the use of condition variables and manual guard condition triggers (just for `initialize()`!!), and also a lower frequency executor wakeup.

- The manual guard condition triggers should resolve the problem where the executor misses either the first callback, or the callback that adds the callback_queue to the node's waitables
- The lower frequency executor wakeup then resolves the issue where for some reason the initialization callback doesn't run
- Finally,
  - I had to remove the lock acquisition on the initialization callback because for some reason it causes a deadlock
  - And I had to add a small sleep because otherwise sometimes the atomic doesn't get set even though the callback gets run (????)
  - And I had to throttle the tests a tiny bit, because again, for some reason, spinning up and shutting down multithreaded executors too rapidly (**sequentially**) causes something to break (???)

On the other hand, I'm able to spin up and down a massive amount of async instances sequentially now, when before it'd randomly block indefinitely, so it might just be good enough (I updated the tests to check for that case specifically.)? I have yet to see it fail, but I am not confident that it will never fail, since I never understood why it failed in the first place.

Some refinement might be helpful though, maybe I'm just not seeing something obvious.

### DIFF
--- a/fuse_core/src/callback_wrapper.cpp
+++ b/fuse_core/src/callback_wrapper.cpp
@@ -112,11 +112,15 @@ void CallbackAdapter::addCallback(const std::shared_ptr<CallbackWrapperBase> & c
 {
   std::lock_guard<std::mutex> lock(queue_mutex_);
   callback_queue_.push_back(callback);
+
+  // NOTE(CH3): Do we need to keep triggering guard conditions if the queue size is > 1?
+  //            Is there a chance that the queue grows because multiple guard condition triggers
+  //            happen in the time the executor takes to pull one callback off the queue?
   if (RCL_RET_OK != rcl_trigger_guard_condition(&gc_)) {
     RCLCPP_WARN(
       rclcpp::get_logger("fuse"),
       "Could not trigger guard condition for callback, pulling callback off the queue.");
-    callback_queue_.pop_back();    // Undo
+    callback_queue_.pop_back();  // Undo
   }
 }
 
@@ -124,11 +128,15 @@ void CallbackAdapter::addCallback(std::shared_ptr<CallbackWrapperBase> && callba
 {
   std::lock_guard<std::mutex> lock(queue_mutex_);
   callback_queue_.push_back(std::move(callback));
+
+  // NOTE(CH3): Do we need to keep triggering guard conditions if the queue size is > 1?
+  //            Is there a chance that the queue grows because multiple guard condition triggers
+  //            happen in the time the executor takes to pull one callback off the queue?
   if (RCL_RET_OK != rcl_trigger_guard_condition(&gc_)) {
     RCLCPP_WARN(
       rclcpp::get_logger("fuse"),
       "Could not trigger guard condition for callback, pulling callback off the queue.");
-    callback_queue_.pop_back();    // Undo
+    callback_queue_.pop_back();  // Undo
   }
 }
 
@@ -138,5 +146,13 @@ void CallbackAdapter::removeAllCallbacks()
   callback_queue_.clear();
 }
 
+void CallbackAdapter::triggerGuardCondition()
+{
+  if (RCL_RET_OK != rcl_trigger_guard_condition(&gc_)) {
+    RCLCPP_WARN(
+      rclcpp::get_logger("fuse"),
+      "Could not trigger guard condition for callback.");
+  }
+}
 
 }  // namespace fuse_core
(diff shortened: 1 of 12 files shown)


methylDragon:

methylDragon:
CI's green: https://build.ros2.org/job/Rpr__fuse__ubuntu_jammy_amd64/8/

methylDragon:

methylDragon:
https://build.ros2.org/job/Rpr__fuse__ubuntu_jammy_amd64/9/

methylDragon:

methylDragon:
https://build.ros2.org/job/Rpr__fuse__ubuntu_jammy_amd64/11

methylDragon:

methylDragon:
https://build.ros2.org/job/Rpr__fuse__ubuntu_jammy_amd64/13

methylDragon:
[Build Status](https://build.ros2.org/job/Rpr__fuse__ubuntu_jammy_amd64/)

(The only tests that should fail are the fuse_loss linters)

fuse_core/src/callback_wrapper.cpp
```diff
@@ -112,23 +112,31 @@ void CallbackAdapter::addCallback(const std::shared_ptr<CallbackWrapperBase> & c
 {
   std::lock_guard<std::mutex> lock(queue_mutex_);
   callback_queue_.push_back(callback);
+
+  // NOTE(CH3): Do we need to keep triggering guard conditions if the queue size is > 1?
```
sloretz, line 116: I'm pretty sure the answer is yes, and this is a good catch.

 IIRC the waitable interface assumes that the waitable will become ready by something that will wake the wait set. Guard conditions wake the wait set, but they only wake once no matter how many times they're triggered.

I think a good solution would be to trigger the guard condition in `take_data()`. If after taking data the queue is not emtpy, then trigger the guard condition again so the executor will wake for the next callback in the queue.

methylDragon:
Closing in favor of https://github.com/locusrobotics/fuse/pull/294

PR #294, opened 2022-12-05
fuse -> ROS 2 fuse_core: Fix Async (Redux)

sloretz:
@methylDragon Here's an alternative to #293. I'm running the `test_async_sensor_model` test locally with `--retest-until-fail 1000`. I think this addresses the two issues found so far.

@svwilliams for visibility

### DIFF
--- a/fuse_core/src/async_motion_model.cpp
+++ b/fuse_core/src/async_motion_model.cpp
@@ -54,10 +54,7 @@ AsyncMotionModel::AsyncMotionModel(size_t thread_count)
 
 AsyncMotionModel::~AsyncMotionModel()
 {
-  executor_->cancel();
-  if (spinner_.joinable()) {
-    spinner_.join();
-  }
+  internal_stop();
 }
 
 bool AsyncMotionModel::apply(Transaction & transaction)
@@ -96,18 +93,21 @@ void AsyncMotionModel::initialize(const std::string & name)
   executor_options.context = ros_context;
   executor_ = rclcpp::executors::MultiThreadedExecutor::make_shared(
     executor_options,
-    executor_thread_count_);
+    executor_thread_count_
+  );
 
   callback_queue_ = std::make_shared<CallbackAdapter>(ros_context);
+
+  // This callback group MUST be re-entrant in order to support parallelization
+  cb_group_ = node_->create_callback_group(rclcpp::CallbackGroupType::Reentrant);
   node_->get_node_waitables_interface()->add_waitable(
-    callback_queue_, (rclcpp::CallbackGroup::SharedPtr) nullptr);
+    callback_queue_, cb_group_);
 
   // Call the derived onInit() function to perform implementation-specific initialization
   onInit();
 
+  // Start the executor to service the local callback queue
   executor_->add_node(node_);
-
-  // TODO(CH3): Remove this if the internal node is removed
   spinner_ = std::thread(
     [&]() {
       executor_->spin();
@@ -136,7 +136,12 @@ void AsyncMotionModel::start()
 
 void AsyncMotionModel::stop()
 {
+  // Optimizers can sometimes ask the models to stop -> start without initializing.
+  // So we only fully wrap up if the context is no longer valid.
+  // Otherwise, we only reset the internal state by clearing the callback queue and calling onStop()
   if (node_->get_node_base_interface()->get_context()->is_valid()) {
+    callback_queue_->removeAllCallbacks();
+
     auto callback = std::make_shared<CallbackWrapper<void>>(
       std::bind(&AsyncMotionModel::onStop, this)
     );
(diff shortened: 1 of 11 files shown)


fuse_core/src/callback_wrapper.cpp
```diff
@@ -92,6 +92,13 @@ std::shared_ptr<void> CallbackAdapter::take_data()
       cb_wrapper = callback_queue_.front();
       callback_queue_.pop_front();
     }
+    if (!callback_queue_.empty()) {
```
methylDragon, line 95: Should this be a while loop instead, or does it "chain"?

sloretz: It will chain, and that's exactly the purpose in putting it in "take_data". Triggering the guard condition here will cause the executor to wake immediately the next time it goes to `rcl_wait()`, see that the waitable is ready, and "take" data from it again.

methylDragon: Ahh, gotcha

methylDragon:
I've updated the PR to include the following changes:
- Add an `initialized_` flag that prevents an async class from getting initialized twice (with the appropriate tests)
- Change the tests to test with multithreaded executors (and made the init test more strict)
- Fix async issues by using reentrant callback groups on top of the chaining `take_data()`
- Update docstrings

We uncovered a potential bug in rclcpp's executors' handling of mutually exclusive callback groups, but in this case it doesn't block the progress for fuse_core

methylDragon:

methylDragon (dismissed):
Approving, but double checking with @sloretz that my changes are alright

sloretz:
I made a few changes after self-reviewing and reading [this bit of fuse_tutorials](https://github.com/locusrobotics/fuse/blob/0461f779f9ac4defca6ecfd102a675add92bb3a9/fuse_tutorials/include/fuse_tutorials/range_sensor_model.h#L66-L84).

* In ROS 1 - [start](https://github.com/locusrobotics/fuse/blob/0461f779f9ac4defca6ecfd102a675add92bb3a9/fuse_core/src/async_motion_model.cpp#L95-L101) and [stop](https://github.com/locusrobotics/fuse/blob/0461f779f9ac4defca6ecfd102a675add92bb3a9/fuse_core/src/async_motion_model.cpp#L107-L110) are scheduling `onStart` and `onStop` and nothing else. I updated this PR to make sure the ROS 2 port behaves the same (meaning I removed the bits where we cleared the callback queue when the `Context` was still valid)
* There's an edge case in `stop()` that was fixed by #182. When ROS is shutdown the `AsyncSpinner` won't call anymore callbacks, so the `onStop` method won't get called. That's why there's a special case where `onStop` is called directly, and it also opportunistically took the opportunity to clean up the `AsyncSpinner` (though IIUC it doesn't have to do that). The ROS 2 executor has the same behavior of no longer servicing callbacks when the `Context` shuts down, so I made the logic the same as it was before.
* I fixed the race condition between `initialize` and destruction by making `initialize()` block until the spinner thread starts. This made `internal_stop()` simpler.

sloretz:

methylDragon (changes requested):
I realized we should have `initialized_ = false` inside `internal_stop` so the user can choose to reinitialize the model

Aside from that, minor comments to consider

methylDragon:

fuse_core/src/async_motion_model.cpp
```diff
@@ -142,42 +153,32 @@ void AsyncMotionModel::start()
 
 void AsyncMotionModel::stop()
 {
-  // Optimizers can sometimes ask the models to stop -> start without initializing.
-  // So we only fully wrap up if the context is no longer valid.
-  // Otherwise, we only reset the internal state by clearing the callback queue and calling onStop()
+  // Prefer to call onStop in executor's thread so downstream users don't have
+  // to worry about threads in ROS callbacks when there's only 1 thread.
```
methylDragon, line 157: If we call `onStop` in the executor's thread, I think we should block with `wait()` in the calling thread, otherwise the calling thread might carry on, expecting onStop to have been called.

Alternatively, maybe we could return the future so the user can decide if they want to wait on it or not?

Since these are base classes users are expected to derive, biasing for flexibility might be the better choice?

methylDragon: Another reason for blocking is that there is a possible race condition where if:
- Stop gets called, then start gets called
  - e.g. https://github.com/sloretz/fuse/blob/sloretz__maybe_fix_async/fuse_optimizers/src/fixed_lag_smoother.cpp#L436

Then, in the multithreaded case, even though the callback queue is a queue, certain `onStop()` invocations might still be resolving when a later `onStart()` gets popped off in a different thread, potentially causing the `onStart()` to finish resolving first.

(That said, there are [comments in the code](https://github.com/locusrobotics/fuse/pull/294/files#diff-5c9e60d5461423c03e8c4c87d9e334f2d66471e815c6c68aa4477cef2dc21c70R125) that tell users to lock appropriately when using a multi-threaded executor, so we don't necessarily need to guard against this case. Though I'm not sure if that comment necessarily pre-empts the case where a `onStart()` might resolve before an `onStop()` resolves (the user would need to guard both invocations in that case.))

sloretz: > If we call onStop in the executor's thread, I think we should block with wait() in the calling thread, otherwise the calling thread might carry on, expecting onStop to have been called.

I think the code is already doing that. It gets the future for the `onStop` callback and waits on it in the calling thread. https://github.com/locusrobotics/fuse/pull/294/files#diff-327f5818028beb79e2c394523234fdeed441dd8d133721c7287273d05d58c892R162-R164

methylDragon: Whoops.. I was looking at the graph callback wait() removal :man_facepalming: My bad!

methylDragon (approved):
Let's add an `initialized_ = false` in the internal stops, other than that, LGTM!

sloretz:
> Let's add an initialized_ = false in the internal stops

Added in 9c95cf6

sloretz:

PR #299, opened 2022-12-15
fuse -> ROS 2 fuse_publishers: Port fuse_publishers

methylDragon:

# Description
This PR ports the entirety of fuse_publishers, including:

- CMakeLists and packages
- Source code
- Tests
- Other stuff like docs are in other PRs.

---

Notably, the AsyncPublisher class in fuse_core was updated (together with its tests)
I also changed the signature of its constructor to make use of `initialize` to set executor threads as well as pass it the appropriate NodeInterfaces.

In order to avoid specifying all manner of node interfaces in the AsyncPublisher class, I made it so that the derived publisher classes shadow their internal interface variable (and also call the parent's initialize method).

TODOs were left in downstream packages (models and tutorials) to update the usage of AsyncPublisher just like how it was done in fuse_publishers.

---

Furthermore, there's a missing signature in `tf2_ros::Buffer` that allows for buffer construction with node-like types. I left a TODO. I don't think it's too important, since it's only for logging/debugging with service calls.

Also, I realized I could just declare parameters instead of loading them in via yaml files, so I changed around some of the tests to reflect that...

# Notes


methylDragon:
@ros-pull-request-builder run tests please!

[Build Status](https://build.ros2.org/job/Rpr__fuse__ubuntu_jammy_amd64/33/)

methylDragon:
@ros-pull-request-builder run tests please!

methylDragon:
@ros-pull-request-builder run tests please!

methylDragon:
@ros-pull-request-builder run tests please!

methylDragon:
@ros-pull-request-builder run tests please!

methylDragon:
@ros-pull-request-builder run tests please!
[Build Status](https://build.ros2.org/job/Rpr__fuse__ubuntu_jammy_amd64/36/)

Only instabilities are linters!

fuse_core/include/fuse_core/publisher.hpp
```diff
@@ -82,7 +82,8 @@ class Publisher
    *
    * @param[in] name A unique name to give this plugin instance
    */
-  virtual void initialize(const std::string & name) = 0;
+  template<typename NodeT>
+  void initialize(NodeT interfaces, const std::string & name, size_t thread_count = 1);
```
sloretz: I'd recommend keeping `thread_count` on the `AsyncPublisher` constructor since the thread count option might not make sense for all `fuse_core::Publisher`s. If I understand correctly it only makes sense for the async one.

This SO post says overriding methods don't inherit the default value, so we'd have to duplicate `thread_count = 1` in all derived classes if it was left on this virtual method: https://stackoverflow.com/a/3533692

fuse_publishers/fuse_plugins.xml
```diff
@@ -2,7 +2,7 @@
   <class type="fuse_publishers::Path2DPublisher" base_class_type="fuse_core::Publisher">
```
sloretz, line 2: Can't quite comment on it because it's just above the changed lines, but change `lib/libfuse_publishers` to `fuse_publishers`

fuse_publishers/src/pose_2d_publisher.cpp
```diff
@@ -105,29 +109,58 @@ namespace fuse_publishers
...
+  const std::string & name,
+  size_t thread_count)
+{
+  interfaces_ = interfaces;
+  fuse_core::AsyncPublisher::initialize(interfaces_, name, thread_count);
+}
+
 void Pose2DPublisher::onInit()
 {
+  logger_ = std::make_shared<rclcpp::Logger>(interfaces_.get_node_logging_interface()->get_logger());
```
sloretz: I'm not sure we gain anything from `std::shared_ptr` here. The [API is returning a copy of a logger](https://github.com/ros2/rclcpp/blob/3fb012e2e979475f5044ab0e0f9b91d336db5f46/rclcpp/include/rclcpp/node_interfaces/node_logging_interface.hpp#L45-L46), so we may as well `std::move` that copy into a member variable instead of creating a shared pointer to another copy.

fuse_publishers/test/test_pose_2d_publisher.cpp
```diff
@@ -232,48 +251,53 @@ TEST_F(Pose2DPublisherTestFixture, PublishPose)
 TEST_F(Pose2DPublisherTestFixture, PublishPoseWithCovariance)
 {
   // Test that the expected PoseWithCovarianceStamped message is published
+  auto node = rclcpp::Node::make_shared("test_pose_2d_publisher_node");
+  executor_->add_node(node);
+
+  fuse_core::getParam(node, "map_frame", std::string("test_map"));
```
sloretz: Recommend a comment that these calls are declaring parameters on the node, which Pose2DPublisher will then use.

Alternatively, could pass a `NodeOptions` instance with [arguments](https://github.com/ros2/rclcpp/blob/3fb012e2e979475f5044ab0e0f9b91d336db5f46/rclcpp/include/rclcpp/node_options.hpp#L117) set to set these values. I think that would be closer to actual usage.

https://docs.ros.org/en/rolling/How-To-Guides/Node-arguments.html

```C++
rclcpp::NodeOptions options;
options.arguments({
 "--ros-args",
 "-p", "map_frame:=test_map",
 "-p", "odom_frame:=test_odom",
 "-p", "base_frame:=test_base",
 "-p", "publish_to_tf:=false});
```

fuse_publishers/test/test_pose_2d_publisher.cpp
```diff
@@ -232,48 +251,53 @@ TEST_F(Pose2DPublisherTestFixture, PublishPose)
...
-  rclcpp::Time timeout = node_->now() + rclcpp::Duration::from_seconds(10.0);
-  while ((!received_pose_with_covariance_msg_) && (node_->now() < timeout))
+  rclcpp::Time timeout = node->now() + rclcpp::Duration::from_seconds(20.0);
+  while ((!received_pose_with_covariance_msg_) && (node->now() < timeout))
   {
-    rclcpp::sleep_for(rclcpp::Duration::from_seconds(0.10));
+    rclcpp::sleep_for(rclcpp::Duration::from_seconds(0.10).to_chrono<std::chrono::nanoseconds>());
   }
 
+  rclcpp::sleep_for(rclcpp::Duration::from_seconds(5).to_chrono<std::chrono::nanoseconds>());
```
sloretz: Why the extra 5 second sleep here?

fuse_publishers/test/test_pose_2d_publisher.cpp
```diff
@@ -284,33 +308,41 @@ TEST_F(Pose2DPublisherTestFixture, PublishPoseWithCovariance)
 TEST_F(Pose2DPublisherTestFixture, PublishTfWithoutOdom)
 {
   // Test that the expected TFMessage is published
+  auto node = rclcpp::Node::make_shared("test_pose_2d_publisher_node");
+  executor_->add_node(node);
+
+  fuse_core::getParam(node, "map_frame", std::string("test_map"));
```
sloretz: Same comment about `NodeOptions`.

fuse_publishers/test/test_pose_2d_publisher.cpp
```diff
@@ -284,33 +308,41 @@ TEST_F(Pose2DPublisherTestFixture, PublishPoseWithCovariance)
...
-  rclcpp::Time timeout = node_->now() + rclcpp::Duration::from_seconds(10.0);
-  while ((!received_tf_msg_) && (node_->now() < timeout))
+  rclcpp::Time timeout = node->now() + rclcpp::Duration::from_seconds(20.0);
+  while ((!received_tf_msg_) && (node->now() < timeout))
   {
-    rclcpp::sleep_for(rclcpp::Duration::from_seconds(0.10));
+    rclcpp::sleep_for(rclcpp::Duration::from_seconds(0.10).to_chrono<std::chrono::nanoseconds>());
   }
 
+  rclcpp::sleep_for(rclcpp::Duration::from_seconds(5).to_chrono<std::chrono::nanoseconds>());
```
sloretz: Same question about extra 5 second sleep.

fuse_publishers/test/test_pose_2d_publisher.cpp
```diff
@@ -324,31 +356,37 @@ TEST_F(Pose2DPublisherTestFixture, PublishTfWithoutOdom)
 TEST_F(Pose2DPublisherTestFixture, PublishTfWithOdom)
 {
   // Test that the expected TFMessage is published
+  auto node = rclcpp::Node::make_shared("test_pose_2d_publisher_node");
+  executor_->add_node(node);
+
+  fuse_core::getParam(node, "map_frame", std::string("test_map"));
```
sloretz: Same comment about `NodeOptions`.

fuse_publishers/test/test_path_2d_publisher.cpp
```diff
@@ -141,93 +160,111 @@ class Path2DPublisherTestFixture : public ::testing::Test
...
+  geometry_msgs::msg::PoseArray pose_array_msg_;
 };
 
 TEST_F(Path2DPublisherTestFixture, PublishPath)
 {
   // Test that the expected PoseStamped message is published
+  auto node = rclcpp::Node::make_shared("test_path_2d_publisher_node");
+  executor_->add_node(node);
+
+  fuse_core::getParam(node, "frame_id", std::string("test_map"));
```
sloretz: Same comment about `NodeOptions`.

fuse_core/include/fuse_core/publisher.hpp
methylDragon: I was having issues with this one; The problem comes from the fact that the publishers are supposed to be plugins that are loaded in by pluginlib. So all derived classes **must** have a constructor with no arguments, so there'd be no way to configure it using constructor args on the derived class (which means the thread count will have to be defined in the class declaration statically (or otherwise as a template argument), or initialized later on (which is the option I went with.)

Let me try removing the arg on `publisher.hpp` and deferring it to AsyncPublisher instead, maybe that'll work? :thinking:

fuse_publishers/src/pose_2d_publisher.cpp
methylDragon: The logger default constructor is private :(
I'll just spin up a logger with the name "uninitialized" until its replaced

fuse_publishers/test/test_path_2d_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/compare/f9d587a8c481bab1398f71dd2a4a5eb28fa82547..6ee89e8d6b24b282fb99c14482fe2d6412dcc3e9

fuse_publishers/test/test_pose_2d_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/compare/f9d587a8c481bab1398f71dd2a4a5eb28fa82547..6ee89e8d6b24b282fb99c14482fe2d6412dcc3e9

fuse_publishers/test/test_pose_2d_publisher.cpp
methylDragon: No idea :o
https://github.com/locusrobotics/fuse/compare/f9d587a8c481bab1398f71dd2a4a5eb28fa82547..6ee89e8d6b24b282fb99c14482fe2d6412dcc3e9

fuse_publishers/test/test_pose_2d_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/compare/f9d587a8c481bab1398f71dd2a4a5eb28fa82547..6ee89e8d6b24b282fb99c14482fe2d6412dcc3e9

fuse_publishers/test/test_pose_2d_publisher.cpp
methylDragon: No idea :o

https://github.com/locusrobotics/fuse/compare/f9d587a8c481bab1398f71dd2a4a5eb28fa82547..6ee89e8d6b24b282fb99c14482fe2d6412dcc3e9

fuse_publishers/test/test_pose_2d_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/compare/f9d587a8c481bab1398f71dd2a4a5eb28fa82547..6ee89e8d6b24b282fb99c14482fe2d6412dcc3e9

fuse_publishers/src/pose_2d_publisher.cpp
methylDragon: https://github.com/locusrobotics/fuse/compare/f9d587a8c481bab1398f71dd2a4a5eb28fa82547..6ee89e8d6b24b282fb99c14482fe2d6412dcc3e9

fuse_publishers/fuse_plugins.xml
methylDragon: https://github.com/locusrobotics/fuse/compare/f9d587a8c481bab1398f71dd2a4a5eb28fa82547..6ee89e8d6b24b282fb99c14482fe2d6412dcc3e9

fuse_core/include/fuse_core/publisher.hpp
methylDragon: https://github.com/locusrobotics/fuse/compare/f9d587a8c481bab1398f71dd2a4a5eb28fa82547..6ee89e8d6b24b282fb99c14482fe2d6412dcc3e9

methylDragon:
[Build Status](https://build.ros2.org/job/Rpr__fuse__ubuntu_jammy_amd64/37/)

Likewise, only linters

fuse_core/include/fuse_core/async_publisher.hpp
```diff
@@ -81,10 +82,25 @@ class AsyncPublisher : public Publisher
    * AsyncPublisher::onInit() method will be called from here, once the internal node is properly
    * configured.
    *
+   * @param[in] interfaces The node interfaces to be used with the publisher
    * @param[in] name A unique name to give this plugin instance
+   * @param[in] name The number of threads to use with this publisher
    * @throws runtime_error if already initialized
    */
-  void initialize(const std::string & name) override;
```
sloretz, line 87: I think the non-virtual `initialize()` is a problem. The optimizer [is going to call initialize(...)](https://github.com/locusrobotics/fuse/blob/6ee89e8d6b24b282fb99c14482fe2d6412dcc3e9/fuse_optimizers/src/optimizer.cpp#L343) without knowing the type of the class, so all classes need to accept the same arguments.

Maybe we could make `initialize()` accept all known interfaces, and then plugins would use only the ones they needed? I'm playing with this locally now.

fuse_core/include/fuse_core/publisher.hpp
sloretz: >  so there'd be no way to configure it using constructor args on the derived class (which means the thread count will have to be defined in the class declaration statically (or otherwise as a template argument), or initialized later on (which is the option I went with.)

True. It looks like the ROS 1 code we're porting went with defining the value in the subclass statically. I opened a PR on your fork that restores that for `AsyncPublisher` https://github.com/methylDragon/fuse/pull/3

Initializing it later on is a little problematic because `pluginlib` requires all derived classes to have the same signature for `initialize()` https://github.com/locusrobotics/fuse/pull/299#discussion_r1058533703 - that means we'd have to make all publishers accept a `thread_count` argument even if they didn't derive from `AsyncPublisher`.

fuse_core/include/fuse_core/async_publisher.hpp
methylDragon: It seems that making initialize accept all known interfaces is the solution.. I was trying to avoid that so as to have the base class have as few concerns as possible, but I can't think of another way around it if the signatures have to stay the same...

sloretz: I updated https://github.com/methylDragon/fuse/pull/3 to make `Publisher::interfaces()` accept all node interfaces

methylDragon: :bow: Let me take an iterative pass

fuse_core/include/fuse_core/async_publisher.hpp
```diff
@@ -81,10 +82,14 @@ class AsyncPublisher : public Publisher
    * AsyncPublisher::onInit() method will be called from here, once the internal node is properly
    * configured.
    *
+   * @param[in] interfaces The node interfaces to be used with the publisher
    * @param[in] name A unique name to give this plugin instance
+   * @param[in] name The number of threads to use with this publisher
```
sloretz: Nit, remove this line

fuse_core/src/async_publisher.cpp
```diff
@@ -48,43 +48,56 @@ AsyncPublisher::~AsyncPublisher()
   internal_stop();
 }
 
-void AsyncPublisher::initialize(const std::string & name)
+void AsyncPublisher::initialize(
+  node_interfaces::NodeInterfaces<
+    node_interfaces::Base,
```
sloretz: Nit, I think this could use `ALL_RCLCPP_NODE_INTERFACES` macro. The macro says `fuse_core::...`, but I think that will still work inside the `fuse_core` namespace.

fuse_core/test/test_async_publisher.cpp
```diff
@@ -45,7 +45,7 @@ class MyPublisher : public fuse_core::AsyncPublisher
 {
 public:
   MyPublisher()
-  : fuse_core::AsyncPublisher(0),
+  : fuse_core::AsyncPublisher(),
```
sloretz: nit, `1` to match the ROS 1 test: https://github.com/locusrobotics/fuse/blob/0461f779f9ac4defca6ecfd102a675add92bb3a9/fuse_core/test/test_async_publisher.cpp#L49

fuse_publishers/include/fuse_publishers/pose_2d_publisher.h
```diff
@@ -149,24 +168,37 @@ class Pose2DPublisher : public fuse_core::AsyncPublisher
...
+    fuse_core::node_interfaces::Waitables
+  > interfaces_;  //!< Shadows AsyncPublisher interfaces_
+
   using Synchronizer = StampedVariableSynchronizer<fuse_variables::Orientation2DStamped,
                                                    fuse_variables::Position2DStamped>;
-
   std::string base_frame_;  //!< The name of the robot's base_link frame
   fuse_core::UUID device_id_;  //!< The UUID of the device to be published
+  rclcpp::Clock::SharedPtr clock_;  //!< The publisher's clock, for timestamping and logging
+  rclcpp::Logger logger_;  //!< The publisher's logger, shared_ptr for deferred init
```
sloretz: nit, update comment to remove `shared_ptr`

fuse_publishers/test/test_pose_2d_publisher.cpp
```diff
@@ -143,85 +144,105 @@ class Pose2DPublisherTestFixture : public ::testing::Test
...
+    "/test_publisher/pose", 1,
+    std::bind(&Pose2DPublisherTestFixture::poseCallback, this, std::placeholders::_1));
 
   // Send the graph to the Publisher to trigger message publishing
   publisher.notify(transaction_, graph_);
 
   // Verify the subscriber received the expected pose
-  rclcpp::Time timeout = node_->now() + rclcpp::Duration::from_seconds(10.0);
-  while ((!received_pose_msg_) && (node_->now() < timeout))
+  rclcpp::Time timeout = node->now() + rclcpp::Duration::from_seconds(20.0);
```
sloretz: why was the timeout increased from 10s to 20s?

fuse_publishers/test/test_pose_2d_publisher.cpp
```diff
@@ -232,48 +253,53 @@ TEST_F(Pose2DPublisherTestFixture, PublishPose)
...
+    "/test_publisher/pose_with_covariance", 1,
+    std::bind(&Pose2DPublisherTestFixture::poseWithCovarianceCallback, this, std::placeholders::_1));
 
   // Send the graph to the Publisher to trigger message publishing
   publisher.notify(transaction_, graph_);
 
   // Verify the subscriber received the expected pose
-  rclcpp::Time timeout = node_->now() + rclcpp::Duration::from_seconds(10.0);
-  while ((!received_pose_with_covariance_msg_) && (node_->now() < timeout))
+  rclcpp::Time timeout = node->now() + rclcpp::Duration::from_seconds(20.0);
```
sloretz: same question about the timeout increase

fuse_publishers/test/test_pose_2d_publisher.cpp
```diff
@@ -284,31 +310,39 @@ TEST_F(Pose2DPublisherTestFixture, PublishPoseWithCovariance)
...
-    &Pose2DPublisherTestFixture::tfCallback,
-    reinterpret_cast<Pose2DPublisherTestFixture*>(this));
-
   // Send the graph to the Publisher to trigger message publishing
   publisher.notify(transaction_, graph_);
 
   // Verify the subscriber received the expected pose
-  rclcpp::Time timeout = node_->now() + rclcpp::Duration::from_seconds(10.0);
-  while ((!received_tf_msg_) && (node_->now() < timeout))
+  rclcpp::Time timeout = node->now() + rclcpp::Duration::from_seconds(20.0);
```
sloretz: same question about the timeout increase

fuse_publishers/test/test_pose_2d_publisher.cpp
```diff
@@ -324,31 +358,39 @@ TEST_F(Pose2DPublisherTestFixture, PublishTfWithoutOdom)
...
```
sloretz: same question about the timeout increase

fuse_publishers/test/test_path_2d_publisher.cpp
```diff
@@ -141,93 +160,113 @@ class Path2DPublisherTestFixture : public ::testing::Test
...
+    "/test_publisher/pose_array", 1,
+    std::bind(&Path2DPublisherTestFixture::poseArrayCallback, this, std::placeholders::_1));
 
   // Send the graph to the Publisher to trigger message publishing
   publisher.notify(transaction_, graph_);
 
   // Verify the subscriber received the expected pose
-  rclcpp::Time timeout = node_->now() + rclcpp::Duration::from_seconds(10.0);
-  while ((!received_path_msg_) && (node_->now() < timeout))
+  rclcpp::Time timeout = node->now() + rclcpp::Duration::from_seconds(20.0);
```
sloretz: same question about the timeout increase

fuse_core/src/async_publisher.cpp
methylDragon: Ah I missed this one

fuse_publishers/test/test_pose_2d_publisher.cpp
methylDragon: I found that it sometimes times out if left at 10 seconds, same with the ones below

sloretz: Interesting! I wonder if the timeout was happening because of the extra 5 second sleep that was removed https://github.com/locusrobotics/fuse/pull/299#discussion_r1054773194

I changed the timeouts back to 10 seconds and didn't see any failures in 100 runs using:

```
colcon test --event-handlers console_direct+ --packages-select fuse_publishers --ctest-args -R _2d_publisher --retest-until-fail 100
```

https://github.com/methylDragon/fuse/pull/4

sloretz (approved):

PR #307, opened 2023-01-10
fuse -> ROS 2 fuse_optimizers: Port fuse_optimizers

methylDragon:

# Description
This PR ports the entirety of fuse_optimizers, including:

- CMakeLists and packages
- Source code
- Tests
- Other stuff like docs are in other PRs.
- This introduces changes to the usage of rclcpp::Time (defaulting to use of RCL_ROS_TIME where time is default initialized.)

---


fuse_core/src/timestamp_manager.cpp
```diff
@@ -100,9 +100,26 @@ void TimestampManager::query(
       const rclcpp::Time & current_stamp = *current_iter;
       // Check if the timestamp pair is exactly an existing pair. If so, don't add it.
       auto history_iter = motion_model_history_.lower_bound(previous_stamp);
+      rclcpp::Time history_beginning_stamp = history_iter->second.beginning_stamp;
+      rclcpp::Time history_ending_stamp = history_iter->second.ending_stamp;
+
+      // lower_bound() can default construct Time, which causes a mismatched clock error
```
sloretz: Interesting. Since this check is happening after `lower_bound()`, it suggests the problem is `lower_bound()` can return a default constructed time stamp. It looks like `lower_bound()` either returns an iterator to an element that was inserted into the map, or a past the end iterator. We're already checking that `history_iter` is not `end()` below before using it, so the default constructed time must have been inserted into the map. Where did that come from?

Can this logic be removed now that you've added `RCL_ROS_TIME` to the existing default constructed time objects?

fuse_optimizers/CMakeLists.txt
```diff
@@ -37,179 +39,95 @@ add_library(${PROJECT_NAME}
...
+  fuse_variables::fuse_variables
+  pluginlib::pluginlib
+  ${std_srvs_TARGETS}
 )
-set_target_properties(${PROJECT_NAME}
-  PROPERTIES
-    CXX_STANDARD 14
-    CXX_STANDARD_REQUIRED YES
+ament_target_dependencies(
+  ${PROJECT_NAME} diagnostic_updater
```
sloretz: It looks like we can depend on the target `diagnostic_updater::diagnostic_updater`. In general I would always recommend modern CMake targets over `ament_target_dependencies()`

https://github.com/ros/diagnostics/blob/12ab0065178a2ebd6b205232af2dea8dff6b8cde/diagnostic_updater/CMakeLists.txt#L35

fuse_optimizers/CMakeLists.txt
```diff
@@ -37,179 +39,95 @@ add_library(${PROJECT_NAME}
...
 )
 
 ## batch_optimizer node
 add_executable(batch_optimizer_node
   src/batch_optimizer_node.cpp
+  src/batch_optimizer.cpp
```
sloretz: These files are already compiled into the `${PROJECT_NAME}` target. I think this should use `target_link_libraries` to avoid recompiling them. It looks like it's already doing so, so these three extra files could be removed from this executable.

fuse_optimizers/CMakeLists.txt
```diff
@@ -37,179 +39,95 @@ add_library(${PROJECT_NAME}
...
-)
-target_link_libraries(batch_optimizer_node
-  ${PROJECT_NAME}
-  ${catkin_LIBRARIES}
-)
-set_target_properties(batch_optimizer_node
-  PROPERTIES
-    CXX_STANDARD 14
-    CXX_STANDARD_REQUIRED YES
+target_include_directories(${PROJECT_NAME} PUBLIC
```
sloretz: This `target_include_directories()` call is a duplicate of the one above, and should be removed. `batch_optimizer_node` will get the include directories it needs automatically by `target_link_libraries` with the `${PROJECT_NAME}` target

fuse_optimizers/CMakeLists.txt
```diff
@@ -37,179 +39,95 @@ add_library(${PROJECT_NAME}
...
+  "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>"
 )
+target_link_libraries(batch_optimizer_node ${PROJECT_NAME})
 
 ## fixed_lag_smoother node
 add_executable(fixed_lag_smoother_node
   src/fixed_lag_smoother_node.cpp
+  src/fixed_lag_smoother.cpp
+  src/optimizer.cpp
+  src/variable_stamp_index.cpp
```
sloretz: Same comment about depending on the `${PROJECT_NAME}` target

fuse_optimizers/CMakeLists.txt
```diff
@@ -37,179 +39,95 @@ add_library(${PROJECT_NAME}
...
-)
-target_link_libraries(fixed_lag_smoother_node
-  ${PROJECT_NAME}
-  ${catkin_LIBRARIES}
-)
-set_target_properties(fixed_lag_smoother_node
-  PROPERTIES
-    CXX_STANDARD 14
-    CXX_STANDARD_REQUIRED YES
+target_include_directories(${PROJECT_NAME} PUBLIC
```
sloretz: Same comment about this being a duplicate that can be removed.

fuse_optimizers/include/fuse_optimizers/batch_optimizer.h
```diff
@@ -108,9 +109,10 @@ class BatchOptimizer : public Optimizer
    * @param[in] private_node_handle A node handle in the node's private namespace
    */
   BatchOptimizer(
-    fuse_core::Graph::UniquePtr graph,
-    const ros::NodeHandle& node_handle = ros::NodeHandle(),
-    const ros::NodeHandle& private_node_handle = ros::NodeHandle("~"));
+    rclcpp::NodeOptions options,
+    std::string node_name = "batch_optimizer_node",
+    fuse_core::Graph::UniquePtr graph = nullptr
```
sloretz, line 111: Interesting. Before `graph` was a required parameter, but now it's optional. I'd recommend instead:

```C++
BatchOptimizer(
    fuse_core::Graph::UniquePtr graph,
    std::string node_name = "batch_optimizer_node",
    rclcpp::NodeOptions options = rclcpp::NodeOptions());
```

fuse_optimizers/include/fuse_optimizers/batch_optimizer_params.h
```diff
@@ -80,25 +80,34 @@ struct BatchOptimizerParams
   /**
    * @brief Method for loading parameter values from ROS.
    *
-   * @param[in] nh - The ROS node handle with which to load parameters
+   * @param[in] node_handle - The node used to load the parameter
```
sloretz: nit, `node_handle` -> `interfaces`

fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother.h
```diff
@@ -118,9 +121,10 @@ class FixedLagSmoother : public Optimizer
    * @param[in] private_node_handle A node handle in the node's private namespace
    */
   FixedLagSmoother(
-    fuse_core::Graph::UniquePtr graph,
-    const ros::NodeHandle& node_handle = ros::NodeHandle(),
-    const ros::NodeHandle& private_node_handle = ros::NodeHandle("~"));
+    rclcpp::NodeOptions options,
+    std::string node_name = "fixed_lag_smoother_node",
+    fuse_core::Graph::UniquePtr graph = nullptr
```
sloretz, line 123: Same comment about keeping the order where `graph` is required and `node_name` and `options` are optional.

fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother_params.h
```diff
@@ -65,22 +65,22 @@ struct FixedLagSmootherParams
...
+   * parameter in Hz. "optimization_frequency" will be prioritized.
    */
-  rclcpp::Duration optimization_period { RCUTILS_S_TO_NS(0.1) };
+  rclcpp::Duration optimization_period { 0, static_cast<uint32_t>(RCUTILS_S_TO_NS(0.1)) };
 
   /**
    * @brief The topic name of the advertised reset service
    */
-  std::string reset_service { "~reset" };
+  std::string reset_service { "reset" };
```
sloretz: I think the equivalent of `~reset` in ROS 2 is `~/reset`.

> must separate a tilde (~) from the rest of the name with a forward slash (/), i.e. ~/foo not ~foo

http://design.ros2.org/articles/topic_and_service_names.html


I honestly didn't know `~abc` worked in ROS 1 :no_mouth:

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -98,15 +101,21 @@ FixedLagSmoother::FixedLagSmoother(
...
     std::bind(&FixedLagSmoother::optimizerTimerCallback, this)
   );
 
   // Advertise a service that resets the optimizer to its initial state
-  reset_service_server_ = node_handle_.advertiseService(
-    ros::names::resolve(params_.reset_service),
-    &FixedLagSmoother::resetServiceCallback,
-    this);
+  reset_service_server_ = create_service<std_srvs::srv::Empty>(
+    fuse_core::joinTopicName(get_name(), params_.reset_service),
```
sloretz: Ah, I see why `reset` instead of `~/reset`. The topic name is being joined here. Using `~/reset` means both the node's namespace and node name will be added automatically, so maybe `joinTopicName()` would no longer be required here?

fuse_optimizers/test/launch_tests/test_fixed_lag_ignition.cpp
```diff
@@ -31,58 +31,95 @@
...
+};
+
+TEST_F(FixedLagIgnitionFixture, SetInitialState)
+{
+  auto node = rclcpp::Node::make_shared("fixed_lag_ignition_test");
+  executor_->add_node(node);
+
+  auto relative_pose_publisher =
+    node->create_publisher<geometry_msgs::msg::PoseWithCovarianceStamped>(
+      "pose_sensor/relative_pose", 1);
```
sloretz: The topic name being `pose_sensor/relative_pose` instead of `/relative_pose` seems like an API change we could maybe avoid. Is this due to `relative_pose` being joined with `get_name()` somewhere?

fuse_optimizers/test/launch_tests/test_fixed_lag_ignition.cpp
```diff
@@ -109,36 +146,33 @@ TEST(FixedLagIgnition, SetInitialState)
...
-  rclcpp::Time result_timeout = node->now() + rclcpp::Duration::from_seconds(3.0);
-  auto odom_msg = nav_msgs::Odometry::ConstPtr();
-  while ((!odom_msg || odom_msg->header.stamp != rclcpp::Time(3, 0)) &&
+  rclcpp::Time result_timeout = node->now() + rclcpp::Duration::from_seconds(1.0);
+  auto odom_msg = nav_msgs::msg::Odometry();
+  while ((odom_msg.header.stamp != rclcpp::Time(3, 0, RCL_ROS_TIME)) &&
          (node->now() < result_timeout))
   {
-    odom_msg = ros::topic::waitForMessage<nav_msgs::Odometry>("/odom", rclcpp::Duration::from_seconds(1.0));
+    rclcpp::wait_for_message(odom_msg, node, "/odometry_publisher/odom", std::chrono::seconds(1));
```
sloretz: Same comment about `/odometry_publisher/odom` being an API change somewhere, especially with bare "odom" being a fairly common topic name: https://www.ros.org/reps/rep-0105.html#odom

sloretz (commented):
In progress review - will continue tomorrow

fuse_optimizers/test/launch_tests/test_fixed_lag_ignition.cpp
methylDragon: Isn't that REP about TF frames and not topics?

Though, you're right. I misrecalled that the publisher was using a private_node_handle and ended up pre-pending the odom name in the publisher code. (It's using a node_handle in the [ROS 1 code](https://github.com/locusrobotics/fuse/blob/8b62c8088211a94b347e223b585aaf2c302bf287/fuse_models/src/odometry_2d_publisher.cpp#L85), so it is indeed bare)

(To clarify: Only the odometry publisher is using a normal node_handle as opposed to a private one. The others still used private node handles and so justified the use of joinTopicName (e.g. the pose publisher))

Note to self to fix it soon.

fuse_optimizers/test/launch_tests/test_fixed_lag_ignition.cpp
methylDragon: In the ROS 1 code, the pose topic for the pose publisher uses a [private_node_handle](https://github.com/locusrobotics/fuse/blob/devel/fuse_publishers/src/pose_2d_publisher.cpp#L173), so the topic would be prepended with the namespace of the node.

But in the port, we're using the node of the optimizer for any fuse publishers (and we're also not using subnodes), so this prepending functionality has been replaced with [joinTopicName](https://github.com/locusrobotics/fuse/blob/rolling/fuse_publishers/src/pose_2d_publisher.cpp#L198))

I just mimicked the same sort of behavior with the test. (In this case the test is just bringing up a bare publisher instead of a fuse publisher.)

fuse_core/src/timestamp_manager.cpp
methylDragon: This explicit check was the last thing I added (after adding RCL_ROS_TIME to all default constructed time objects that I could find.) Not having this check results in a mismatched clock type comparison error.

I had assumed that lower_bound() would end up returning default constructed instances instead of end()... But since it doesn't it means that there looks to be an initial default constructed(??) stamp getting added somewhere (but I don't see a place where it could get inserted :scream: )

I think it'll take me awhile to trace, so I might TODO this since the functionality works as of now.
I'll update the comment accordingly.

---

Also I think this is unrelated but I am aware of a single comparison somewhere between two instances of rclcpp::Time(0, 0, **RCL_CLOCK_UNINITIALIZED**) that happens on init, but I couldn't trace where they were coming from.

fuse_optimizers/CMakeLists.txt
methylDragon: For some reason I couldn't use that target, which is why I ended up falling back to ament_target_dependencies

```
CMake Error at CMakeLists.txt:36 (add_library):
  Target "fuse_optimizers" links to target
  "diagnostic_updater::diagnostic_updater" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?
```

fuse_optimizers/test/launch_tests/test_fixed_lag_ignition.cpp
methylDragon: Fixed! https://github.com/locusrobotics/fuse/pull/307/commits/e2b585edd347f37248d9f5ad51c23fc4c9826a09

fuse_optimizers/src/fixed_lag_smoother.cpp
methylDragon: https://github.com/locusrobotics/fuse/pull/307/commits/e2b585edd347f37248d9f5ad51c23fc4c9826a09 !

fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother_params.h
methylDragon: https://github.com/locusrobotics/fuse/pull/307/commits/e2b585edd347f37248d9f5ad51c23fc4c9826a09

fuse_core/src/timestamp_manager.cpp
methylDragon: Comment updated

fuse_optimizers/include/fuse_optimizers/batch_optimizer.h
methylDragon: I needed to be able to defer the initialization of the graph because a Node is required to pull the graph parameters to start the graph up. (Before, in ROS 1, the global parameter server could be queried without a node, so a graph could be easily created before the optimizer)

The way the classes have been architected, one easy place to get access to a node is the optimizer object itself (but the graph can't be created while said Node is getting constructed.)

So later on, in the constructor body, the node is used if the graph isn't populated to create the graph.

```cpp
  if (!graph) {
    fuse_graphs::HashGraphParams hash_graph_params;
    hash_graph_params.loadFromROS(*this);
    graph_ = std::move(fuse_graphs::HashGraph::make_unique(hash_graph_params));
  }
```

fuse_optimizers/CMakeLists.txt
methylDragon: Fixed!

fuse_optimizers/CMakeLists.txt
methylDragon: Fixed!

fuse_optimizers/CMakeLists.txt
methylDragon: Fixed!

fuse_optimizers/CMakeLists.txt
methylDragon: Fixed!

fuse_optimizers/include/fuse_optimizers/batch_optimizer_params.h
methylDragon: Fixed!

fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother.h
methylDragon: Same note about graph initialization deferral

fuse_core/src/timestamp_manager.cpp
sloretz: I did some debugging with where the mismatched timestamps were coming from. It looks like it was the construction of the `TimestampManager::MotionModelSegment()` here: https://github.com/locusrobotics/fuse/pull/307/files#diff-30ff4b063bdf335221719ed9125d42c7473a269f56d2a638392bf8f482e7a723R154

That class had two time stamps `begginning_stamp` and `ending_stamp` that were being default constructed. I fixed it by removing the default constructor and replacing it with one that takes a clock type and uses it to initialize those stamps. The advantage is it allows `TimestampManager` to use any clock type, as long as the same clock is consistently used. I made sure the `test_timestamp_manager.cpp` test worked with the `RCL_ROS_TIME` insertions, then reverted them and found it still works :tada:

fuse_optimizers/include/fuse_optimizers/batch_optimizer.h
sloretz: > I needed to be able to defer the initialization of the graph because a Node is required to pull the graph parameters to start the graph up.

Ah that's very interesting. The `Node` needs to exist before a `Graph` (or a `HashGraph` instance) can exist, but the node is being created inside the Optimizer, and the Optimizer has an argument that expects the `Graph` to already exist. Seems like a chicken and egg problem.

There's another issue: Optimizer doesn't allow lifecycle nodes to be used. A user would have to create a separate `LifecycleOptimizer` to create a lifecycle node inside.

Looking at the ROS 1 code, I notice the Optimizers are passed in node handles, which seems sort of like passing in nodes. What if we make the `Optimizer` accept `rclcpp::NodeInterfaces` instead? Then users can both create lifecycle nodes, and could initialize a Graph ahead of time if they wanted to. That would mean changing the constructor args here to `BatchOptimizer(fuse_core::Graph::UniquePtr graph, rclcpp::NodeInterfaces interfaces)`. The downside is now the user is forced to provide the interfaces argument instead of having a node created for them.

methylDragon: I'll do: `BatchOptimizer(rclcpp::NodeInterfaces interfaces, fuse_core::Graph::UniquePtr graph)`

So that the graph is an optional param for potential deferred creation

methylDragon: Fixed! https://github.com/locusrobotics/fuse/pull/307/commits/864acc55cadf39b4b23eea1b087b2169c6c059e9

fuse_optimizers/src/batch_optimizer.cpp
```diff
@@ -47,21 +48,24 @@ namespace fuse_optimizers
...
+  params_.loadFromROS(interfaces_);
 
   // Configure a timer to trigger optimizations
-  optimize_timer_ = this->create_timer(
-    params_.optimization_period,
-    std::bind(&BatchOptimizer::optimizerTimerCallback, this)
+  optimize_timer_ = rclcpp::create_timer(
+    interfaces_,
+    clock_,
+    params_.optimization_period.to_chrono<std::chrono::nanoseconds>(),
```
sloretz: `rclcpp::create_timer()` takes an `rclcpp::Duration`, so no need to convert this. ba487fa

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -97,16 +99,28 @@ FixedLagSmoother::FixedLagSmoother(
   optimization_thread_ = std::thread(&FixedLagSmoother::optimizationLoop, this);
 
   // Configure a timer to trigger optimizations
-  optimize_timer_ = this->create_timer(
-    params_.optimization_period,
-    std::bind(&FixedLagSmoother::optimizerTimerCallback, this)
+  optimize_timer_ = rclcpp::create_timer(
+    interfaces_,
+    clock_,
+    params_.optimization_period.to_chrono<std::chrono::nanoseconds>(),
```
sloretz: Same comment about not needed to convert to chrono 0ce06cd

fuse_optimizers/src/optimizer.cpp
```diff
@@ -123,64 +109,83 @@ Optimizer::~Optimizer()
 
 void Optimizer::loadMotionModels()
 {
-  // Read and configure all of sensor model plugins
-  if (!private_node_handle_.hasParam("motion_models"))
-  {
-    return;
+  const std::string param_prefix = "motion_models.";
+  const std::string list_param_full_name = param_prefix + "motion_model_list";
```
sloretz: I think we can avoid adding the list parameters, which makes the parameter interface a little closer to the original. I opened a PR on your fork, mind taking a look?

https://github.com/methylDragon/fuse/pull/5

fuse_optimizers/include/fuse_optimizers/batch_optimizer.h
sloretz: ping @svwilliams  This PR conveniently creates a `HashGraph` instance automatically if the given graph argument is `nullptr`. Previously the user was required to construct a graph and pass it in. Thoughts?

fuse_models/src/graph_ignition.cpp
```diff
@@ -180,13 +188,24 @@ void GraphIgnition::process(const fuse_msgs::msg::SerializedGraph& msg)
...
 
     auto srv = std::make_shared<std_srvs::srv::Empty::Request>();
-    // No need to spin since node is optimizer node, which should be spinning
-    auto result_future = reset_client_->async_send_request(srv);
-    result_future.wait();
+    // Don't block the executor.
+    // It needs to be free to handle the response to this service call.
+    // Have a callback do the rest of the work when a response comes.
+    auto result_future = reset_client_->async_send_request(srv,
+      [this, post_process, &graph, msg](rclcpp::Client<std_srvs::srv::Empty>::SharedFuture result){
```
sloretz: It looks like `graph` is a local variable (of type `std::unique_ptr<...>`), so if I understand correctly it will probably already be destructed by the time this lambda is run. I think this needs to be a move capture `, graph = std::move(graph), ` instead

fuse_optimizers/src/optimizer.cpp
```diff
@@ -31,15 +31,18 @@
...
 #include <fuse_core/callback_wrapper.hpp>
 #include <fuse_core/graph.hpp>
 #include <fuse_core/time.hpp>
 #include <fuse_core/transaction.hpp>
 #include <fuse_core/uuid.hpp>
 #include <fuse_optimizers/optimizer.h>
+#include <fuse_graphs/hash_graph.hpp>
 
-#include <XmlRpcValue.h>
+//#include <XmlRpcValue.h>
```
sloretz: nit, delete this line

fuse_optimizers/src/optimizer.cpp
```diff
@@ -51,60 +54,43 @@
...
-
-  std::string name;            //!< Plugin name
-  std::string type;            //!< Plugin type
-  XmlRpc::XmlRpcValue config;  //!< The entire configuration, that might have additional optional parameters
-};
-
-// TODO(CH3): pass a rclcpp::Context, and a CallbackGroup
-Optimizer::Optimizer(rclcpp::NodeOptions options, fuse_core::Graph::UniquePtr graph) :
-    Node("optimizer_node", options),
+// XXX pass a rclcpp::Context, and a CallbackGroup
```
sloretz: Nit, comment no longer required

fuse_optimizers/test/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
+# CORE GTESTS ======================================================================================
+set(TEST_TARGETS
+  # test_fixed_lag_ignition
+  # test_optimizer
+  test_variable_stamp_index
+)
+
+foreach(test_name ${TEST_TARGETS})
```
sloretz: nit, looks like the loop and `TEST_TARGETS` variable are no longer necessary. I'd recommend removing them

fuse_optimizers/test/launch_tests/test_parameters.cpp
```diff
@@ -0,0 +1,244 @@
...
+{
+  // Load parameters enforcing they are positive:
+  const double default_value{1.0};
+
+  auto node = rclcpp::Node::make_shared("test_parameters_node");
+
+  // Load a positive parameter:
+  {
+    double parameter{default_value};
+    fuse_core::getPositiveParam(node, "positive_parameter", parameter);
```
sloretz: This looks like a copy of a test from `fuse_core`. I think this file can be deleted.

sloretz (commented):
These are my last comments on this PR. Once they're addressed and https://github.com/methylDragon/fuse/pull/5 is included then this PR will LGTM :rocket:

fuse_models/src/graph_ignition.cpp
methylDragon: - I had to convert the graph uniqueptr into a shared_ptr so it can be copyable.
	- This is because the async service call signature we're using requires a std::function, which must be copyable (and so requires all args to be copyable)

But aside from that, this bug is fixed! Nice catch!

PR #311, opened 2023-01-28
Fix SerializedPublisher not being able to read it's parameters

sloretz:
Part of #276 

This fixes `SerializedPublisher` not being able to read it's parameters in ROS 2.

@svwilliams FYI

here's an updated fuse_simple_tutorial.yaml that shows how to pass parameters to it

```yaml
state_estimator:
  ros__parameters:
    # Fixed-lag smoother configuration
    optimization_frequency: 20.0
    transaction_timeout: 0.01
    lag_duration: 0.5

    motion_models:
      unicycle_motion_model:
        type: fuse_models::Unicycle2D

    unicycle_motion_model:
      #                         x      y      yaw    vx     vy     vyaw   ax   ay
      process_noise_diagonal: [0.100, 0.100, 0.100, 0.100, 0.100, 0.100, 0.1, 0.1]

    sensor_models:
      initial_localization_sensor:
        type: fuse_models::Unicycle2DIgnition
        motion_models: [unicycle_motion_model]
        ignition: true
      odometry_sensor:
        type: fuse_models::Odometry2D
        motion_models: [unicycle_motion_model]
      imu_sensor:
        type: fuse_models::Imu2D
        motion_models: [unicycle_motion_model]

    initial_localization_sensor:
      publish_on_startup: true
      #                x      y      yaw    vx     vy     vyaw    ax     ay
      initial_state: [0.000, 0.000, 0.000, 0.000, 0.000, 0.000, 0.000, 0.000]
      initial_sigma: [0.100, 0.100, 0.100, 0.100, 0.100, 0.100, 0.100, 0.100]

    odometry_sensor:
      topic: "odom"
      twist_target_frame: "base_footprint"
      linear_velocity_dimensions: ['x', 'y']
      angular_velocity_dimensions: ['yaw']

    imu_sensor:
      topic: "imu"
      angular_velocity_dimensions: ['yaw']
      linear_acceleration_dimensions: ['x', 'y']
      twist_target_frame: "base_footprint"

    publishers:
      filtered_publisher:
        type: fuse_models::Odometry2DPublisher
      serialized_publisher:
        type: fuse_publishers::SerializedPublisher

    filtered_publisher:
      topic: "odom_filtered"
      base_link_frame_id: "base_footprint"
      odom_frame_id: "odom"
      map_frame_id: "map"
      world_frame_id: "odom"
      publish_tf: true
      publish_frequency: 10.0

    serialized_publisher:
      frame_id: "odom"
```

PR #314, opened 2023-01-28
Use getParameterName and namespace parameters for publishers

methylDragon:
Part of #276

With the merge of #311, a new method was introduced to join parameter names together.

Also, it surfaced an inconsistency in how the publishers in fuse_publishers read their parameters. So I brought them in-line with the way SerializedPublisher was doing it following #311

In the ROS 1 code, private node handles were used to load the parameters, so by right they should be namespaced. This PR fixes that issue.


sloretz (approved):
I didn't build this locally, but assuming the only tests that fail are in the un-linted packages then this LGTM!

Issue #300, opened 2022-12-19
Sensor model constraints with initial guess from current graph optimization


fabianhirmann:
Hi,

We are fusing sensor data from odometry, IMU and laser data scan-to-scan and scan-to-map using _fuse_ with the fixed lag smoother optimizer and noticed an issue for our scan-to-scan and scan-to-map sensor models/constraints. Especially our used scan-to-map algorithm needs a good initial guess and therefore we wrote our own ceres cost function to take the guess from the optimization loop. At our own cost function we take at the overridden function `Evaluate` the (first) incoming guess, compute the scan-to-map pose estimate, cache it for subsequent calls, and then calculate the residuals and jacobians based on the pose error and covariance of the pose estimate. 

This approach works most times fine but at some strange situations we noticed that our localization is moving either just shortly for a few samples but sometimes even permanent although the robot is at standstill. We found then out that the initial guess at the `Evaluate` function of the cost function is sometimes uninitialized and exactly zero which in some situations causes the above described behavior. 

So our questions are now:
1. How would you set an initial guess (for a scan-to-map algorithm) when at the callback when getting the sensor data (in this case laser pointcloud) the initial guess is not known? At best we would like to incorporate also the other new sensor data (in this case odometry and IMU) in the initial guess such that the initial guess is as good as possible.
2. If there is no _good_ option for the first question, what would you think of extending the fixed lag smoother optimizer with a variant of stepwise optimization with configurable stages with for example first optimize using only odometry and IMU and then optimize using all sensors? If you think this is a good idea without any bad side effects, we could also try to implement this in a PR ourselves.

If you have any further questions, don't hesitate asking.

Thanks,
Fabian

jakemclaughlin6:
You could take the most recent pose in the graph (T0), then use all the odometry and imu measurements that came after that pose to estimate the relative change from it. You could do that by directly using fuse's hash graph to add whatever odometry and imu constraints between T0 and your pose at the desired timestamp (T1) to do a standalone optimisation in your sensor model. The initial guess for T1 could be obtained from the odometry? Then optimizing with the imu constraints as well gives you the fused initial estimate

svwilliams:
[...]

PR #319, opened 2023-02-28
Add step-wise optimization for the fixed-lag smoother (issue #300)

fabianhirmann:
This PR adds a step-wise optimization for the fixed-lag smoother for the issue with the existing fixed-lag smoother described in #300.

Please provide feedback if this PR is ready to be merged or what is missing. If you have questions, feel free to ask.

### DIFF
--- a/fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother.h
+++ b/fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother.h
@@ -126,6 +126,8 @@ public:
    */
   virtual ~FixedLagSmoother();
 
+  virtual void startOptimization();
+
 protected:
   /**
    * Structure containing the information required to process a transaction after it was received.
--- a/fuse_optimizers/include/fuse_optimizers/step_wise_fixed_lag_smoother_params.h
+++ b/fuse_optimizers/include/fuse_optimizers/step_wise_fixed_lag_smoother_params.h
@@ -57,11 +57,11 @@ namespace fuse_optimizers
 struct StepWiseFixedLagSmootherParams : FixedLagSmootherParams
 {
 public:
-
   /**
    * @brief The ordered list of steps for the optimization with each source inside
    *
-   * The optimization is performed sequentially where each step adds more constraints from different sources to the graph. 
+   * The optimization is performed sequentially where each step adds
+   * more constraints from different sources to the graph.
    */
   std::vector<std::vector<std::string>> optimization_steps;
 
@@ -73,13 +73,13 @@ public:
   void loadFromROS(const ros::NodeHandle& nh)
   {
     FixedLagSmootherParams::loadFromROS(nh);
-    
+
     // load optimization steps
     XmlRpc::XmlRpcValue optimization_steps_raw;
     fuse_core::getParamRequired<XmlRpc::XmlRpcValue>(nh, "optimization_steps", optimization_steps_raw);
 
     ROS_ASSERT(optimization_steps_raw.getType() == XmlRpc::XmlRpcValue::TypeArray);
-    
+
     for (int step_i = 0; step_i < optimization_steps_raw.size(); ++step_i)
     {
       ROS_ASSERT(optimization_steps_raw[step_i].getType() == XmlRpc::XmlRpcValue::TypeStruct);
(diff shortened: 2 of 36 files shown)


fabianhirmann:
After I haven't heard any comments on that for a while, I would like to make a short ping to  @svwilliams, @carlos-m159 and @ayrton04 after mostly you had a look at the last pull requests in this repository.

fuse_optimizers/include/fuse_optimizers/step_wise_fixed_lag_smoother_params.h
```diff
@@ -0,0 +1,106 @@
...
+   * @brief Method for loading parameter values from ROS.
+   *
+   * @param[in] nh - The ROS node handle with which to load parameters
+   */
+  void loadFromROS(const ros::NodeHandle& nh)
+  {
+    FixedLagSmootherParams::loadFromROS(nh);
+
+    // load optimization steps
+    XmlRpc::XmlRpcValue optimization_steps_raw;
```
svwilliams, line 78: Can you add an include for XmlRpcValue?

fuse_optimizers/include/fuse_optimizers/fixed_lag_smoother.h
```diff
@@ -126,6 +126,8 @@ class FixedLagSmoother : public Optimizer
    */
   virtual ~FixedLagSmoother();
 
+  virtual void startOptimization();
```
svwilliams, line 129: Can you add a function description comment for `startOptimization()`?

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -93,6 +93,15 @@ FixedLagSmoother::FixedLagSmoother(
...
   autostart();
 
+  // Advertise a service that resets the optimizer to its initial state
+  reset_service_server_ = node_handle_.advertiseService(
+    ros::names::resolve(params_.reset_service),
+    &FixedLagSmoother::resetServiceCallback,
+    this);
+}
+
+void FixedLagSmoother::startOptimization()
```
svwilliams, line 103: We should probably initialize the `optimization_running_` flag to false in the constructor, and then set it to true in the `startOptimization()` method. Just to adhere to the principle of least surprise.

fuse_optimizers/src/step_wise_fixed_lag_smoother.cpp
```diff
@@ -0,0 +1,236 @@
...
+          {
+            constraints_for_removal.insert(constraint.uuid());
+          }
+        }
+
+        // remove all constraints marked for removal
+        for (auto &&i : constraints_for_removal)
+        {
+          new_transaction_for_loop->removeConstraint(i);
+        }
```
svwilliams, line 168: Here you are making a copy of everything, and then deleting the things you don't want. I have some efficiency questions about this approach.

Also, you are only processing the AddedConstraints set. This means that:
* All the new variables from all of the new constraints are added during the first iteration. This potentially leaves variables unconstrained during the optimization. L-M optimization often let's you get away with this, but it is not ideal.
* If a sensor requested that a previous constraint or variable gets removed, all of the removals are also happening during the first iteration. Again, if a constraint gets removed but its replacement is not added back, you could end up with an under-constrained system during optimization.

The fixed-lag smoother keeps a queue of all of the original Transactions from the sensors. The `processQueue()` method applies the motion models, then merges all of the pending transactions into a single Transaction so the Graph can be updated in a single operation. My personal recommendation would be to replace the `processQueue()` call with your own implementation. You could loop through the queue multiple times, selecting only the Transactions from sensors you want to include during a given iteration. That ensures that variables/constraints are added and deleted at the correct times.

svwilliams (commented):
First, I like to reiterate that the design intent of fuse was to allow people to extend the variables, constraints, publishers, etc. from outside the main project. This allows contributors share it under different terms, tailor it to specific use-cases, and generally retain control of their own code. If you are interested in releasing the StepWiseFixedLagSmoother from your own repo, I will happily adapt the current FixedLagSmoother class to make derived classes easier to implement.

Second, while most of fuse was designed to be extended, the BatchOptimizer and FixedLagSmoother don't quite live up to the rest. The implementation became complex and a bit unwieldy. Having created a derived FixedLagSmoother, I'd be interested in any design changes or recommendations you have to make life easier in the future.

And finally, I have a few concerns about how the constraints are being selected for each optimization phase. The sensors send Transaction objects the the optimizer. Each transaction may add or remove variables, and add or remove constraints. It is assumed that each sensor is smart enough to perform these operations in a correct and consistent order. If a sensor wants to remove a variable, it will also remove all of the constraints involved with that variable at the same time. A Transaction was intended to be an atomic unit of work. But the way the current PR separates work from different sensors only handles parts of each Transaction. And processing partial transactions could lead to unexpected results.

Looking forward to hearing your thoughts about any of this.

fabianhirmann:
Hi @svwilliams,

Thank you for your comments to my pull request. Our intention was indeed to include this step wise fixed lag smoother in the public repository such that it is accessible and available for everyone in the same way as fuse as public repository is helpful for us.

About the other comments, I would first like to resolve the issue described in #300 and then we can decide together in which way this implementation is useful for everyone (including us).

I look forward hearing from you in my issue #300,
Fabian

PR #320, opened 2023-03-03
Fix rviz dependency

ayrton04:
`bloom` was unhappy because it couldn't find the `rviz` dependency. The package is now called `rviz2`:

https://github.com/ros/rosdistro/blob/ead725a32836272f66d80d96efb587fa63374511/rolling/distribution.yaml#L5386


ayrton04:
I accidentally opened this against `devel` and then changed the base, so the ROS build farm is grumpy because it tried to run the PR build against the ROS 1 distros.

ayrton04:
@svwilliams I am going to merge this and let your review tag serve as an FYI.

svwilliams (approved):

PR #323, opened 2023-04-27
[Issue-300] Handle prior on last stamp

svwilliams:
As discussed in Issue #300, there is a bug in the TimestampManager.

When a Transaction is sent to the TimestampManager, it is expected that the variables involved in that Transaction will be updated to the values predicted by the motion model. This was not happening in the case where the Transaction involved a single timestamp. The code to update the output Transaction with the motion model values was included in the loop that processes timestamp pairs. If there is only one timestamp, then there are zero timestamp pairs and that loop never executes.

This PR first creates a few new unit tests to capture this condition: 0ab7189
Then moves the code block out of the pair handling loop: 9729c83
The code block now exists in its own loop, after all of the involved timestamps have been located. This specific location allows us to reuse lookups into the motion model history.


fabianhirmann:
Hi @svwilliams,

I can confirm that these changes fix our issue. I look forward to this PR being merged.

Thank you very much,
Fabian

carlos-m159 (approved):

fabianhirmann:
@svwilliams @carlos-m159 Just a quick question, are there any blocking points preventing this PR to be merged?

svwilliams:
My biggest blocker is getting enough time to port this into the ROS2 implementation at the same time.

fabianhirmann:
@svwilliams Thank you very much for your feedback. Sorry that I was bothering you. I wasn't aware that the ROS2 implementation is desired before merging this PR.

PR #326, opened 2023-05-11
Update devel to build on Ubuntu Jammy (22.04)

svwilliams:
This updates the C++ version to 2017 and suppresses a few warnings from included libraries.
I'm also grabbing the pluginlib .hpp commit from Lucas Walter <wsacul@gmail.com>
This should compile with ROS Noetic on Ubuntu 20.04 and 22.04. However, because Ubuntu 22.04 is not an officially supported OS for ROS Noetic, I have created a noetic-devel branch that does not include this change.


garyservin (approved):
LGTM

carlos-m159 (approved):

PR #334, opened 2023-08-02
[RST-7809] Fix optimization errors when the orientation is initialized at +PI

svwilliams:
If you initialize a robot with an orientation of exactly +PI, the Ceres optimizer will fail. This results in an unusable output, and the entire optimizer node shuts down.

The reason for this is a rather complex series of things. I'll do my best to summarize:
* Orientation variables are not technically linear because of the 2PI roll-over effect. Orientation variables use special local parameterizations when applying updates to force the updates to wrap when needed.
* Fuse uses a valid angle range of [-PI, +PI). So +PI is not technically a valid value. It should be wrapped to -PI. The choice of -PI versus +PI is arbitrary, but one of them is not valid.
* When optimizing a variable with non-minimum phase (-2PI, +PI, +2PI, +3PI, etc.), the cost functions properly treat the difference between +PI and -PI as a zero error. This is correct.
* However, the local parameterization we report a delta to be applied of 2PI. This is also correct....but it is strange to have a correction when the error is zero.
* When optimizing, the Ceres optimizer has a set of criteria it uses to decide when to stop iterating. There is no absolute error condition. Instead we have several ratios (things like `(new_cost - old_cost) < function_tolerance * old_cost`). Because the original error is zero, none of these ratio or proportional stopping criteria are triggered after the first iteration. 0 is not less than 0.
* There is an absolute check on the gradient value which terminates the iterations in normal cases of zero error. However, in the orientation case, the gradient is large because of the 2PI rollover correction. So that stopping criteria does not trigger either.
* The second and all subsequent iterations fail because it is impossible to find a step that reduces the error below zero.

In order to fix this issue without modifying how the Ceres stopping criteria work, the orientation variables must always be minimum phase. However, I have not found any bullet-proof ways of enforcing that requirement.

This PR:
* Marks the original `Orientation2D::yaw()` methods as deprecated. The way these accessors were written prevents me from inserting additional checks or enforcing requirements.
* Adds a new getter() and setter() for the yaw value of the Orientation2D class. These new methods do allow me to wrap the user input to the proper range before setting the variable value.
* Update all uses to the new getter()/setter() methods.

Why is this not an ideal solution?
* Even though they are marked as deprecated, the old methods still exist and can be used. So it is still possible to set the orientation variable to a non-minimum phase.
* The Ceres Solver interface requires read-write access to the variable values. This is a requirement of the NLS solver and cannot be avoided. So there will always be a method to set the variable values to anything the user desires. The API is just less friendly.
* I should really update _all_ of the variables to have getters() and setters() for consistency. I will do that at some point, but I didn't want to clutter up this PR any more than required.

If anyone has an alternate solution, I am happy to test it out.

### DIFF
--- a/fuse_constraints/CMakeLists.txt
+++ b/fuse_constraints/CMakeLists.txt
@@ -139,6 +139,32 @@ if(CATKIN_ENABLE_TESTING)
       CXX_STANDARD_REQUIRED YES
   )
 
+  # Absolute Orientation 2D Stamped Constraint Tests
+  catkin_add_gtest(test_absolute_orientation_2d_stamped_constraint
+    test/test_absolute_orientation_2d_stamped_constraint.cpp
+  )
+  add_dependencies(test_absolute_orientation_2d_stamped_constraint
+    ${catkin_EXPORTED_TARGETS}
+  )
+  target_include_directories(test_absolute_orientation_2d_stamped_constraint
+    PRIVATE
+      include
+      ${catkin_INCLUDE_DIRS}
+      ${CERES_INCLUDE_DIRS}
+      ${EIGEN3_INCLUDE_DIRS}
+  )
+  target_link_libraries(test_absolute_orientation_2d_stamped_constraint
+    ${PROJECT_NAME}
+    ${catkin_LIBRARIES}
+    ${CERES_LIBRARIES}
+    ${EIGEN3_LIBRARIES}
+  )
+  set_target_properties(test_absolute_orientation_2d_stamped_constraint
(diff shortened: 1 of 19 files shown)


carlos-m159 (approved):
LGTM

The only other option I can think of is to create a custom type and then establish your own rules for assignments, such as:

```
MyCustomAngleType& operator=(const double& yaw)
{
  this->val_ = fuse_core::wrapAngle2D(yaw);
  return *this;
 }
```

DavidLocus (approved):
LGTM

PR #337, opened 2023-09-17
Fix openmp compile error

jakemclaughlin6:
Fixes the following error when compiling:

```
/usr/bin/ld: /usr/local/lib/libceres.a(local_parameterization.cc.o): undefined reference to symbol 'omp_get_max_threads@@OMP_1.0'
/usr/bin/ld: /lib/x86_64-linux-gnu/libgomp.so.1: error adding symbols: DSO missing from command line
```

### DIFF
--- a/fuse_optimizers/CMakeLists.txt
+++ b/fuse_optimizers/CMakeLists.txt
@@ -12,6 +12,8 @@ set(build_depends
   std_srvs
 )
 
+find_package(OpenMP)
+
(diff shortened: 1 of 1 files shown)


svwilliams:
Do you know if this is still an issue? Nothing in fuse_optimizers uses OpenMP directly. So the issue must be some missing dependency somewhere. I know it's been forever since this patch was submitted.

PR #341, opened 2023-10-18
Support gcc12 and ceres 2.1.0

efernandez:
This MR has all the changes required to support `gcc12` and Ceres `2.1.0`.

```bash
$ g++ --version
g++ (GCC) 12.2.0
```

With Ceres `2.1.0` the `ceres::LocalParameterization` was [marked as deprecated](https://github.com/ceres-solver/ceres-solver/commit/0141ca090c315db2f3c38e1731f0fe9754a4e4cc) in favour of the new `ceres::Manifold` that implements the `Plus` and `Minus` methods; in Ceres `2.2.0` the `ceres::LocalParameterization` is [already removed](https://github.com/ceres-solver/ceres-solver/commit/68c53bb39552cd4abfd6381df08638285f7386b3).

`gcc12` spotted some issues:
* An out of bounds access in the `CostFunctor` defined for the `Poin2DFixedLandmark` unit test
* A `maybe-unitialized` warning in `fuse_graphs` related to `boost::make_transform_iterator` and a lambda that takes a reference from constraints or variables. I simply added `-Wno-maybe-uninitialized` for now
* A `array-bounds` warning due to [this `gcc12` bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106247), also mentioned in [Eigen issue](https://gitlab.com/libeigen/eigen/-/issues/2506). I simply added `-Wno-array-bounds`, conditional on the compiler being `gcc` and `>=12.0`.

I also added `SYSTEM` to 3rd party headers to ignore warnings from those libraries.

The `Minus` **Jacobian** implementation of `ceres::Manifold` only [computes the jacobian wrt the 1st argument](https://github.com/ceres-solver/ceres-solver/blob/77497373193a6304a67d0/include/ceres/autodiff_manifold.h#L244-L246), which differs from the `fuse_core::LocalParameterization` implementation, that [computes it wrt the 2nd argument](https://github.com/locusrobotics/fuse/blob/c8ab2f6e971a21ecf74e9debeef0d74d1c24655d/fuse_core/include/fuse_core/autodiff_local_parameterization.h#L214-L215), which seems to be wrong. For that reason, I fixed the following bugs:
* Fixed the `fuse_core::AutoDiffLocalParameterization`
* Used `ceres::AutoDiffManifold` instead of `fuse_core::AutoDiffLocalParameterization` for Ceres `>=2.1.0`
* Updated the unit tests that had hard-coded expected Jacobians. They basically changed in signed. This only impacted the orientation 2D and 3D local parameterizations, which are the only ones provided.
* Fixed the analytic `Minus` Jacobians for the `Orientation2DLocalParameterization` and `Orientation3DLocalParameterization`, since they had the wrong sign. This bug was hiden by the bug in `fuse_core::AutoDiffParameterization`, which in a way was cancelling it out.

### DIFF
--- a/fuse_constraints/CMakeLists.txt
+++ b/fuse_constraints/CMakeLists.txt
@@ -105,6 +105,9 @@ add_dependencies(${PROJECT_NAME}
 target_include_directories(${PROJECT_NAME}
   PUBLIC
     include
+)
+target_include_directories(${PROJECT_NAME}
+  SYSTEM PUBLIC
     ${catkin_INCLUDE_DIRS}
     ${CERES_INCLUDE_DIRS}
     ${EIGEN3_INCLUDE_DIRS}
--- a/fuse_constraints/CMakeLists.txt
+++ b/fuse_constraints/CMakeLists.txt
@@ -35,7 +35,47 @@ catkin_package(
 ###########
 ## Build ##
 ###########
-add_compile_options(-Wall -Werror)
+
+# Disable warnings about array bounds with -Wno-array-bounds until gcc 12 fixes this bug:
+# https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106247
+#
+# Also reported in Eigen, and confirmed to be due to gcc 12:
+# https://gitlab.com/libeigen/eigen/-/issues/2506
+#
+# In file included from include/immintrin.h:43,
+#                  from include/eigen3/Eigen/src/Core/util/ConfigureVectorization.h:361,
+#                  from include/eigen3/Eigen/Core:22,
+#                  from include/fuse_core/fuse_macros.h:63,
+#                  from include/fuse_core/loss.h:37,
+#                  from include/fuse_core/constraint.h:37,
+#                  from src/fuse/fuse_constraints/include/fuse_constraints/relative_orientation_3d_stamped_constraint.h:37,
+#                  from src/fuse/fuse_constraints/src/relative_orientation_3d_stamped_constraint.cpp:34:
+# In function ‘__m256d _mm256_loadu_pd(const double*)’,
+#     inlined from ‘Packet Eigen::internal::ploadu(const typename unpacket_traits<T>::type*) [with Packet = __vector(4) double]’ at include/eigen3/Eigen/src/Core/arch/AVX/PacketMath.h:582:129,
+#     inlined from ‘Packet Eigen::internal::ploadt(const typename unpacket_traits<T>::type*) [with Packet = __vector(4) double; int Alignment = 0]’ at include/eigen3/Eigen/src/Core/GenericPacketMath.h:969:26,
+#     inlined from ‘PacketType Eigen::internal::evaluator<Eigen::PlainObjectBase<Derived> >::packet(Eigen::Index) const [with int LoadMode = 0; PacketType = __vector(4) double; Derived = Eigen::Matrix<double, 3, 1>]’ at include/eigen3/Eigen/src/Core/CoreEvaluators.h:245:40,
+#     inlined from ‘void Eigen::internal::generic_dense_assignment_kernel<DstEvaluatorTypeT, SrcEvaluatorTypeT, Functor, Version>::assignPacket(Eigen::Index) [with int StoreMode = 32; int LoadMode = 0; PacketType = __vector(4) double; DstEvaluatorTypeT = Eigen::internal::evaluator<Eigen::Map<Eigen::Matrix<double, -1, 1> > >; SrcEvaluatorTypeT = Eigen::internal::evaluator<Eigen::Matrix<double, 3, 1> >; Functor = Eigen::internal::assign_op<double, double>; int Version = 0]’ at include/eigen3/Eigen/src/Core/AssignEvaluator.h:681:114,
+#     inlined from ‘static void Eigen::internal::dense_assignment_loop<Kernel, 3, 0>::run(Kernel&) [with Kernel = Eigen::internal::generic_dense_assignment_kernel<Eigen::internal::evaluator<Eigen::Map<Eigen::Matrix<double, -1, 1> > >, Eigen::internal::evaluator<Eigen::Matrix<double, 3, 1> >, Eigen::internal::assign_op<double, double>, 0>]’ at include/eigen3/Eigen/src/Core/AssignEvaluator.h:437:75,
+#     inlined from ‘void Eigen::internal::call_dense_assignment_loop(DstXprType&, const SrcXprType&, const Functor&) [with DstXprType = Eigen::Map<Eigen::Matrix<double, -1, 1> >; SrcXprType = Eigen::Matrix<double, 3, 1>; Functor = assign_op<double, double>]’ at include/eigen3/Eigen/src/Core/AssignEvaluator.h:785:37,
+#     inlined from ‘static void Eigen::internal::Assignment<DstXprType, SrcXprType, Functor, Eigen::internal::Dense2Dense, Weak>::run(DstXprType&, const SrcXprType&, const Functor&) [with DstXprType = Eigen::Map<Eigen::Matrix<double, -1, 1> >; SrcXprType = Eigen::Matrix<double, 3, 1>; Functor = Eigen::internal::assign_op<double, double>; Weak = void]’ at include/eigen3/Eigen/src/Core/AssignEvaluator.h:954:31,
+#     inlined from ‘void Eigen::internal::call_assignment_no_alias(Dst&, const Src&, const Func&) [with Dst = Eigen::Map<Eigen::Matrix<double, -1, 1> >; Src = Eigen::Matrix<double, 3, 1>; Func = assign_op<double, double>]’ at include/eigen3/Eigen/src/Core/AssignEvaluator.h:890:49,
+#     inlined from ‘void Eigen::internal::call_assignment(Dst&, const Src&, const Func&, typename enable_if<evaluator_assume_aliasing<Src>::value, void*>::type) [with Dst = Eigen::Map<Eigen::Matrix<double, -1, 1> >; Src = Eigen::Product<Eigen::Matrix<double, 3, 3, 1>, Eigen::Map<Eigen::Matrix<double, -1, 1> >, 0>; Func = assign_op<double, double>]’ at include/eigen3/Eigen/src/Core/AssignEvaluator.h:851:27,
+#     inlined from ‘void Eigen::internal::call_assignment(Dst&, const Src&) [with Dst = Eigen::Map<Eigen::Matrix<double, -1, 1> >; Src = Eigen::Product<Eigen::Matrix<double, 3, 3, 1>, Eigen::Map<Eigen::Matrix<double, -1, 1> >, 0>]’ at include/eigen3/Eigen/src/Core/AssignEvaluator.h:836:18,
+#     inlined from ‘Derived& Eigen::MatrixBase<Derived>::operator=(const Eigen::DenseBase<OtherDerived>&) [with OtherDerived = Eigen::Product<Eigen::Matrix<double, 3, 3, 1>, Eigen::Map<Eigen::Matrix<double, -1, 1> >, 0>; Derived = Eigen::Map<Eigen::Matrix<double, -1, 1> >]’ at include/eigen3/Eigen/src/Core/Assign.h:66:28,
+#     inlined from ‘void Eigen::EigenBase<Derived>::applyThisOnTheLeft(Dest&) const [with Dest = Eigen::Map<Eigen::Matrix<double, -1, 1> >; Derived = Eigen::Matrix<double, 3, 3, 1>]’ at include/eigen3/Eigen/src/Core/EigenBase.h:114:9:
+# include/avxintrin.h:893:24: error: array subscript ‘__m256d_u[0]’ is partly outside array bounds of ‘Eigen::internal::plain_matrix_type<Eigen::Product<Eigen::Matrix<double, 3, 3, 1>, Eigen::Map<Eigen::Matrix<double, -1, 1> >, 0>, Eigen::Dense>::type [1]’ {aka ‘Eigen::Matrix<double, 3, 1> [1]’} [-Werror=array-bounds]
+#   893 |   return *(__m256d_u *)__P;
+#       |                        ^~~
+# In file included from include/eigen3/Eigen/Core:278:
+# include/eigen3/Eigen/src/Core/AssignEvaluator.h: In member function ‘void Eigen::EigenBase<Derived>::applyThisOnTheLeft(Dest&) const [with Dest = Eigen::Map<Eigen::Matrix<double, -1, 1> >; Derived = Eigen::Matrix<double, 3, 3, 1>]’:
+# include/eigen3/Eigen/src/Core/AssignEvaluator.h:850:41: note: at offset [0, 16] into object ‘tmp’ of size 24
+#   850 |   typename plain_matrix_type<Src>::type tmp(src);
+#       |                                         ^~~
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 12.0)
+  add_compile_options(-Wall -Werror -Wno-array-bounds)
+else()
+  add_compile_options(-Wall -Werror)
+endif()
 
 # fuse_constraints library
 add_library(${PROJECT_NAME}
--- a/fuse_constraints/include/fuse_constraints/marginal_constraint.h
+++ b/fuse_constraints/include/fuse_constraints/marginal_constraint.h
@@ -36,9 +36,9 @@
 
 #include <fuse_core/constraint.h>
 #include <fuse_core/eigen.h>
+#include <fuse_core/fuse_macros.h>
 #include <fuse_core/local_parameterization.h>
 #include <fuse_core/manifold.h>
-#include <fuse_core/fuse_macros.h>
 #include <fuse_core/serialization.h>
 #include <fuse_core/variable.h>
 
@@ -58,10 +58,8 @@
 #include <string>
 #include <vector>
 
-
 namespace fuse_constraints
 {
-
 /**
  * @brief A constraint that represents remaining marginal information on a set of variables
  *
@@ -94,7 +92,7 @@ public:
    * @param[in] last_A         Iterator pointing to one past the last A matrix
    * @param[in] b              The b vector of the marginal cost (of the form A*(x - x_bar) + b)
    */
-  template<typename VariableIterator, typename MatrixIterator>
+  template <typename VariableIterator, typename MatrixIterator>
   MarginalConstraint(
     const std::string& source,
     VariableIterator first_variable,
@@ -131,14 +129,11 @@ public:
   {
     return local_parameterizations_;
   }
-#else 
+#else
   /**
    * @brief Read-only access to the variable local parameterizations
    */
-  const std::vector<fuse_core::Manifold::SharedPtr>& manifolds() const
-  {
-    return manifolds_;
-  }
+  const std::vector<fuse_core::Manifold::SharedPtr>& manifolds() const { return manifolds_; }
 #endif
 
   /**
@@ -179,24 +174,23 @@ private:
    * @param[in/out] archive - The archive object that holds the serialized class members
    * @param[in] version - The version of the archive being read/written. Generally unused.
    */
-  template<class Archive>
+  template <class Archive>
   void serialize(Archive& archive, const unsigned int /* version */)
   {
-    archive & boost::serialization::base_object<fuse_core::Constraint>(*this);
-    archive & A_;
-    archive & b_;
+    archive& boost::serialization::base_object<fuse_core::Constraint>(*this);
+    archive& A_;
+    archive& b_;
 #if !CERES_VERSION_AT_LEAST(2, 1, 0)
-    archive & local_parameterizations_;
+    archive& local_parameterizations_;
 #else
-    archive & manifolds_;
+    archive& manifolds_;
 #endif
-    archive & x_bar_;
+    archive& x_bar_;
   }
 };
 
 namespace detail
 {
-
 /**
  * @brief Return the UUID of the provided variable
  */
@@ -233,7 +227,7 @@ inline fuse_core::Manifold::SharedPtr const getManifold(const fuse_core::Variabl
 
 }  // namespace detail
 
-template<typename VariableIterator, typename MatrixIterator>
+template <typename VariableIterator, typename MatrixIterator>
 MarginalConstraint::MarginalConstraint(
   const std::string& source,
   VariableIterator first_variable,
@@ -241,40 +235,39 @@ MarginalConstraint::MarginalConstraint(
   MatrixIterator first_A,
   MatrixIterator last_A,
   const fuse_core::VectorXd& b) :
-    Constraint(source,
-               boost::make_transform_iterator(first_variable, &fuse_constraints::detail::getUuid),
-               boost::make_transform_iterator(last_variable, &fuse_constraints::detail::getUuid)),
-    A_(first_A, last_A),
-    b_(b),
+  Constraint(
+    source,
+    boost::make_transform_iterator(first_variable, &fuse_constraints::detail::getUuid),
+    boost::make_transform_iterator(last_variable, &fuse_constraints::detail::getUuid)),
+  A_(first_A, last_A),
+  b_(b),
 #if !CERES_VERSION_AT_LEAST(2, 1, 0)
-    local_parameterizations_(boost::make_transform_iterator(first_variable,
-                                                            &fuse_constraints::detail::getLocalParameterization),
-                             boost::make_transform_iterator(last_variable,
-                                                            &fuse_constraints::detail::getLocalParameterization)),
-#else 
-    manifolds_(boost::make_transform_iterator(first_variable,
-                                              &fuse_constraints::detail::getManifold),
-               boost::make_transform_iterator(last_variable,
-                                              &fuse_constraints::detail::getManifold)),
+  local_parameterizations_(
+    boost::make_transform_iterator(first_variable, &fuse_constraints::detail::getLocalParameterization),
+    boost::make_transform_iterator(last_variable, &fuse_constraints::detail::getLocalParameterization)),
+#else
+  manifolds_(
+    boost::make_transform_iterator(first_variable, &fuse_constraints::detail::getManifold),
+    boost::make_transform_iterator(last_variable, &fuse_constraints::detail::getManifold)),
 #endif
-    x_bar_(boost::make_transform_iterator(first_variable, &fuse_constraints::detail::getCurrentValue),
-           boost::make_transform_iterator(last_variable, &fuse_constraints::detail::getCurrentValue))
+  x_bar_(
+    boost::make_transform_iterator(first_variable, &fuse_constraints::detail::getCurrentValue),
+    boost::make_transform_iterator(last_variable, &fuse_constraints::detail::getCurrentValue))
 {
   assert(!A_.empty());
   assert(A_.size() == x_bar_.size());
 #if !CERES_VERSION_AT_LEAST(2, 1, 0)
   assert(A_.size() == local_parameterizations_.size());
-#else 
+#else
   assert(A_.size() == manifolds_.size());
 #endif
   assert(b_.rows() > 0);
-  assert(std::all_of(A_.begin(), A_.end(), [this](const auto& A){ return A.rows() == this->b_.rows(); }));  // NOLINT
-  assert(std::all_of(boost::make_zip_iterator(boost::make_tuple(A_.begin(), first_variable)),
-                     boost::make_zip_iterator(boost::make_tuple(A_.end(), last_variable)),
-                     [](const boost::tuple<const fuse_core::MatrixXd&, const fuse_core::Variable&>& tuple)  // NOLINT
-                     {
-                       return static_cast<size_t>(tuple.get<0>().cols()) == tuple.get<1>().localSize();
-                     }));  // NOLINT
+  assert(std::all_of(A_.begin(), A_.end(), [this](const auto& A) { return A.rows() == this->b_.rows(); }));  // NOLINT
+  assert(std::all_of(
+    boost::make_zip_iterator(boost::make_tuple(A_.begin(), first_variable)),
+    boost::make_zip_iterator(boost::make_tuple(A_.end(), last_variable)),
+    [](const boost::tuple<const fuse_core::MatrixXd&, const fuse_core::Variable&>& tuple)  // NOLINT
+    { return static_cast<size_t>(tuple.get<0>().cols()) == tuple.get<1>().localSize(); }));  // NOLINT
 }
 
 }  // namespace fuse_constraints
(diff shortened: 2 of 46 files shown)


fuse_constraints/CMakeLists.txt
```diff
@@ -35,7 +35,47 @@ catkin_package(
 ###########
 ## Build ##
 ###########
-add_compile_options(-Wall -Werror)
+
+# Disable warnings about array bounds with -Wno-array-bounds until gcc 12 fixes this bug:
```
ayrton04, line 39: Would it make sense to file a ticket with this error and then reference it here?

fuse_constraints/test/test_marginal_constraint.cpp
```diff
@@ -207,11 +207,11 @@ TEST(MarginalConstraint, LocalParameterization)
   fuse_core::Vector1d expected_residuals;
   expected_residuals << 10.581088914;  // 5.0 * 0.15  +  6.0 * -0.2  +  7.0 * 0.433012702   +   8.0
   fuse_core::MatrixXd expected_jacobian1(1, 4);
-  expected_jacobian1 << -13.29593, 3.86083, 9.164586, 12.818822;
+  expected_jacobian1 << 13.29593, -3.86083, -9.164586, -12.818822;
```
ayrton04: Why would this test result change?

ayrton04 (approved):

fuse_constraints/test/test_marginal_constraint.cpp
ayrton04: Scratch that, missed this bullet point in your PR description.

efernandez: Yep, it's explained in the description. :smiley:

fuse_constraints/CMakeLists.txt
efernandez: There's a gcc bug and Eigen issue already reported and reference here.

fuse_graphs/CMakeLists.txt
```diff
@@ -28,7 +28,28 @@ catkin_package(
 ###########
 ## Build ##
 ###########
-add_compile_options(-Wall -Werror)
+
+# Disable warnings about maybe uninitialized variables with -Wno-maybe-uninitialized until we fix the following error:
```
efernandez, line 32: @ayrton04 For this one though, I don't know if there's any issue already reported in Boost.

The thing is that I'm not even sure if the warning is caused by `boost::iterators` or the lamdba we're using. :thinking:

fuse_constraints/test/test_marginal_constraint.cpp
svwilliams: I haven't done a ton of digging into the Ceres Manifold class, but I suspect Ceres implements the Minus() function differently than I do.

https://github.com/ceres-solver/ceres-solver/blob/2.1.0/include/ceres/autodiff_manifold.h#L166-L170
```
  bool Minus(const double* y,
             const double* x,
             double* y_minus_x) const override
```
Based on the parameter names, I assume this results in `y - x` for a linear system.

Whereas the Minus function in fuse is written such that:
https://github.com/locusrobotics/fuse/blob/devel/fuse_core/include/fuse_core/local_parameterization.h#L62-L79
```
if Plus(x1, delta) -> x2, then Minus(x1, x2) -> delta
```
which using Ceres variable names would be
```
  bool Minus(const double* x,
             const double* y,
             double* y_minus_x) const override
```
The reason, such that there is one, for this variable order is that the general form of "subtraction" would be:
X^-1 * Y. For example, when doing frame transformations you would have T1^1 * T2 to compute "T2 - T1".


That would explain why I'm taking the derivative with respect to the second term and not the first. I'm not sure if it explains the sign change or not.

efernandez: That could also explain the difference between `fuse` and `ceres` local parameterization/manifold implementation of `Minus` and its jacobian.

I've just added a couple of new test cases to the local parameterization unit test: https://github.com/locusrobotics/fuse/pull/341/commits/9d2d148da3cb8e3c4c1f603145ff2a656601d474

I'm basically testing the conditions mentioned in https://github.com/locusrobotics/fuse/blob/c8ab2f6e971a21ecf74e9debeef0d74d1c24655d/fuse_core/include/fuse_core/local_parameterization.h#L68-L69, that is:
* `Minus(x, x) -> 0`
* `if Plus(x1, delta) -> x2, then Minus(x1, x2) -> delta`

The tests are passing with `gcc12` + `Ceres 2.1.0` and should also pass on older versions.

I think with this MR I actually changed your implementation to match the one in `ceres::Manifold`. :laughing: 

Is that ok, or do you think that shouldn't be changed?

efernandez:
Not sure why tests are failing now. I run all tests with both configurations, and all tests passed in both cases. That is, gcc `11.3.0` with Ceres `2.0` and gcc `12.2.0` with Ceres `2.1`.

If you share the output errors with me, I can have a look.

fuse_constraints/test/test_marginal_constraint.cpp
svwilliams: In the grand scheme of things, I want the fuse version to match the Ceres version. It keeps things much easier.
My worry here is about backwards compatibility. Updating the two local parameterizations contained within the fuse repo is easy. But if anyone has created their own local parameterization classes, then strange things are going to happen.

What would be great is if we can make a translation layer that warns the old methods are dead, and that the new methods must be implemented. But I'm not sure how possible that will be without digging into this MR and the Ceres 2.1 changes in more detail.

svwilliams:
Looks like some transient issues we are having with our own build server. I don't think it is related to this MR at all.

fuse_constraints/test/test_marginal_constraint.cpp
efernandez: Actually, I just realized I didn't really change the `Minus` implementation of `Orientation2DLocalParameterization` and `Orientation3DLocalParameterization`.

In the corresponding unit test for those local parameterization we also have functors for the **minus** implementations, but they're also the same as the ones in the classes. So I didn't really change that.

I only changed the analytic and autodiff jacobians, so they match the ones in `ceres::Manifold`.

jakemclaughlin6: @svwilliams I'm going to be working on refactoring this a bit so that it can be backwards compatible with ceres. My thoughts are to implement a fuse wrapper around the new manifold, and adding deprecated attributes to all localparameterization stuff.

The general logic would be: 
1. if ceres < 2.1 -> use localParameterization, manifold not compiled
2. if ceres > 2.1 -> use both manifold and localParameterizations where available (favour manifold), update internal fuse variables to use manifold
3. if ceres > 2.2 -> use manfiold, localParameterization not compiled

Would this work to get a change like this upstream?

svwilliams: > My thoughts are to implement a fuse wrapper

Can you expand on that a bit?

The fuse `LocalParameterization` class is a wrapper around the Ceres Solver `LocalParameterization` class.
```
class LocalParameterization : public ceres::LocalParameterization
```

Are you proposing doing something similar with the manifolds?
```
class Manifold : public ceres::Manifold
```
So that fuse would contain a wrapper class for both `ceres::LocalParameterization` and `ceres::Manifold`?

And then every derived `fuse_core::Variable` class would need to be updated to use the `fuse::Manifold` instead of the `fuse::LocalParameterization` based on the detected version of Ceres Solver?
https://github.com/locusrobotics/fuse/blob/devel/fuse_core/include/fuse_core/variable.h#L315

Or did I misunderstand the proposal completely?

jakemclaughlin6: Yes thats spot on, then i'd add deprecated flags to whatever needs to be deprecated based on the ceres version.

svwilliams: I worried this will create a required API change for any external packages that exist out in the world. And further, that the API change is dependent not on the version of Fuse or ROS, but on the version of one of the Fuse dependencies. And that feels odd.

I haven't worked out the details of the following idea. It might not be easy or even possible. But what if we:
* Create the Fuse wrapper class for the Ceres Manifold class: `class Manifold : public ceres::Manifold`
* Change the Fuse LocalParameterization to derive from the Fuse Manifold class: `class LocalParameterization : public fuse_core::Manifold`
* Modify the implementation of the Fuse LocalParameterization class such that the public API remains the same, but internally it mutates methods into the format expected by the Fuse Manifold class. This is the bit that I don't know if it is feasible.
* Modify the Fuse Variable base class to accept a Fuse Manifold pointer instead of a Fuse LocalParameterization. Since the Fuse LocalParameterization class is now derived from the Fuse Manifold, all existing code should still work.

All of those changes would need to be conditional on the Ceres Solver version >= 2.1. But that keeps the scope of conditional compilation down to two classes (LocalParameterization and Manifold), which seems easier than modifying every concrete Variable implementation.

Does that sound workable at all? Do you see any holes in that plan?

jakemclaughlin6: I'm thinking that would work but maybe with this slight tweak:
1. Implement a fuse Manifold (extend ceres manifold) as specified, and add the required functions so that it's API is the same as ceres localParameterization
2. fuse::LocalParameterization either extends the fuse::Manifold (if ceres >= 2.1) or it extends ceres::LocalParameterization (if ceres < 2.1) and would be implementing the exact same functions in either case

Then no change is required for the Fuse Variable. I'll try to implement something like this quickly and get your thoughts on it so its more clear

svwilliams: > fuse::LocalParameterization either extends the fuse::Manifold (if ceres >= 2.1) or it extends ceres::LocalParameterization (if ceres < 2.1)

Agree completely.

> Implement a fuse Manifold (extend ceres manifold) as specified, and add the required functions so that it's API is the same as ceres localParameterization

I'm not sure the `[fuse::Manifold] API is the same as ceres localParameterization` is necessarily required. I _think_ the main difference is the Ceres Manifold class adds a `Minus()` implementation, something that the Fuse LocalParameterization class has always had. The issue is that I defined the Fuse `Minus()` function slightly differently than the Ceres Manifold `Minus()` function. And it might be nice to sync up the Fuse Manifold API with the Ceres Manifold API as much as possible to avoid confusion. But that means the Fuse LocalParameterization would need to translate the Fuse LocalParameterization function into a Ceres Manifold function.

Regardless, starting a quick implementation sounds great. Then we can start having discussions about concrete concerns rather than theory and supposition. Thanks for this.

jakemclaughlin6: This is how ceres defines the minus function for manifold:

  // Given two points on the manifold, Minus computes the change to x in the
  // tangent space at x, that will take it to y.

So basically finding delta such that: x + delta = y, which is exactly how it's defined in fuse:

https://github.com/locusrobotics/fuse/blob/devel/fuse_core/include/fuse_core/local_parameterization.h#L68-L69 

SO I don't think any conversions need to be done, but maybe just an updating on the documentation?

jakemclaughlin6: Just so I can get it straight: The main difference is just the ordering of the two variables in the minus function?

i.e. ceres has Minus as:
`minus(x2, x1) -> x2_minus_x1`

and fuse has it as:
`minus(x1, x2) -> x2_minus_x1`

I don't think it'll be possible to automatically convert between the two since the function signature is the exact same. But I could match the convention of ceres Manifold for the fuse Manifold so that users know for the future. And I don't think any existing code outside of fuse would break since as long as the jacobian for minus is implemented correctly for the operation it shouldn't matter?

svwilliams: I believe the input parameters are reversed the Fuse `Minus()` implementation versus the Ceres `Minus()` implementation.

In Fuse:
https://github.com/locusrobotics/fuse/blob/devel/fuse_core/include/fuse_core/local_parameterization.h#L68-L69

```
   *  - Minus(x, x) -> 0
   *  - if Plus(x1, delta) -> x2, then Minus(x1, x2) -> delta
```
Or, to write that a slightly different way
```
if Plus(x, delta) -> x_plus_delta, then Minus(x, x_plus_delta) -> delta
```

Whereas in Ceres:
https://github.com/ceres-solver/ceres-solver/blob/master/include/ceres/manifold.h#L77-L78

```
// 1. x_plus_delta = Plus(x, delta)
// 2. delta = Minus(x_plus_delta, x)
```

jakemclaughlin6: Maybe, I add a deprecated attribute to the local parameterization Minus and ComputeMinusJacobian implementation warning of the change in convention

svwilliams: > I don't think it'll be possible to automatically convert between the two since the function signature is the exact same.

It took me a minute, but I eventually understand your point.

What about using an Adapter Pattern?
* Create the Fuse Manifold wrapper class: `class fuse_core::Manifold : public ceres::Manifold`
* Create a ManifoldAdapter class that inherits from Manifold: `class ManifoldAdapter : public fuse_core::Manifold`
* That adapter class would contain an old LocalParameterization object: `class ManifoldAdapter { private: fuse_core::LocalParameterization local_parameterization_; }`
* And the adapter would implement the Manifold API using the LocalParameterization member object: `Minus(x, y) { return local_parameterization_->Minus(y, x); }`
* Add a manifold getter virtual method to the Variable base class that returns the LocalParameterization pointer wrapped in the LocalParameterizationAdapter. This allows any derived Variable to override the manifold getter function to return a real Manifold object.

```
class Manifold : public ceres::Manifold
{
  virtual double Minus(x, y);
  ...
}

class ManifoldAdapter : public fuse_core::Manifold
{
public:
  ManifoldAdapter(LocalParameterization* local_parameterization) : local_parameterization_(local_parameterization) {}

  double Minus(x, y) override
  {
    return local_parameterization_->Minus(y, x);
  }
private:
  fuse_core::LocalParameterization* local_parameterization_;
}

class Variable
{
  // Existing code
  virtual fuse_core::LocalParameterization* localParameterization() const
  {
    return nullptr;
  }

  // New Code
  virtual fuse_core::Manifold* manifold() const
  {
    return new ManifoldAdapter(this->localParameterization());
  }
}

class LegacyVariable : public Variable
{
public:
  fuse_core::LocalParameterization* localParameterization() const override
  {
    return new MyParameterization();
  }
}

class NewVariable : public Variable
{
public:
  fuse_core::Manifold* manifold() const override
  {
    return new MyManifold();
  }
}
```

* `LegacyVariable` still works because the default implementation of `manifold()` in the base class will construct a `Manifold` pointer using the `ManifoldAdapter`.
* `NewVariable` also works because the derived implementation of `manifold()` replaces the base class version and returns a `Manifold` pointer without all the extra indirection.
* If `NewVariable` also overrides the `localParameterization()` function, that override will never be used. I'm not sure how to solve that other than with documentation and maybe some pragma warnings during compilation.

Thoughts?

jakemclaughlin6: I think the problem I see is that this attempts to keep both LocalParameterization and Manifold, where once ceres is >= 2.2, LocalParameterization will be completely gone so this solution would also include lots of conditional compilations.

I'm not sure if I agree that the changing in conventions for the minus requires drastic measures to alleviate. Since the minus function would be implemented by the user,and the jacobian along with it, even if there were to mix up the inputs, but their jacobian also mixed up the inputs, the result would be the same (i think). So i think some compile warnings when the user switches to ceres >= 2.1 that the convention has changed would be enough.

svwilliams: > so this solution would also include lots of conditional compilations.

I believe it will only require conditional compilation of the Manifold, ManifoldAdapter, LocalParameterization, and Variable classes. And all of that is contained inside of the `fuse_core` package. So at least all of the ugliness would be contained.

> Since the minus function would be implemented by the user,and the jacobian along with it, even if there were to mix up the inputs, but their jacobian also mixed up the inputs, the result would be the same (i think).

If you can prove that is true, then I'll happily accept a simpler alternative.

jakemclaughlin6: There will be some ugliness in the marginal cost function, having to store both Manifolds and LocalParameterizations. But this is doable, I'll work on this today

jakemclaughlin6: @svwilliams is there a clang format file I can use to format to this repo's style?

svwilliams: I wrote most of the fuse code before I had clang-format configured. As such, any config file I give you will result in unintended format changes.

But here is my personal, over-engineered clang format file...renamed to make github happy. If that helps you any.
[clang-format.txt](https://github.com/locusrobotics/fuse/files/14184999/clang-format.txt)

jakemclaughlin6:
The general logic for the conditional compilation is as follows:

- If ceres is < 2.1, everything remains the same and LocalParameterization is used
- If ceres == 2.1, LocalParameterization is still supported for fuse::Variables, but everything in fuse switches to using Manifold, fuse::Variables implement a manifold from their existing LocalParameterization using the ManifoldAdapter
- If ceres > 2.1, LocalParameterization is completely removed, along with the ManifoldAdapter, everything must implement manifold directly

fuse_constraints/test/test_marginal_constraint.cpp
jakemclaughlin6: All tests are passing now, let me know what you think of the solution, and if I should add some compile warnings/deprecated attributes to anything

fuse_constraints/include/fuse_constraints/marginal_constraint.h
```diff
@@ -122,13 +121,20 @@ class MarginalConstraint : public fuse_core::Constraint
    */
   const std::vector<fuse_core::VectorXd>& x_bar() const { return x_bar_; }
 
+#if !CERES_VERSION_AT_LEAST(2, 1, 0)
```
svwilliams: I wonder if we should define a macro variable with an easier to read name? Like:
```
#define CERES_SUPPORTS_MANIFOLDS CERES_VERSION_AT_LEAST(2, 1, 0)
...
#if CERES_SUPPORTS_MANIFOLDS
  std::vector<fuse_core::LocalParameterization::SharedPtr> local_parameterizations_;  //!< Parameterizations
#else
  std::vector<fuse_core::Manifold::SharedPtr> manifolds_;  //!< Manifolds
#endif
```

fuse_constraints/include/fuse_constraints/marginal_constraint.h
```diff
@@ -164,20 +174,23 @@ class MarginalConstraint : public fuse_core::Constraint
...
-    archive & local_parameterizations_;
-    archive & x_bar_;
+    archive& boost::serialization::base_object<fuse_core::Constraint>(*this);
+    archive& A_;
+    archive& b_;
+#if !CERES_VERSION_AT_LEAST(2, 1, 0)
+    archive& local_parameterizations_;
+#else
+    archive& manifolds_;
+#endif
```
svwilliams: I need to think about this. This may break serialization between versions. And I'm not sure there is an obvious solution.

fuse_core/include/fuse_core/autodiff_local_parameterization.h
```diff
@@ -212,7 +212,9 @@ bool AutoDiffLocalParameterization<PlusFunctor, MinusFunctor, kGlobalSize, kLoca
   double delta[kLocalSize] = {};  // zero-initialize
 
   const double* parameter_ptrs[2] = {x, x};
-  double* jacobian_ptrs[2] = {NULL, jacobian};
+  // We only need to compute the Jacobian w.r.t. the first argument, see:
+  // https://github.com/ceres-solver/ceres-solver/blob/77497373193a6304a67d0/include/ceres/autodiff_manifold.h#L244-L246
+  double* jacobian_ptrs[2] = {jacobian, NULL};
```
svwilliams: This is part of the LocalParameterization code. Shouldn't this be unchanged now?

fuse_constraints/test/test_marginal_constraint.cpp
svwilliams: Well, now I'm confused about Tom's original question:

> Why would this test result change?

With the compile time toggle between LocalParameterization and Manifold classes, I would expect the LocalParameterization test cases would be unchanged.

fuse_core/include/fuse_core/local_parameterization.h
```diff
@@ -34,16 +34,20 @@
...
 #include <fuse_core/serialization.h>
 
 #include <boost/serialization/access.hpp>
-#include <ceres/local_parameterization.h>
 
+#if !CERES_VERSION_AT_LEAST(2, 2, 0)
+// Local parameterizations is removed in favour of Manifold in
+// version 2.2.0, see
+// https://github.com/ceres-solver/ceres-solver/commit/0141ca090c315db2f3c38e1731f0fe9754a4e4cc
+#include <ceres/local_parameterization.h>
```
svwilliams: I wonder if the LocalParameterization class should exist regardless of the Ceres version. For versions less than 2.1.0, it would remain as-is. And for versions >=2.1.0, it would no longer inherit from `ceres::LocalParameterization`. The ManifoldAdapter does not care about the Ceres base class, so it can continue to function indefinitely.

fuse_core/include/fuse_core/manifold.h
```diff
@@ -0,0 +1,147 @@
...
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_CORE_MANIFOLD_H
+#define FUSE_CORE_MANIFOLD_H
+
+#include <fuse_core/ceres_macros.h>
+#include <fuse_core/fuse_macros.h>
+#include <fuse_core/serialization.h>
+
+#include <boost/serialization/access.hpp>
```
svwilliams, line 43: All of the includes except `ceres_macros.h` can be inside the `#if CERES_VERSION_AT_LEAST(2, 1, 0)`
But I don't think it will affect anything in practice because you will be including the LocalParameterization header as well.

fuse_core/include/fuse_core/manifold_adapter.h
```diff
@@ -0,0 +1,138 @@
...
+{
+class ManifoldAdapter : public fuse_core::Manifold
+{
+public:
+  /**
+   * @brief Constructor to adapt a fuse::LocalParameterization into a fuse::Manifold
+   *
+   * @param[in] local_parameterization fuse::LocalParameterization
+   */
+  explicit ManifoldAdapter(LocalParameterization* local_parameterization) :
```
svwilliams: I believe the ManifoldAdapter needs to own the local parameterization pointer. Normally Ceres Solver takes ownership of the LocalParameterization objects and handles deleting them. The easy way is to make the `local_parameterization_` member variable a `unique_ptr`, and then the `unique_ptr` destructor will delete it when the ManifoldAdapter is deleted.

fuse_core/include/fuse_core/variable.h
```diff
@@ -291,31 +298,69 @@ class Variable
...
+  virtual fuse_core::Manifold* manifold() const
   {
+#if !CERES_VERSION_AT_LEAST(2, 2, 0)
+    if (!localParameterization())
+    {
+      return nullptr;
+    }
+    std::shared_ptr<fuse_core::Manifold> adapted_manifold =
+      std::make_shared<fuse_core::ManifoldAdapter>(localParameterization());
+    return adapted_manifold.get();
```
svwilliams: The `localParameterization()` should not be called twice. Each call is creating a new LocalParameterization instance.

Nor do I think wrapping the ManifoldAdapter in a shared_ptr is needed/a good idea. Calling `get()` returns the raw pointer _without updating the reference count_. Therefore the only usage of the shared_ptr is by the `adapted_manifold` local variable. When that goes out of scope at the end of this function call, the ManifoldAdapter _will be deleted_.

I think this should be something along the lines of:
```
auto local_parameterization = localParameterization();
if (!local_parameterization)
{
  return std::nullptr;
}
else
{
  return new fuse_core::ManifoldAdapter(local_parameterization);
}
```

fuse_core/test/test_local_parameterization.cpp
```diff
@@ -31,46 +31,68 @@
  *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  *  POSSIBILITY OF SUCH DAMAGE.
  */
+#include <fuse_core/ceres_macros.h>
+
+#if CERES_VERSION_AT_LEAST(2, 1, 0)
+#include <ceres/autodiff_manifold.h>
```
svwilliams: The purpose of this unit test is to verify the `fuse_core/autodiff_local_parameterization.h` was implemented correctly. Since Ceres Solver provides a `ceres/autodiff_manifold.h` itself and we didn't need to create our own, I'm not sure there is much motivation to test the Ceres AutoDiff implementation. There could just be a single `#if !CERES_VERSION_AT_LEAST(2, 1, 0)` at the top of the file, and that should be fine?

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
```diff
@@ -99,7 +101,13 @@ TEST(CostFunction, evaluateCostFunction)
 
   // Check jacobians are correct using a gradient checker
   ceres::NumericDiffOptions numeric_diff_options;
-  ceres::GradientChecker gradient_checker(&cost_function, NULL, numeric_diff_options);
+  ceres::GradientChecker gradient_checker(&cost_function,
+#if CERES_VERSION_AT_LEAST(2, 1, 0)
+      static_cast<const std::vector<const ceres::Manifold*>*>(nullptr),
+#else
+      static_cast<const std::vector<const ceres::LocalParameterization*>*>(nullptr),
+#endif
```
svwilliams: Does the GradientChecker not accept a single nullptr any longer?

svwilliams (changes requested):
> The general logic for the conditional compilation is as follows:
If ceres is < 2.1, everything remains the same and LocalParameterization is used
If ceres == 2.1, LocalParameterization is still supported for fuse::Variables, but everything in fuse switches to using Manifold, fuse::Variables implement a manifold from their existing LocalParameterization using the ManifoldAdapter
If ceres > 2.1, LocalParameterization is completely removed, along with the ManifoldAdapter, everything must implement manifold directly

This scheme still breaks backwards compatibility when upgrading to Ceres 2.2. For the most part, I don't think the "ceres > 2.1" case should exist. Either "ceres is < 2.1" is in use and everything remains the same, or "ceres is >= 2.1" is in use and the interfaces switch to use Manifolds, with the automatic ManifoldAdapter in place to handle "legacy" code.

The one place that 2.1 vs 2.2 matters would be in the `fuse_core::LocalParameterization` class itself. If "ceres <= 2.1" then we must inherit from `ceres::LocalParameterization`. If "ceres > 2.1", then we cannot inherit and must define the full interface inside of fuse.

Am I missing something that makes that option unworkable?

---

There are still some sign changes in a unit test that I do not understand. The LocalParameterization class should be left unchanged, so I would not expect any of the unit test results to change either.

---

There are a couple of issues with pointer ownership that need to be fixed. There should be inline comments for those.

---

I need to think through how the serialization works. It would be unpleasant if all of your old serialized graphs are unusable after you upgrade Ceres. But I can handle those concerns if you don't want to think about it.

fuse_constraints/test/test_marginal_constraint.cpp
jakemclaughlin6: I've updated that test to be conditional on if manifolds are used or not. It would seem the new Manifold computes the minus jacobian wrt the 1st argument, and LocalParameterization wrt the 2nd so this test. 

The marginal constraint test would change because although LocalParameterization is still supported past 2.1, it is only supported in a manner such that it can be used to generate a manifold.

This leads me to another question for ManifoldAdapter. While converting between the minus conventions is trivial, how can we convert the MinusJacobian? I am not sure/confident that the MinusJacobian is simply negated in every case

jakemclaughlin6: I think I finally understand, the jacobians shouldnt change between manifold and localparameterization because theyre computing wrt the same variable (they just change spots in the Minus function call). So the only change to the tests are which Minus function is called (which just flips the order of the parameters). So yes the localparameterization test remains the same.

jakemclaughlin6: I've also updated the local parameterization param naming convention to match the manifold so that its easier to compare across the two

fuse_constraints/include/fuse_constraints/marginal_constraint.h
jakemclaughlin6: Would the easiest solution be some sort of migration script? i.e. users can take their old graphs and migrate them into new graphs with manifolds? I don't think I have the bandwidth to take that on atm

svwilliams: We should be able to build the migration directly into the Boost deserialization code.

> I don't think I have the bandwidth to take that on atm

Not a problem. I can handle that. As soon as I get caught up again myself.

clalancette:
It would be great to get this PR merged; the use of `ceres::LocalParameterization` is currently causing fuse_core to fail to build on Ubuntu 24.04 in ROS 2 Rolling: https://build.ros2.org/view/Rbin_uN64/job/Rbin_uN64__fuse_core__ubuntu_noble_amd64__binary/9/console

svwilliams:
Yep, it's just not ready yet. I'm hoping to get some time next week to finish this up, and then port it over to ROS 2.

svwilliams:
I've finished up the work on serializing the Manifolds. I didn't have permission to push to this fork/branch, so I created a new PR here for safe-keeping: #363

Once this is reviewed by my team, my intention is to merge this PR then rebase #363 on top of it. That should hopefully keep the contributions clear.

Let me know if you have any concerns on the changes that were made. And thanks for doing the vast majority of the work here.
I'll be starting on a ROS 2 port of this work next week.

jakemclaughlin6:
That works for me. The changes look fine to me, lets wait for @efernandez to see if its good with him as well

fuse_constraints/include/fuse_constraints/marginal_constraint.h
```diff
@@ -122,13 +121,20 @@ class MarginalConstraint : public fuse_core::Constraint
...
   /**
    * @brief Read-only access to the variable local parameterizations
    */
   const std::vector<fuse_core::LocalParameterization::SharedPtr>& localParameterizations() const
   {
     return local_parameterizations_;
   }
+#else
+  /**
+   * @brief Read-only access to the variable local parameterizations
```
efernandez, line 134: Pendantic, but this comment should be updated:
```suggestion
   * @brief Read-only access to the variable manifolds
```

fuse_constraints/include/fuse_constraints/marginal_cost_function.h
```diff
@@ -77,6 +77,21 @@ class MarginalCostFunction : public ceres::CostFunction
...
     const std::vector<fuse_core::VectorXd>& x_bar,
     const std::vector<fuse_core::LocalParameterization::SharedPtr>& local_parameterizations);
+#else
+  /**
+   * @brief Construct a cost function instance
+   *
+   * @param[in] A                       The A matrix of the marginal cost (of the form A*(x - x_bar) + b)
+   * @param[in] b                       The b vector of the marginal cost (of the form A*(x - x_bar) + b)
+   * @param[in] x_bar                   The linearization point of the involved variables
+   * @param[in] manifolds               The manifold associated with the variable
```
efernandez, line 87:
```suggestion
   * @param[in] A         The A matrix of the marginal cost (of the form A*(x - x_bar) + b)
   * @param[in] b         The b vector of the marginal cost (of the form A*(x - x_bar) + b)
   * @param[in] x_bar     The linearization point of the involved variables
   * @param[in] manifolds The manifold associated with the variable
```

efernandez (commented):
LGTM I've done a quick review and only made a few minor comments

svwilliams:
Reviews for the final set of changes is here: #363
Merging this PR and rebasing #363

PR #348, opened 2023-11-17
Fix tests

paulbovbel:

PR #349, opened 2023-11-24
Vision constraints

oscarmendezm:
I added the variables necessary to represent a pinhole camera, along with constraints for fixed landmarks to optimize over the cameras 6-dof position using reprojection error, and optionally optimize for intrinsics. Use cases are represented in the unit tests. We can formalize a marker as a 6-dof pose and turn it into an arbitrary set of points in the marker coordinate frame.

### DIFF
--- a/fuse_constraints/CMakeLists.txt
+++ b/fuse_constraints/CMakeLists.txt
@@ -44,6 +44,7 @@ add_library(${PROJECT_NAME}
   src/absolute_orientation_3d_stamped_euler_constraint.cpp
   src/absolute_pose_2d_stamped_constraint.cpp
   src/absolute_pose_3d_stamped_constraint.cpp
+  src/fixed_3d_landmark_constraint.cpp
   src/marginal_constraint.cpp
   src/marginal_cost_function.cpp
   src/marginalize_variables.cpp
@@ -236,6 +237,30 @@ if(CATKIN_ENABLE_TESTING)
     ${CERES_LIBRARIES}
   )
   set_target_properties(test_absolute_pose_2d_stamped_constraint
+    PROPERTIES
+      CXX_STANDARD 17
+      CXX_STANDARD_REQUIRED YES
+  )
+
+  # Absolute Pose 3D Stamped Constraint Tests
+  catkin_add_gtest(test_fixed_3d_landmark_constraint
+    test/test_fixed_3d_landmark_constraint.cpp
+  )
+  add_dependencies(test_fixed_3d_landmark_constraint
+    ${catkin_EXPORTED_TARGETS}
+  )
+  target_include_directories(test_fixed_3d_landmark_constraint
+    PRIVATE
+      include
+      ${catkin_INCLUDE_DIRS}
+      ${CERES_INCLUDE_DIRS}
+  )
+  target_link_libraries(test_fixed_3d_landmark_constraint
+    ${PROJECT_NAME}
+    ${catkin_LIBRARIES}
+    ${CERES_LIBRARIES}
+  )
+  set_target_properties(test_fixed_3d_landmark_constraint
     PROPERTIES
       CXX_STANDARD 17
       CXX_STANDARD_REQUIRED YES
--- a/fuse_constraints/CMakeLists.txt
+++ b/fuse_constraints/CMakeLists.txt
@@ -242,7 +242,7 @@ if(CATKIN_ENABLE_TESTING)
       CXX_STANDARD_REQUIRED YES
   )
 
-  # Absolute Pose 3D Stamped Constraint Tests
+  # Fixed 3D Landmark Constraint Tests
   catkin_add_gtest(test_fixed_3d_landmark_constraint
     test/test_fixed_3d_landmark_constraint.cpp
   )
--- a/fuse_constraints/include/fuse_constraints/fixed_3d_landmark_constraint.h
+++ b/fuse_constraints/include/fuse_constraints/fixed_3d_landmark_constraint.h
@@ -1,0 +1,206 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Author: Oscar Mendez
+ *  Created on Mon Nov 12 2023
+ *
+ *  Copyright (c) 2023, Locus Robotics
+ *  All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without
+ *  modification, are permitted provided that the following conditions
+ *  are met:
+ *
+ *   * Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above
+ *     copyright notice, this list of conditions and the following
+ *     disclaimer in the documentation and/or other materials provided
+ *     with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its
+ *     contributors may be used to endorse or promote products derived
+ *     from this software without specific prior written permission.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ *  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ *  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_CONSTRAINTS_FIXED_3D_LANDMARK_CONSTRAINT_H
+#define FUSE_CONSTRAINTS_FIXED_3D_LANDMARK_CONSTRAINT_H
+
+#include <fuse_core/constraint.h>
+#include <fuse_core/eigen.h>
+#include <fuse_core/fuse_macros.h>
+#include <fuse_core/serialization.h>
+#include <fuse_core/uuid.h>
+#include <fuse_variables/orientation_3d_stamped.h>
+#include <fuse_variables/position_3d_stamped.h>
+#include <fuse_variables/pinhole_camera.h>
+
+#include <boost/serialization/access.hpp>
+#include <boost/serialization/base_object.hpp>
+#include <boost/serialization/export.hpp>
+#include <Eigen/Dense>
+
+#include <ostream>
+#include <string>
+#include <vector>
+
+namespace fuse_constraints
+{
+
+/**
+ * @brief A constraint that represents either prior information about a 3D pose, or a direct measurement of the 3D pose.
+ *
+ * A 3D pose is the combination of a 3D position and a 3D orientation variable. As a convenience, this class applies
+ * an absolute constraint on both variables at once. This type of constraint arises in many situations. In mapping
+ * it is common to define the very first pose as the origin. Some sensors, such as GPS, provide direct measurements
+ * of the robot's pose in the global frame. And localization systems often match laserscans or pointclouds to a prior
+ * map (scan-to-map measurements). This constraint holds the measured 3D pose and the measurement
+ * uncertainty/covariance. Orientations are represented as quaternions.
+ */
+class Fixed3DLandmarkConstraint : public fuse_core::Constraint
+{
+public:
+  FUSE_CONSTRAINT_DEFINITIONS_WITH_EIGEN(Fixed3DLandmarkConstraint);
+
+  /**
+   * @brief Default constructor
+   */
+  Fixed3DLandmarkConstraint() = default;
+
+  /**
+   * @brief Create a constraint using a measurement/prior of the 3D pose
+   *
+   * @param[in] source        The name of the sensor or motion model that generated this constraint
+   * @param[in] position      The variable representing the position components of the marker pose
+   * @param[in] orientation   The variable representing the orientation components of the marker pose
+   * @param[in] calibraton    The intrinsic calibration parameters of the camera (4x1 vector: cx, cy, fx, fy). 
+   *                          NOTE: Best practice is to fix this variable unless we have several observations 
+   *                          with the same camera
+   * @param[in] marker_size          The marker_size of the marker, in meters. Assumed to be square.
+   * @param[in] observations  The 2D (pixel) observations of each marker's corners.
+   * @param[in] mean          The measured/prior pose of the markeras a vector (7x1 vector: x, y, z, qw, qx, qy, qz)
+   * @param[in] covariance    The measurement/prior marker pose covariance (6x6 matrix: x, y, z, qx, qy, qz)
+   */
+  Fixed3DLandmarkConstraint(const std::string& source, const fuse_variables::Position3DStamped& position,
+                            const fuse_variables::Orientation3DStamped& orientation,
+                            const fuse_variables::PinholeCamera& calibraton,
+                            const fuse_core::Vector1d& marker_size, 
+                            const fuse_core::MatrixXd& observations,
+                            const fuse_core::Vector7d& mean, const fuse_core::Matrix6d& covariance);
+
+  /**
+   * @brief Destructor
+   */
+  virtual ~Fixed3DLandmarkConstraint() = default;
+
+  /**
+   * @brief Read-only access to the 1D vector of marker marker_size.
+   *
+   * Order is (s)
+   */
+  const fuse_core::Vector1d& marker_size() const
+  {
+    return marker_size_;
+  }
+
+  /**
+   * @brief Read-only access to the measured/prior vector of mean values.
+   *
+   * Order is (x, y, z, qw, qx, qy, qz)
+   */
+  const fuse_core::Vector7d& mean() const
+  {
+    return mean_;
+  }
+
+  /**
+   * @brief Read-only access to the square root information matrix.
+   *
+   * Order is (x, y, z, qx, qy, qz)
+   */
+  const fuse_core::Matrix6d& sqrtInformation() const
+  {
+    return sqrt_information_;
+  }
+
+  /**
+   * @brief Compute the measurement covariance matrix.
+   *
+   * Order is (x, y, z, qx, qy, qz)
+   */
+  fuse_core::Matrix6d covariance() const
+  {
+    return (sqrt_information_.transpose() * sqrt_information_).inverse();
+  }
+
+  /**
+   * @brief Read-only access to the observation Matrix (Nx2).
+   *
+   * Order is (x, y)
+   */
+  const fuse_core::MatrixXd& observations() const
+  {
+    return observations_;
+  }
+
+  /**
+   * @brief Print a human-readable description of the constraint to the provided stream.
+   *
+   * @param[out] stream The stream to write to. Defaults to stdout.
+   */
+  void print(std::ostream& stream = std::cout) const override;
+
+  /**
+   * @brief Construct an instance of this constraint's cost function
+   *
+   * The function caller will own the new cost function instance. It is the responsibility of the caller to delete
+   * the cost function object when it is no longer needed. If the pointer is provided to a Ceres::Problem object, the
+   * Ceres::Problem object will takes ownership of the pointer and delete it during destruction.
+   *
+   * @return A base pointer to an instance of a derived CostFunction.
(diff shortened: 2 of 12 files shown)


oscarmendezm:
The errors shown here are fixed by #348

ayrton04:
@oscarmendezm is this ready for review? Do you want to tag any reviewers?

fuse_constraints/CMakeLists.txt
```diff
@@ -241,6 +242,30 @@ if(CATKIN_ENABLE_TESTING)
       CXX_STANDARD_REQUIRED YES
   )
 
+  # Absolute Pose 3D Stamped Constraint Tests
```
ayrton04: Copy/pasted comment.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_constraint.h
```diff
@@ -0,0 +1,229 @@
...
+
+namespace fuse_constraints
+{
+
+/**
+ * @brief A constraint that represents an observation of a 3D landmark (ARTag or Similar)
+ *
+ * A landmark is represented as a 3D pose (3D position and a 3D orientation). This class takes 
+ * the ground truth location of the 3D landmark and applies a reprojection-error based constraint
+ * an constraint on The position, orientation and calibration of the camera that observed the landmark.
```
ayrton04: Lower-case T in `the`

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_constraint.h
```diff
@@ -0,0 +1,229 @@
...
+  /**
+   * @brief Default constructor
+   */
+  Fixed3DLandmarkConstraint() = default;
+
+    /**
+   * @brief Create a constraint using a known 3D fiducial marker
+   *
+   * @param[in] source        The name of the sensor or motion model that generated this constraint
+   * @param[in] position      The variable representing the position components of the marker pose
```
ayrton04: Maybe "...of the landmark pose"?

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_constraint.h
```diff
@@ -0,0 +1,229 @@
...
+   * @param[in] orientation   The variable representing the orientation components of the marker pose
+   * @param[in] calibraton    The intrinsic calibration parameters of the camera (4x1 vector: cx, cy, fx, fy).
+   *                          NOTE: Best practice is to fix this variable unless we have several observations
+   *                          with the same camera
+   * @param[in] pts3d         Matrix of 3D points in marker coordiate frame.
+   * @param[in] observations  The 2D (pixel) observations of each marker's corners.
+   * @param[in] mean          The measured/prior pose of the markeras a vector (7x1 vector: x, y, z, qw, qx, qy, qz)
```
ayrton04: *marker as (space). Same on line 115.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_constraint.h
```diff
@@ -0,0 +1,229 @@
...
+   */
+  Fixed3DLandmarkConstraint(const std::string& source, const fuse_variables::Position3DStamped& position,
+                            const fuse_variables::Orientation3DStamped& orientation,
+                            const fuse_variables::PinholeCamera& calibraton,
+                            const fuse_core::MatrixXd& pts3d,
+                            const fuse_core::MatrixXd& observations, const fuse_core::Vector7d& mean,
+                            const fuse_core::Matrix6d& covariance);
+
+  /**
+   * @brief Create a constraint using a known 3D fiducial marker. Convinience constructor for the special case of
```
ayrton04: *Convenience

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_constraint.h
```diff
@@ -0,0 +1,229 @@
...
+   * The function caller will own the new cost function instance. It is the responsibility of the caller to delete
+   * the cost function object when it is no longer needed. If the pointer is provided to a Ceres::Problem object, the
+   * Ceres::Problem object will takes ownership of the pointer and delete it during destruction.
+   *
+   * @return A base pointer to an instance of a derived CostFunction.
+   */
+  ceres::CostFunction* costFunction() const override;
+
+protected:
+  fuse_core::MatrixXd pts3d_;             //!< THe 3D points in marker Coordinate frame
```
ayrton04: *The (lower-case H)

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_constraint.h
```diff
@@ -0,0 +1,229 @@
...
+  fuse_core::MatrixXd observations_;      //!< The 2D observations of the marker at postion mean_
```
ayrton04: Maybe add "in pixel space" here?

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
```diff
@@ -0,0 +1,206 @@
...
+ * applies a prior constraint on the 3D position, orientation and calibration variables at once.
+ *
+ * The cost function is of the form:
+ *
+ *   cost(x) = || A * (K * [R_q | p] * [R_{b(3:6)} | b(0:2))] * X) - x  ||
+ *
+ * where, the matrix A and the vector b are fixed, p is the camera position variable, and q is the camera orientation
+ * variable, K is the calibration matrix created from the calibration variable, X is the set of marker 3D points,
+ * R_b(0:3) is the Rotation matrix from the fixed landmark orentation (b(3:6)), b(0:2) is the fixed landmark position
+ * and x is the 2D ovservations.
```
ayrton04, line 63: *observations

ayrton04: Would the residual weighting matrix not apply to the entire difference between the projected points and the observations? In other words, don't we need brackets enclosing the difference?

```
|| A * (K * [R_q | p] * [R_{b(3:6)} | b(0:2))] * X - x) ||
```

Also, I think other cost functions square the costs. Is that not needed here?

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
```diff
@@ -0,0 +1,206 @@
...
+
+Fixed3DLandmarkCostFunctor::Fixed3DLandmarkCostFunctor(const fuse_core::MatrixXd& A, const fuse_core::Vector7d& b,
+                                                       const fuse_core::MatrixXd& obs, const fuse_core::MatrixXd& pts3d)
+  : A_(A)
+  , b_(b)
+  , obs_(obs)
+  , pts3d_(pts3d)
+  , orientation_functor_(fuse_core::Matrix3d::Identity(), b_.tail<4>())  // Delta will not be scaled
+{
+  pts3d_.transposeInPlace();  // Assume we get Nx3, so transform to 3xN
```
ayrton04: Should we have some kind of validation check for these? Maybe that can be handled at the constraint level.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
```diff
@@ -0,0 +1,206 @@
...
+      rm[3], rm[4], rm[5], b_(1),    // NOLINT
+      rm[6], rm[7], rm[8], b_(2),    // NOLINT
+      0.0, 0.0, 0.0, 1.0;            // NOLINT
+
+  // TODO(omendez): Can we do this directly on quaternions?
+  // ceres::QuaternionProduct(q, position, difference);
+  // ceres::QuaternionRotatePoint()
+
+  // Transform points to marker pose and store for later use.
+  pts3d_ = Tm * pts3d_.colwise().homogeneous();
```
ayrton04, line 147: If this is transforming the points into the marker frame, then what frame were the points in the `pts3d` parameter given in? The [documentation](https://github.com/locusrobotics/fuse/pull/349/files#diff-f25bb0cf963397a795c16bbc24207bef704767c370a91003457417a4b78020afR101) says they are already in the marker frame.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
```diff
@@ -0,0 +1,206 @@
...
+  // Create Marker Transform Matrix (Tm)
+  T qm[4] = { T(b_(3)), T(b_(4)), T(b_(5)), T(b_(6)) };
+  T rm[9];
+  ceres::QuaternionToRotation(qm, rm);
+
+  Eigen::Matrix<T, 4, 4, Eigen::RowMajor> Tm;
+  Tm << rm[0], rm[1], rm[2], T(b_(0)),  // NOLINT
+      rm[3], rm[4], rm[5], T(b_(1)),    // NOLINT
+      rm[6], rm[7], rm[8], T(b_(2)),    // NOLINT
+      T(0.0), T(0.0), T(0.0), T(1.0);   // NOLINT
```
ayrton04: This block appears to be duplicating the operations in the [constructor](https://github.com/locusrobotics/fuse/pull/349/files#diff-f25bb0cf963397a795c16bbc24207bef704767c370a91003457417a4b78020afR131-R140). Should we make `Tm` a member variable and store that matrix at the time of construction?

Also, I don't see `Tm` even being used here.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
```diff
@@ -0,0 +1,206 @@
...
+
+  auto d = (obs_.cast<T>() - xp.block(0, 0, xp.rows(), 2));
+
+  for (uint i = 0; i < 4; i++)
+  {
+    residual[i * 2] = d.row(i)[0];
+    residual[i * 2 + 1] = d.row(i)[1];
+  }
+
+  // TODO(omendez): how do we apply covariance weigthing to point losses from a pose uncertainty?
```
ayrton04, line 188: Right, I see `A_` is not being used here.

fuse_variables/include/fuse_variables/pinhole_camera.h
```diff
@@ -0,0 +1,184 @@
...
+namespace fuse_variables
+{
+/**
+ * @brief Variable representing intrinsic parameters of a camera.
+ *
+ * The UUID of this class is constant after
+ * construction and dependent on a user input database id. As such, the database id cannot be altered after
+ * construction.
+ */
+class PinholeCamera : public FixedSizeVariable<4>
```
ayrton04, line 60: NBD, but in the future, could probably PR this independently of the landmark constraint.

ayrton04 (commented):
Mostly small nits with comments that you are free to ignore. The only thing I am not sure about code-wise is the seemingly superfluous `Tm` construction in `Fixed3DLandmarkCostFunctor::operator()`.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
ayrton04: Ignore the squaring comment, apologies.

oscarmendezm: I think you are right, I'll change it, but I need to work out how this will work. 

The weighting matrix should apply to all the residuals, and would weight each reprojection error individually, but I suspect we won't get something like that directly at runtime. Instead, I assume our weights will come from the landmark pose covariance, rather than the observations?

So basically, i agree, but I need to work out how to calculate A from the covariance matrix. I've been looking at [this document](https://arxiv.org/pdf/2103.15980.pdf) for the 3D constraints, and on p.59 there are some starting points for the jacobians, but I need to sit down and think it through properly.

fuse_constraints/CMakeLists.txt
oscarmendezm: Fixed.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
oscarmendezm: Sorry, this is confusing wording on my part. The 3D points are in marker coordinate frame, so something like: 

(-1,-1)
(-1, 1)
( 1,-1)
( 1, 1)

(assuming the marker is 2x2m). We then multiply on the left by the marker pose (in map coordnate frame) to get the points in map coordinate frame.

I'll fix the wording on the comment.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
oscarmendezm: It doesn't get used here, just some left over code. I'll remove it. Probably not worth creating a member variable, as we don't use it after transforming the points to map coordinate frame.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
oscarmendezm: Fixed this.

ayrton04 (approved):
Whatever check is being done with GH actions seems unhappy for unrelated reasons, but LGTM.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_constraint.h
```diff
@@ -0,0 +1,229 @@
...
+
+namespace fuse_constraints
+{
+
+/**
+ * @brief A constraint that represents an observation of a 3D landmark (ARTag or Similar)
+ *
+ * A landmark is represented as a 3D pose (3D position and a 3D orientation). This class takes 
+ * the ground truth location of the 3D landmark and applies a reprojection-error based constraint
+ * an constraint on the position, orientation and calibration of the camera that observed the landmark.
```
svwilliams, line 66: the ground truth location of the 3D landmark and applies a reprojection-error based constraint
 ~an constraint~ on the position, orientation and calibration of the camera that observed the landmark.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_constraint.h
```diff
@@ -0,0 +1,229 @@
...
+ * 
+ * In most cases, the camera calibration should be held fixed as a single landmark does not present enough
+ * points to accurate constrain the pose AND the calibraton.
```
svwilliams, line 69: points to accurate**ly** constrain the pose AND the calibraton.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
```diff
@@ -0,0 +1,195 @@
...
+class Fixed3DLandmarkCostFunctor
+{
+public:
+  FUSE_MAKE_ALIGNED_OPERATOR_NEW();
+
+  /**
+   * @brief Construct a cost function instance
+   *
+   * @param[in] A The residual weighting matrix, most likely derived from the square root information
+   *              matrix in order (x, y, z, qx, qy, qz)
```
svwilliams, line 82: Typically I would expect A to be a 2x2 matrix that models the measurement error in the image space. Something like half a pixel. And that A matrix weighting would be applied to each corner residual. The robot pose uncertainties and landmark pose uncertainties would get jointly estimated as part of the graph optimization, since they would all be graph variables. Assuming that sounds correct, the matrix order should really be (u, v) or (x, y).

---

However, this is a "fixed" landmark constraint. The landmark pose is not actually a graph variable, and so the uncertainty of the landmark pose will not be handled automatically. You have mentioned wanting to derive the A matrix from some notion of a fixed landmark uncertainty. Other than saying that sounds possible, it's not something I have done before. My first thought is to compute the Jacobian of the cost function with respect to the landmark variables, and then apply that jacobian to the landmark covariance: J * Cov *J' which should result in a 2x2 matrix (J is 2x6, Cov is 6x6). With all of the standard caveats about linear approximations of nonlinear functions.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
```diff
@@ -0,0 +1,195 @@
...
+  template <typename T>
+  bool operator()(const T* const position, const T* const orientation, const T* const calibration, T* residual) const;
+
+private:
+  fuse_core::MatrixXd A_;
+  fuse_core::Vector7d b_;
+  fuse_core::MatrixXd obs_;
+  fuse_core::MatrixXd pts3d_;
+
+  NormalPriorOrientation3DCostFunctor orientation_functor_;
```
svwilliams, line 118: I see the orientation functor getting constructed, but I'm not seeing how it is used.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
```diff
@@ -0,0 +1,195 @@
...
+  xp = (K * Ta * pts3d_.cast<T>());
+  xp.transposeInPlace();
+  xp = xp.array().colwise() / xp.col(2).array();
+
+  auto d = (obs_.cast<T>() - xp.block(0, 0, xp.rows(), 2));
+
+  for (uint i = 0; i < 4; i++)
+  {
+    residual[i * 2] = d.row(i)[0];
+    residual[i * 2 + 1] = d.row(i)[1];
```
svwilliams, line 185: I would think the A matrix would be applied here
```
[ r(2*i + 0) ]  =  A * [ d(i).x ]
[ r(2*i + 1) ]         [ d(i).y ]
```

fuse_constraints/test/test_fixed_3d_landmark_constraint.cpp
svwilliams, line 1: I'm guessing you have some sort of Python/Octave script you used to generate all of the observation values. Is that something you would feel comfortable checking in as well? If we ever need to add/modify/debug test cases, it could be useful to have around.

fuse_variables/include/fuse_variables/pinhole_camera.h
```diff
@@ -0,0 +1,184 @@
...
+   * @brief Default constructor
+   */
+  PinholeCamera() = default;
+
+  /**
+   * @brief Construct a pinhole camera variable given a camera id
+   *
+   * @param[in] camera_id  The id associated to a camera
+   */
+  explicit PinholeCamera(const uint64_t& camera_id);
```
svwilliams, line 86: For positions/orientations, I use a UUID as a mechanism for differentiating between poses from different robots. My reasoning goes like this: Most robots have some sort of name string, and a UUID can be easily generated directly from the name. That makes distributed systems easier. Each robot can generate an ID number for itself without having to coordinate with some central system. And if the user really just wants an index, you can cram an integer inside a UUID container _without too much hassle_.

If I imagine a multi-robot system where each robot has multiple cameras, how do I coordinate assigning each camera a unique camera index correctly? Whereas I could generate a UUID from the robot name + camera name strings easily. I might even be able to pass in the camera image header frame name directly.

For landmarks, I'll be honest. I didn't really have a plan. Landmarks don't seem tied to a specific piece of hardware in the same way. If the landmarks are all fixed, then they probably come from a central database anyway, and assigning ID numbers is probably easy. If the landmarks come from some sort of ARTag, then they have an ID built into them. But if you are trying to build up a database of image feature landmarks from multiple robots, the ID field may get cumbersome.

---

Unrelated to ID numbers. A robot general has multiple sensors. Those sensors are mounted at some fixed location on the robot. Extrinsic calibration if you will. If I have a variable that represents the pose of the robot in the global frame, and that robot has a camera mounted with some non-Identity transform from the main robot frame, how do we include observations from the camera in the same graph as wheel encoder data in the robot frame? Would we need a "MountedCamera" variable that includes the base->camera transform? Or would that really get handled in the constraint equation? This is a topic on the ToDo list I keep in my head. I just haven't needed to tackle it yet.

---

And so ends this edition of "Steve's Random Thoughts".

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
oscarmendezm: This is something I'm still trying to understand. 

It makes sense to me that we would weight each residual independently, and that seems to be what [Ceres BA does](http://ceres-solver.org/nnls_tutorial.html#bundle-adjustment). There may be a case here for abstracting away some of the behaviour here into N "reprojection_error" constraints, which would let auto-diff work out the details. I started implementing it this way to make use of the efficient matrix multiplication for larger number of points. 

In this implementation, my thoughts here are that the full expression is:

cost(x) = || A(K * Tc * Tm * X - x) ||

where Tc is the camera pose, and Tm is the marker pose, so A is a Nx2 matrix, which encodes the uncertainty of each projected point, but it should not be equal for all points. In a very convoluted example, if most of the uncertainty is on the rotation of the camera along the viewing direction, we'd expect a point that projects in the middle of the image to have less uncertainty than one at the edges. 

Not sure what the right thing to do here is.

I agree regarding fixed landmark constraint, thanks for clarifying that.

fuse_constraints/include/fuse_constraints/fixed_3d_landmark_cost_functor.h
oscarmendezm: Probably left over from old code, i'll remove it.

fuse_variables/include/fuse_variables/pinhole_camera.h
oscarmendezm: The UUID was actually something I had doubts about. Seems like a sensible solution to a distributed problem now that you've explained it. Using the robot name + camera frame_id makes sense to me, as it would create a unique UUID for each camera that is somewhat guaranteed to not repeat across a swarm. 

For landmarks, I think a UUID per ARTag makes sense, as does a UUID per 3D point (e.g. VSLAM landmark) but that would have to be unique across multiple robots, so some merging might be required in a distributed system. I don't think we want one per 2D observation of each point though, as those are your "measurements", and we'd probably never want to optimize those? 

---
Multiple sensors: I'd be in favour of a "Mounted Camera" variable, one that is shared across multiple constraints, so each "landmark" constraint goes via a shared mounted camera variable that can get optimized for extrinsics, or maybe even fixed?

PR #350, opened 2023-11-27
Fixing CI tests

oscarmendezm:
Removed the unused tolerance parameter from ExpectCostFunctionsAreEqual, and replaced EXPECT_NEAR with EXPECT_DOUBLE_EQ to prevent jenkins error.


oscarmendezm:
Already done in #348

PR #354, opened 2024-01-16
3D fuse models

giafranchini:
We implemented the following sensor models:
- odometry_3d
- imu_3d

And the following motion models:
- omnidirectional_3d
- omnidirectional_3d_ignition

With the associated sensor processing, constraints and tests


fuse_optimizers/include/fuse_optimizers/batch_optimizer_component.hpp
svwilliams, line 1: These changes to the fixed lag smoother and batch optimizer are unrelated to this PR's stated goal of adding "3D fuse models". Can you split the changes to the optimizers into a separate PR?

fuse_core/include/fuse_core/util.hpp
```diff
@@ -150,6 +151,178 @@ Eigen::Matrix<T, 2, 2, Eigen::RowMajor> rotationMatrix2D(const T angle)
...
+
+/**
+ * @brief Compute product of two quaternions and the function jacobian
+ * TODO(giafranchini): parametric jacobian computation? Atm this function is only used in 
+ * normal_prior_pose_3d cost function. There we only need the derivatives wrt quaternion W,
+ * so at the time we are only computing the jacobian wrt W
+ * 
+ * @param[in] z Pointer to the first quaternion array (4x1)
+ * @param[in] w Pointer to the second quaternion array (4x1)
+ * @param[in] z Pointer to the first quaternion array (4x1)
```
svwilliams:
```suggestion
 * @param[in] zw Pointer to the output quaternion array (4x1) that will be populated with the result of z * w
```
Or whatever description you want

fuse_core/include/fuse_core/util.hpp
```diff
@@ -150,6 +151,178 @@ Eigen::Matrix<T, 2, 2, Eigen::RowMajor> rotationMatrix2D(const T angle)
...
+/**
+ * @brief Compute product of two quaternions and the function jacobian
+ * TODO(giafranchini): parametric jacobian computation? Atm this function is only used in 
+ * normal_prior_pose_3d cost function. There we only need the derivatives wrt quaternion W,
+ * so at the time we are only computing the jacobian wrt W
+ * 
+ * @param[in] z Pointer to the first quaternion array (4x1)
+ * @param[in] w Pointer to the second quaternion array (4x1)
+ * @param[in] z Pointer to the first quaternion array (4x1) 
+ * @param[in] jacobian Pointer to the jacobian matrix (4x4, optional)
```
svwilliams:
```suggestion
 * @param[in] jacobian Pointer to the jacobian of zw with respect to w (4x4, optional)
```

fuse_core/test/test_util.cpp
```diff
@@ -81,3 +81,29 @@ TEST(Util, wrapAngle2D)
...
+  q[1] = 0.0640713;
+  q[2] = 0.0911575;
+  q[3] = -0.1534393;
+
+  fuse_core::quaternion2rpy(q, rpy);
+  EXPECT_NEAR(0.1, rpy[0], 1e-6);
+  EXPECT_NEAR(0.2, rpy[1], 1e-6);
+  EXPECT_NEAR(-0.3, rpy[2], 1e-6);
+
+  //TODO(giafranchini): Add test for jacobian
```
svwilliams: I'd like to see a test comparing your Jacobian calculations to the auto-diff Jacobians computed using the `ceres::QuaternionProduct()` function.

fuse_constraints/src/normal_prior_pose_3d.cpp
```diff
@@ -0,0 +1,104 @@
...
+  double j_product[16];
+  double j_quat2angle[12];
+
+  // Compute position residuals
+  residuals[0] = parameters[0][0] - b_[0];
+  residuals[1] = parameters[0][1] - b_[1];
+  residuals[2] = parameters[0][2] - b_[2];
+  // Compute orientation residuals 
+  fuse_core::quaternionProduct(observation_inverse, variable, difference, j_product);
+  fuse_core::quaternionToAngleAxis(difference, &residuals[3], j_quat2angle); // orientation angle-axis
```
svwilliams, line 84: Populating `j_product` and `j_quat2angle` could be conditional on `(jacobians[1] != nullptr)`. Can you at least add a TODO note for that?

fuse_constraints/src/absolute_pose_3d_stamped_constraint.cpp
```diff
@@ -74,8 +74,17 @@ void AbsolutePose3DStampedConstraint::print(std::ostream & stream) const
...
+  
+  // Here we return a cost function that computes the analytic derivatives/jacobians, but we could
+  // use automatic differentiation as follows:
+
+  // return new ceres::AutoDiffCostFunction<NormalPriorPose3DCostFunctor, 6, 3, 4>(
+  //   new NormalPriorPose3DCostFunctor(sqrt_information_, mean_);
+    
+  // And including the followings:
+  // #include <ceres/autodiff_cost_function.h>
+  // #include <fuse_constraints/normal_prior_pose_3d_cost_functor.hpp>
```
svwilliams: These comments can be removed. If the analytic derivatives are correct, there is no reason to switch back to autodiff.

fuse_constraints/src/normal_prior_orientation_3d.cpp
```diff
@@ -31,17 +31,61 @@
...
+    -b_(2),
+    -b_(3)
+  };
+
+  double difference[4];
+  double j_product[16];
+  double j_quat2angle[12];
+
+  fuse_core::quaternionProduct(observation_inverse, variable, difference, j_product);
+  fuse_core::quaternionToAngleAxis(difference, residuals, j_quat2angle); // orientation angle-axis
```
svwilliams, line 78: Same note as the 3D pose prior: Populating the jacobians could be conditioned on the `(jacobians != nullptr)`.
Also, this implementation is basically identical to the one in the pose prior. It would be nice to reuse that implementation in both places.

Can you add TODO notes for both of those things?

fuse_constraints/src/normal_prior_orientation_3d.cpp
```diff
@@ -31,17 +31,61 @@
...
+ 
+  // Scale the residuals by the square root information matrix to account for the measurement
+  // uncertainty.
+  Eigen::Map<Eigen::Vector3d> residuals_map(residuals);
+  residuals_map.applyOnTheLeft(A_);
+
+  if (jacobians != nullptr) {
```
svwilliams, line 85: You still need to check that `jacobians[0]` is not null. I don't know why it ever would be, but it is theoretically possible to populate the `jacobians` array with N nullptrs.

fuse_constraints/test/cost_function_gtest.hpp
```diff
@@ -70,8 +71,22 @@ static void ExpectCostFunctionsAreEqual(
   }
 
   std::unique_ptr<double[]> parameters(new double[num_parameters]);
-  for (size_t i = 0; i < num_parameters; ++i) {
-    parameters[i] = static_cast<double>(i) + 1.0;
+  if ((num_parameters == 7) && (parameter_block_sizes[0] == 3) &&
+    (parameter_block_sizes[1] == 4))
```
svwilliams, line 74: This is making a set of assumptions that I am not comfortable with. There is no reason to believe every cost function that involves two variables of size (3, 4) implies that the second variable is a quaternion. And symmetrically, there is no reason to believe the only time a cost function involves a quaternion will be when there are two variables of size (3, 4).

Maybe the `ExpectCostFunctionsAreEqual()` function needs an optional array of quaternion parameter block indices? Or an optional array of parameter initializations? Then you could handle any type of variable with special initialization requirements.

fuse_core/include/fuse_core/util.hpp
```diff
@@ -150,6 +151,178 @@ Eigen::Matrix<T, 2, 2, Eigen::RowMajor> rotationMatrix2D(const T angle)
...
+    const double & q0 = q[0];
+    const double & q1 = q[1];
+    const double & q2 = q[2];
+    const double & q3 = q[3];
+    const double q_sum2 = q0 * q0 + q1 * q1 + q2 * q2 + q3 * q3;
+    const double sin_theta2 = q1 * q1 + q2 * q2 + q3 * q3;
+    const double sin_theta = std::sqrt(sin_theta2);
+    const double cos_theta = q0;
+
+    if (std::fpclassify(sin_theta) != FP_ZERO) {
```
svwilliams: What happens in the FP_SUBNORMAL case? Are the jacobians still computed correctly?

fuse_core/include/fuse_core/util.hpp
```diff
@@ -150,6 +151,178 @@ Eigen::Matrix<T, 2, 2, Eigen::RowMajor> rotationMatrix2D(const T angle)
...
+  if (jacobian) {
+    Eigen::Map<Eigen::Matrix<double, 3, 4, Eigen::RowMajor>> jacobian_map(jacobian);
+    const double qw = q[0];
+    const double qx = q[1];
+    const double qy = q[2];
+    const double qz = q[3];
+    const double discr = qw * qy - qx * qz; 
+    jacobian_map.setZero();
+
+    if (discr > 0.49999) {
```
svwilliams: In the `quaternionToAngleAxis()` function, you check for the FP_ZERO condition. Is there something similar that can be done here to formalize "really close to 0.5"? For example, would this work?
```
if (discr >= 0.5 - 2.0 * numeric_limits<double>::epsilon())
```

fuse_constraints/test/test_normal_prior_pose_3d.cpp
```diff
@@ -0,0 +1,92 @@
...
+  const auto num_residuals = full_sqrt_information.rows();
+
+  AutoDiffNormalPriorPose3D autodiff_cost_function(
+    new fuse_constraints::NormalPriorPose3DCostFunctor(full_sqrt_information, full_mean),
+    num_residuals);
+  
+  // Compare the expected, automatic differentiation, cost function and the actual one
+  // N.B. in ExpectCostFunctionsAreEqual constructor, the first argument is the expected cost function
+  // and the second argument is the actual cost function 
+  ExpectCostFunctionsAreEqual(cost_function, autodiff_cost_function, 1e-12);
```
svwilliams: The `ExpectCostFunctionsAreEqual()` function checks the function output and derivatives at one specific linearization point. However, there is at least one special case around zero that would be nice to verify as well.

ayrton04:
Thanks for the PR! Just a note that Locus folks _are_ looking at this, but it's a big PR and people are pretty swamped, so it could be a bit longer before the reviews are finished.

giafranchini:
> Thanks for the PR! Just a note that Locus folks _are_ looking at this, but it's a big PR and people are pretty swamped, so it could be a bit longer before the reviews are finished.

No problem, I can imagine that! There is no rush. Feel free to reach me if you need anything.

svwilliams (commented):
As Tom said, we are in the process of reviewing this PR. Sorry it is taking so long to complete. But as you well know, this is an immense amount of very technical work. So thank you for this submission, and for your patience as we work through the code.

But to get things moving forward a little bit, here's a partial review of just the absolute pose and orientation constraints and cost functions.

My big asks so far are:
* Can you split the changes to the optimizers into a separate PR? Changing the optimizers into "components" is unrelated to adding 3D sensor models.
* There is a lot of Jacobian math in this PR that is difficult to review with any certainty. I would like to see unit tests that compare the analytic jacobians provided in this PR with the auto-diff versions generated by Ceres to prove correctness. These should also test over a range of initial conditions and/or special cases of interest.

giafranchini:
@svwilliams ok, I'll start working on these changes.

fuse_optimizers/include/fuse_optimizers/batch_optimizer_component.hpp
giafranchini: I opened a new PR https://github.com/locusrobotics/fuse/pull/358

fuse_core/test/test_util.cpp
giafranchini: Done in https://github.com/locusrobotics/fuse/pull/354/commits/ee059b622832c42924929d89443f1df25fc0f80a

fuse_constraints/test/cost_function_gtest.hpp
giafranchini: Yes I agree with you. This was just a workaround to be able to test my cost functions, but it's not general at all. Can you please explain a little bit more your idea?

fuse_core/include/fuse_core/util.hpp
giafranchini: I followed the approach that I found for the implementation of this function in ceres:
 https://github.com/ceres-solver/ceres-solver/blob/62c03d6ff3b1735d4b0a88006486cfde39f10063/include/ceres/rotation.h#L355

If `sin_theta` is FP_SUBNORMAL then it's because also `sin_theta2` it's of the same form, meaning that some quaternion components have strange values. 
The FP_ZERO case will of course cause trouble since it's at denominator, and this is handled computing a first order taylor approximation.
The other cases will also cause problems in the outputs, but then either the conversion and the jacobians doesn't make sense anymore and there should be some other kind of error handling.

fuse_core/include/fuse_core/util.hpp
giafranchini: Yes, this works fine. I edited it.

jakemclaughlin6:
Just curious, will this also be ported to devel for ros 1? Seems like an ordeal in and of itself

svwilliams:
> will this also be ported to devel for ros 1

That is my plan. Most of this work is not overly dependent on ROS 2 functionality. It should port back to ROS 1 without too much effort. :crossed_fingers:

giafranchini:
Hi, are there any news on this?

roncapat:
@svwilliams may I ask again some updates on this?

svwilliams:
Yes. Sorry this is taking so long to resolve. Adding 3D variable types to the system is highly valuable and we greatly appreciate all of this work. We are actively working on reviewing this PR at the moment. I hoping to have some feedback for you shortly.

fuse_models/include/fuse_models/omnidirectional_3d.hpp
```diff
@@ -0,0 +1,244 @@
...
+
+namespace fuse_models
+{
+
+/**
+ * @brief A fuse_models 3D kinematic model that generates kinematic constraints between provided
+ *        time stamps, and adds those constraints to the fuse graph.
+ *
+ * This class uses a unicycle kinematic model for the robot. It is equivalent to the motion model
+ * in the robot_localization state estimation nodes.
```
henrygerardmoore: I think this is leftover from reusing the `unicycle_2d` model.

fuse_models/include/fuse_models/omnidirectional_3d_ignition.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+namespace fuse_models
+{
+
+/**
+ * @brief A fuse_models ignition sensor designed to be used in conjunction with the Omnidirectional 3D
+ *        motion model.
+ *
+ * This class publishes a transaction that contains a prior on each state subvariable used in the
+ * Omnidirectional 3D motion model (x, y, z, qx, qy, qz, x_vel, y_vel, z_vel, roll_vel, pitch_vel,
```
henrygerardmoore: Should the qx, qy, qz here (and on line 80) be roll, pitch, and yaw to match the velocity and the definitions in `omnidirectional_3d.hpp`?

fuse_models/include/fuse_models/omnidirectional_3d.hpp
```diff
@@ -0,0 +1,244 @@
...
+                                                   //!< scaling the process noise covariance
+  double velocity_angular_norm_min_{1e-3};         //!< The minimum twist norm allowed when
+                                                   //!< scaling the process noise covariance
+  bool disable_checks_{false};    //!< Whether to disable the validation checks for the current and
+                                  //!< predicted state, including the process noise covariance after
+                                  //!< it is scaled and multiplied by dt
+  StateHistory state_history_;    //!< History of optimized graph pose estimates
+};
+
+std::ostream & operator<<(std::ostream & stream, const Omnidirectional3D & unicycle_2d);
```
henrygerardmoore:
```suggestion
std::ostream & operator<<(std::ostream & stream, const Omnidirectional3D & omnidirectional_3d);
```

fuse_models/include/fuse_models/omnidirectional_3d.hpp
```diff
@@ -0,0 +1,244 @@
...
+
+}  // namespace fuse_models
+
+#endif  // FUSE_MODELS__UNICYCLE_3D_HPP_
```
henrygerardmoore:
```suggestion
#endif  // FUSE_MODELS__OMNIDIRECTIONAL_3D_HPP_
```

fuse_models/include/fuse_models/omnidirectional_3d.hpp
```diff
@@ -0,0 +1,244 @@
...
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef FUSE_MODELS__UNICYCLE_3D_HPP_
+#define FUSE_MODELS__UNICYCLE_3D_HPP_
```
henrygerardmoore:
```suggestion
#ifndef FUSE_MODELS__OMNIDIRECTIONAL_3D_HPP_
#define FUSE_MODELS__OMNIDIRECTIONAL_3D_HPP_
```

fuse_models/include/fuse_models/omnidirectional_3d.hpp
```diff
@@ -0,0 +1,244 @@
...
+
+#include <map>
+#include <string>
+#include <utility>
+#include <vector>
```
henrygerardmoore:
```suggestion
#include <string>
#include <vector>
```

`<utility>` is unused, I think

henrygerardmoore (commented):
Found a couple leftovers in `omnidirectional_3d.hpp`.
Thanks so much for making this PR! I hope soon to contribute a basic tutorial using these in a similar vein to the existing tutorials to help people get up and running with the 3D models, and noticed these while I was working on that.

fuse_models/include/fuse_models/omnidirectional_3d_ignition.hpp
giafranchini: Yes, these should be roll, pitch, yaw since they are converted to quaternion later in the code. I will clean this too

giafranchini:
> Found a couple leftovers in `omnidirectional_3d.hpp`. Thanks so much for making this PR! I hope soon to contribute a basic tutorial using these in a similar vein to the existing tutorials to help people get up and running with the 3D models, and noticed these while I was working on that.

Yes, I left these behind during some copy-paste. Thank you for the findings and for the help with the tutorials!!

fuse_models/include/fuse_models/common/sensor_proc.hpp
```diff
@@ -1225,6 +2097,117 @@ inline bool processAccelWithCovariance(
...
+  }
+
+  // Create the full mean vector and covariance for the constraint
+  fuse_core::Vector3d accel_mean;
+  accel_mean << 
+    transformed_message.accel.accel.linear.x, 
+    transformed_message.accel.accel.linear.y,
+    transformed_message.accel.accel.linear.z;
+
+  Eigen::Map<const fuse_core::Matrix3d> accel_covariance_map(transformed_message.accel.covariance.data());
```
henrygerardmoore:
```suggestion
  Eigen::Map<const fuse_core::Matrix3d, 0, Eigen::Stride<6, 1>> accel_covariance_map(transformed_message.accel.covariance.data());
```

I think you need this stride here. Without it, feeding it a cov mat with 1,2,3,4,5,6,7,8,9 resulted in:
```
1 2 3
0 0 0
4 5 6
```

With it, it results in 
```
1 2 3
4 5 6
7 8 9
```
As you'd expect. This is because without the stride term, Eigen has no way to know you want it to treat the double[36] as a 6x6, so it just grabs the first 9 elements (1 2 3 0 0 0 4 5 6) instead of skipping the zeros

henrygerardmoore: You can verify this by just launching an IMU node and then manually publishing to its topic. I found it in the process of adding the tutorials I mentioned.

henrygerardmoore (commented):
Found one other thing while working on the 3D tutorial.

fuse_models/include/fuse_models/common/sensor_proc.hpp
giafranchini: I agree! thanks

fuse_core/include/fuse_core/util.hpp
```diff
@@ -150,6 +151,180 @@ Eigen::Matrix<T, 2, 2, Eigen::RowMajor> rotationMatrix2D(const T angle)
...
+        (q2 * q3 * two_theta) / std::pow(sin_theta2, 1.5);
+      jacobian_map(2, 3) =
+        two_theta / sin_theta + 
+        (2.0 * q0 * q3 * q3) / (sin_theta2 * q_sum2) - 
+        (q3 * q3 * two_theta) / std::pow(sin_theta2, 1.5);
+    } else {
+      jacobian_map.setZero();
+      jacobian_map(1, 1) = 2.0;
+      jacobian_map(2, 2) = 2.0;
+      jacobian_map(3, 3) = 2.0;
```
henrygerardmoore:
```suggestion
      jacobian_map(0, 0) = 2.0;
      jacobian_map(1, 1) = 2.0;
      jacobian_map(2, 2) = 2.0;
```

I believe (3,3) is out of bounds and this is a typo.

henrygerardmoore (commented):
I think I found an error (found because Eigen will complain when you hit this condition).

fuse_core/include/fuse_core/util.hpp
giafranchini: Thank you for the findings! I checked that function and I believe that yes, there is an error since the jacobian is a 3x4 matrix so index (3,3) is out of bounds, but I believe that your suggested changes are not entirely correct. 
I implemented the jacobians here by symbolic differentiation of https://github.com/ceres-solver/ceres-solver/blob/c29b5257e23f91d6a47c4db9d57350ed4985ea46/include/ceres/rotation.h#L345. Specifically, this is the case where the quaternion represents a "zero" rotations, so lines: https://github.com/ceres-solver/ceres-solver/blob/c29b5257e23f91d6a47c4db9d57350ed4985ea46/include/ceres/rotation.h#L378-L387.

The jacobian of function output (angle axis) wrt input (quaternion) is then a 3x4 matrix where elements in every line are the derivative of angle axis component wrt all the quaternion components in the order `qw, qx, qy, qz`, resulting in:

```cpp
jacobian = [0 2 0 0
            0 0 2 0
            0 0 0 2]
```

So the changes should be:

```cpp
jacobian_map(0, 1) = 2.0;
jacobian_map(1, 2) = 2.0;
jacobian_map(2, 3) = 2.0;
```
Do you agree?

henrygerardmoore: Yes, that makes sense! Good catch.

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -253,6 +253,8 @@ void FixedLagSmoother::optimizationLoop()
       // Optimize the entire graph
       summary_ = graph_->optimize(params_.solver_options);
 
+      // std::cout << summary_.FullReport() << std::endl;
```
DavidLocus: This can probably be removed

DavidLocus (approved):
I have now been through all of the code in this PR. There's a couple of comments from Stephen that I think are still outstanding, but from what I see, structurally and mathematically, this all looks good.

fuse_core/include/fuse_core/util.hpp
giafranchini: should be solved now, I add a test similar to the one used in ceres to check that the jacobians are equal even for very small rotations. For reference see: https://github.com/ceres-solver/ceres-solver/blob/c29b5257e23f91d6a47c4db9d57350ed4985ea46/internal/ceres/rotation_test.cc#L287-L296

fuse_models/src/omnidirectional_3d.cpp
```diff
@@ -0,0 +1,629 @@
...
+
+// Register this motion model with ROS as a plugin.
+PLUGINLIB_EXPORT_CLASS(fuse_models::Omnidirectional3D, fuse_core::MotionModel)
+
+namespace std
+{
+
+inline bool isfinite(const fuse_core::Vector3d & vector)
+{
+  return std::isfinite(vector.x()) && std::isfinite(vector.y() && std::isfinite(vector.z()));
```
henrygerardmoore:
```suggestion
  return std::isfinite(vector.x()) && std::isfinite(vector.y()) && std::isfinite(vector.z());
```

I think this is a typo. This implicitly casts `vector.y()` to a bool and then calls `std::isfinite` on a bool (not even sure why `std::isfinite` *can* be called on a bool :sweat_smile:)

fuse_models/src/omnidirectional_3d.cpp
```diff
@@ -0,0 +1,629 @@
...
+  if (!disable_checks_) {
+    try {
+      validateMotionModel(state1, state2, process_noise_covariance);
+    } catch (const std::runtime_error & ex) {
+      RCLCPP_ERROR_STREAM_THROTTLE(
+        logger_, *clock_, 10.0 * 1000,
+        "Invalid '" << name_ << "' motion model: " << ex.what());
+      return;
+    }
+  }
```
henrygerardmoore, line 464:
```suggestion
  // Scale process noise covariance pose by the norm of the current state twist
  auto process_noise_covariance = process_noise_covariance_;
  if (scale_process_noise_) {
    common::scaleProcessNoiseCovariance(
      process_noise_covariance, state1.vel_linear, state1.vel_angular,
      velocity_linear_norm_min_, velocity_angular_norm_min_);
  }

  // Validate
  process_noise_covariance *= dt;

  if (!disable_checks_) {
    try {
      validateMotionModel(state1, state2, process_noise_covariance);
    } catch (const std::runtime_error & ex) {
      RCLCPP_ERROR_STREAM_THROTTLE(
        logger_, *clock_, 10.0 * 1000,
        "Invalid '" << name_ << "' motion model: " << ex.what());
      return;
    }
  }
  
  state_history_.emplace(beginning_stamp, std::move(state1));
  state_history_.emplace(ending_stamp, std::move(state2));
```
I believe `state1` and `state2` are being used after move here. Does switching the order still work? Seems like it should since none of the switched stuff after relies on `state_history_` as far as I can tell.

giafranchini: Yes, it's true that in this case `state1` and `state2` are being used after move. I basically copied the implementation of `generateMotionModel` of the `Unicycle2D` class, where it's done in the same way.
Anyway, after some testing I found that the two object member values are not modified after a call to `std::move()`, I believe that this is since `state1` and `state2` are of `StateHistoryElement` struct type that:
- does not implement a custom move constructor
- has member types that recall to primitive types such as float, double (I'm not entirely sure about UUIDs which are `boost::uuids::uuid` but they behave in the same way), which remain unchanged after move

For example, I tried adding another `std::string` member to `StateHistoryElement` struct, assign a value to it, and I checked that after move it is left in valid and uninitialized state (an empty string in this case).

About your suggested changes, the only thing that I am not sure of is that in this case, if `validateMotionModel` fails than the two states are not inserted in the state history, while this happens in the current implementation. I don't know the exact consequences of this.
Maybe @svwilliams can you helps us here? Thank you!

servos:
@svwilliams any update on getting this merged?

fuse_models/include/fuse_models/omnidirectional_3d_predict.hpp
```diff
@@ -0,0 +1,671 @@
...
+    vel_angular2.x(),
+    vel_angular2.y(),
+    vel_angular2.z(),
+    acc_linear2.x(),
+    acc_linear2.y(),
+    acc_linear2.z(),
+    jacobians.data(),
+    jacobian_quat2rpy);
+
+  jacobian << J[0], J[1], J[2], J[3], J[4];
```
henrygerardmoore, line 661: `jacobian` is a 15x15, right? But isn't this inserting 15 rows and 16 columns (since J[1] is 4 columns)? Sorry I keep coming back with random questions, thank you so much for making this PR!

I ran into this only while using `predict` while running in debug, where I hit an eigen_assert (`Too many coefficients passed to comma initializer (operator<<)`) that only happens in debug mode.

fuse_models/src/odometry_3d_publisher.cpp
```diff
@@ -0,0 +1,624 @@
...
+      velocity_linear,
+      velocity_angular,
+      acceleration_linear,
+      dt,
+      position,
+      orientation,
+      velocity_linear,
+      velocity_angular,
+      acceleration_linear,
+      jacobian);
```
henrygerardmoore, line 514: should `acceleration_linear` be initialized to zero here? otherwise predict will use an uninitialized value if you aren't predicting with acceleration.

roncapat:
@svwilliams any updates / roadmap for this to get merged?
Is there a clear list of blocking points?

PR #355, opened 2024-01-29
Fix qwt deprecated declarations

efernandez:
This fixes `-Werror=deprecated-declarations` errors in a unit test that is built with `colcon build --mixin ws-test --cmake-args BUILD_WITH_PLOT_TESTS=ON --packages-select fuse_loss` (only one error shown):
```bash
fuse/fuse_loss/include/fuse_loss/qwt_loss_plot.h:254:65: error: ‘static QVector<T> QVector<T>::fromStdVector(const std::vector<T>&) [with T = double]’ is deprecated: Use QVector<T>(vector.begin(), vector.end()) instead. [-Werror=deprecated-declarations]
  254 |     curve->setSamples(residuals_, QVector<double>::fromStdVector(values));
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
```


efernandez:
Sorry, wrong target

PR #361, opened 2024-03-21
Port the "Add reset to batch optimizer" patch to ROS 2 Rolling

svwilliams:
Port the "Add reset to batch optimizer" patch (#360) to ROS 2 Rolling

### DIFF
--- a/fuse_optimizers/include/fuse_optimizers/batch_optimizer.hpp
+++ b/fuse_optimizers/include/fuse_optimizers/batch_optimizer.hpp
@@ -38,6 +38,7 @@
 #include <atomic>
 #include <condition_variable>
 #include <map>
+#include <memory>
 #include <mutex>
 #include <set>
 #include <string>
@@ -51,6 +52,9 @@
 #include <fuse_optimizers/batch_optimizer_params.hpp>
 #include <fuse_optimizers/optimizer.hpp>
 #include <fuse_graphs/hash_graph.hpp>
+
+#include <rclcpp/rclcpp.hpp>
+#include <std_srvs/srv/empty.hpp>
 
 
 namespace fuse_optimizers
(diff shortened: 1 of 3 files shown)


fuse_optimizers/include/fuse_optimizers/batch_optimizer.hpp
```diff
@@ -147,30 +151,49 @@ class BatchOptimizer : public Optimizer
    */
   using TransactionQueue = std::multimap<rclcpp::Time, TransactionQueueElement>;
 
-  fuse_core::Transaction::SharedPtr combined_transaction_;  //!< Transaction used aggregate
-                                                            //!< constraints and variables from
-                                                            //!< multiple sensors and motions models
-                                                            //!< before being applied to the graph.
-  std::mutex combined_transaction_mutex_;  //!< Synchronize access to the combined transaction
-                                           //!< across different threads
+  // Read-only after construction
```
svwilliams, line 154: I'm doing a bit of variable rearranging. This is not strictly required for this patch, but I wanted to mirror the FixedLagSmoother pattern in the BatchOptimizer.

fuse_optimizers/test/launch_tests/test_fixed_lag_ignition.cpp
```diff
@@ -131,6 +131,11 @@ TEST_F(FixedLagIgnitionFixture, SetInitialState)
   pose_msg1.pose.covariance[35] = 1.0;
   relative_pose_publisher->publish(pose_msg1);
 
+  /// @todo(swilliams) There is a timing or message queuing issue. This sleep "solves"
```
svwilliams: This unrelated unit test was failing on my machine. It appears to be a timing or message queue issue. The Pose2D sensor is only receiving the second published messages with a timestamp of 3.0s. The previous message with a timestamp of 2.0s is never processed. As a quick fix, I just added a delay. But this seems like a problem that should be fixable by changing a message queue size somewhere.

carlos-m159 (approved):

DavidLocus (approved):

PR #362, opened 2024-03-22
Adding stop and start functionality to the fixed lag smoother

DavidLocus:
This PR adds the ability to stop the fixed lag smoother and restart it at a later time, to keep it from consuming CPU when it's not needed.

This PR makes the assumption that whenever you restart the optimizer, any data that was previously held in the optimizer is no longer valid, and should be removed.

### DIFF
--- a/fuse_optimizers/src/fixed_lag_smoother.cpp
+++ b/fuse_optimizers/src/fixed_lag_smoother.cpp
@@ -436,42 +436,34 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction, const r
 
 bool FixedLagSmoother::resetServiceCallback(std_srvs::Empty::Request&, std_srvs::Empty::Response&)
 {
-  // Tell all the plugins to stop
-  stopPlugins();
-  // Reset the optimizer state
-  {
-    std::lock_guard<std::mutex> lock(optimization_requested_mutex_);
-    optimization_request_ = false;
-  }
-  started_ = false;
-  ignited_ = false;
-  setStartTime(ros::Time(0, 0));
-  // DANGER: The optimizationLoop() function obtains the lock optimization_mutex_ lock and the
-  //         pending_transactions_mutex_ lock at the same time. We perform a parallel locking scheme here to
-  //         prevent the possibility of deadlocks.
-  {
-    std::lock_guard<std::mutex> lock(optimization_mutex_);
-    // Clear all pending transactions
-    {
-      std::lock_guard<std::mutex> lock(pending_transactions_mutex_);
-      pending_transactions_.clear();
-    }
-    // Clear the graph and marginal tracking states
-    graph_->clear();
-    marginal_transaction_ = fuse_core::Transaction();
-    timestamp_tracking_.clear();
-    lag_expiration_ = ros::Time(0, 0);
-  }
-  // Tell all the plugins to start
-  startPlugins();
-  // Test for auto-start
-  autostart();
+  stop();
+  start();
 
   return true;
 }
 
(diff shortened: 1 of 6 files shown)


fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -460,6 +470,48 @@ bool FixedLagSmoother::resetServiceCallback(std_srvs::Empty::Request&, std_srvs:
   return true;
 }
 
+bool FixedLagSmoother::stopServiceCallback(std_srvs::Empty::Request&, std_srvs::Empty::Response&)
```
svwilliams: To avoid code duplication, can you update the "reset" callback make use of the stop and start callbacks somehow? For example, you could move the start and stop code into "simple" functions. And then have the service callbacks use the simple functions. Something like:
```
stopServiceCallback(std_srvs::Empty::Request&, std_srvs::Empty::Response&)
{
  stop();
}
startServiceCallback(std_srvs::Empty::Request&, std_srvs::Empty::Response&)
{
  start();
}
resetServiceCallback(std_srvs::Empty::Request&, std_srvs::Empty::Response&)
{
  stop();
  start();
}

protected:
void start()
{
  ...
}

void stop()
{
  ...
}
```

svwilliams (changes requested):
* Reduce code duplication between the stop and reset services
* Can you implement this same functionality in the batch optimizer? The batch optimizer now has a reset service (as of a day ago), so it should be easy to follow this same pattern.

fuse_optimizers/src/fixed_lag_smoother.cpp
DavidLocus: Done and done

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -425,6 +435,34 @@ void FixedLagSmoother::processQueue(fuse_core::Transaction& transaction, const r
...
+{
+  stop();
+  return true;
+}
+
+bool FixedLagSmoother::startServiceCallback(std_srvs::Empty::Request&, std_srvs::Empty::Response&)
+{
+  start();
+  return true;
+}
```
needphpsql, line 455: The request and response are not used. I assume this is only a stub for future filling in.

needphpsql (approved):

fuse_optimizers/src/fixed_lag_smoother.cpp
DavidLocus: Not a stub, there's just no information currently needed for a start or stop, the same as how the reset callback also doesn't use the request or response fields.

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -497,6 +529,7 @@ void FixedLagSmoother::transactionCallback(
       // ...check if we should
       if (sensor_models_.at(sensor_name).ignition)
       {
+        ROS_INFO_STREAM("Ignition occured");
```
svwilliams: I have no opinions here whatsoever. Just asking questions:
Do you like this new log line printed all the time? Or would DEBUG be a better fit? Should the same message be implemented in the batch optimizer?

svwilliams (approved):

fuse_optimizers/src/fixed_lag_smoother.cpp
DavidLocus: Whoops, left this in by accident. I was testing... something, I don't remember what, and forgot to remove this. Doing that now

PR #365, opened 2024-04-15
Rotate position in motion model constraint

jakemclaughlin6:


jakemclaughlin6:
I've only implemented this for the autodiff cost functor, but I could probably do the manual diff as well it will just take me a little more time

jakemclaughlin6:
@svwilliams @ayrton04 I'm not sure how to request a review directly so just commenting here

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -115,9 +113,7 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
    *                         computed for the parameters where jacobians[i] is not NULL.
    * @return The return value indicates whether the computation of the residuals and/or jacobians was successful or not.
    */
-  bool Evaluate(double const* const* parameters,
-                double* residuals,
-                double** jacobians) const override
+  bool Evaluate(double const* const* parameters, double* residuals, double** jacobians) const override
```
ayrton04, line 116: Just to minimize changes, it's probably a good idea to not reformat this sort of hting. If it passed linting before (and it must have, or the build farm would have complained), I'd recommend keeping it as-is.

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -274,9 +279,7 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
   fuse_core::Matrix8d A_;  //!< The residual weighting matrix, most likely the square root information matrix
 };
 
-Unicycle2DStateCostFunction::Unicycle2DStateCostFunction(const double dt, const fuse_core::Matrix8d& A) :
-  dt_(dt),
-  A_(A)
+Unicycle2DStateCostFunction::Unicycle2DStateCostFunction(const double dt, const fuse_core::Matrix8d& A) : dt_(dt), A_(A)
```
ayrton04, line 277: Same comment re: formatting.

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
```diff
@@ -128,7 +128,7 @@ TEST(CostFunction, evaluateCostFunction)
 
   for (size_t i = 0; i < num_parameter_blocks; ++i)
   {
-    EXPECT_MATRIX_NEAR(J_autodiff[i], J[i], std::numeric_limits<double>::epsilon())
+    EXPECT_MATRIX_NEAR(J_autodiff[i], J[i], 1e-15)
```
ayrton04: Odd that this test should now fail.

ayrton04:
I'm going to tag @efernandez, as `blame` tells me he wrote this originally.

Anyway, just thinking aloud here. We have three poses:

```
pose1           = (x1,    y1,    yaw1   ) = (parameters[0][0], parameters[0][1], parameters[1][0])
pose2           = (x2,    y2,    yaw2   ) = (parameters[5][0], parameters[5][1], parameters[6][0])
pose2_predicted = (xpred, ypred, yawpred) = (position_pred_x,  position_pred_y,  yaw_pred        )
```

The residual we want is effectively the `pose2 - pose2_predicted`. Given that both poses can be modeled as transform matrices, wouldn't we therefore want `pose2_predicted^-1 * pose2`? That would yield:

```
    double cos_yaw = std::cos(-yaw_pred);  // Note negative sign for inverse of pose2_predicted
    double sin_yaw = std::sin(-yaw_pred);  //
    ...
    <the rest as-is>
```

But I would also probably eschew the Eigen rotation matrix and just do it directly:

```
auto delta_x = parameters[5][0] - position_pred_x;
auto delta_y = parameters[5][1] - position_pred_y;
auto sin_pred_inv = std::sin(-yaw_pred);
auto cos_pred_inv= std::cos(-yaw_pred);
residuals[0] = cos_pred_inv * delta_x - sin_pred_inv * delta_y;
residuals[1] = sin_pred_inv * delta_x + cos_pred_inv * delta_y;
```

In any case, I'm not sure if it's correct for the residual to be in the frame of `pose1`? Again, I might be wrong about that.

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
jakemclaughlin6: I'll double check that it passes before with epsilon, regardless std::numeric_limits<double>::epsilon() seems very small for a jacobian test. I'm assuming it doesn't pass now due to rotations in the residual computation

jakemclaughlin6:
@ayrton04 The reasoning behind putting the pose error into pose1's frame is mainly due to an issue found [in this PR](https://github.com/locusrobotics/fuse/pull/325) where the covariance is not in the same frame as the error. Since these constraints are relative, its normal to assume the covariance should be in the local frame. Currently the covariance and pose error are in the global frame, which can lead to someone accidentally defining their covariance in the local frame

svwilliams:
I'm trying to get my head wrapped around this. "What frame is the covariance represented in?" is a perennial question in my experience.

In devel right now, the residuals are computed as:
```
  residuals_map(0) = position2[0] - position_pred[0];
  residuals_map(1) = position2[1] - position_pred[1];
  residuals_map(2) = yaw2[0] - yaw_pred[0];
```

`position2` is represented in the map frame, as is `position_pred`. The residual computation is not doing any frame transformations, and so the residuals are represented in the map frame (or at least in the map frame orientation).

But that doesn't seem like how we would want a relative pose to work.
Instead, if we compute the predicted change/delta between Pose1 and Pose2 using the motion model (delta_predicted),
and we compute our "true" change between Pose1 and Pose2 (delta), then the residual would look something like:
```
  delta = Pose1^1 * Pose2;
  delta_predicted = Pose1^1 * Pose2Predicted;

  residuals_map(0) = delta[0] - delta_predicted[0];
  residuals_map(1) = delta[1] - delta_predicted[1];
  residuals_map(2) = delta[2] - delta_predicted[2];
```
Both `delta` and `delta_predicted` are represented in the Pose1 frame, and so the residual is also in the Pose1 frame.

And that matches my intuition about what frame the covariance should be in. The measurement is a transform from Pose1 to Pose2. I.e. The robot moved forward 1m. Thus, the measurement is represented in the Pose1 frame. It makes certain intuitive sense for the covariance to be represented in the same frame as the measurement.

svwilliams:
And as an implementation note, we should be able to compute the predicted delta directly from the `predict()` function by providing the starting point as the origin instead of as Pose1. There should be no need to accumulate the position change starting at Pose1, only to then "subtract out" Pose1 afterwards.
```
  T delta_position_pred[2];
  T delta_yaw_pred[1];
  T vel_linear_pred[2];
  T vel_yaw_pred[1];
  T acc_linear_pred[2];
  predict(
    [0.0, 0.0],
    [0.0],
    vel_linear1,
    vel_yaw1,
    acc_linear1,
    T(dt_),
    delta_position_pred,
    delta_yaw_pred,
    vel_linear_pred,
    vel_yaw_pred,
    acc_linear_pred);
```

ayrton04:
Makes sense to me!

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
ayrton04: That makes sense.

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -147,8 +147,12 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
...
 
-    residuals[0] = parameters[5][0] - position_pred_x;
-    residuals[1] = parameters[5][1] - position_pred_y;
+    // rotate the position residual into the local frame
+    auto sin_pred_inv = std::sin(-yaw_pred);
+    auto cos_pred_inv= std::cos(-yaw_pred);
+
+    residuals[0] = cos_pred_inv * delta_position_pred_x - sin_pred_inv * delta_position_pred_y;
+    residuals[1] = sin_pred_inv * delta_position_pred_x + cos_pred_inv * delta_position_pred_y;
     residuals[2] = parameters[6][0] - yaw_pred;
```
efernandez: I don't see why `residuals[0]` and `residual[1]` no longer depend on the previous pose (defined in `parameters[0]` and `parameters[1]`, i.e. `position1_x`, `position1_y` and `yaw1`) and the next pose (defined in `parameters[5]` and `parameters[6]`, i.e. `position2_x`, `position2_y` and `yaw2`).

Shouldn't we be computing the delta between those two and then compute the residuals as the delta between that and the predicted pose. That is:
```suggestion
    const Eigen::Matrix<T, 2, 1> position1(parameters[0][0], parameters[0][1]);
    const Eigen::Matrix<T, 2, 1> position2(parameters[5][0], parameters[5][1]);

    const Eigen::Matrix<T, 2, 1> delta_position = fuse_core::rotationMatrix2D(parameters[1][0]).transpose() * (position2 - position1);
    const T delta_yaw = parameters[6][0] - parameters[1][0];  // omitting fuse_core::wrapAngle2D because it is done later on

    const Eigen::Matrix<T, 2, 1> delta_position_pred(delta_position_pred_x, delta_position_pred_y);

    residuals.template head<2>() = fuse_core::rotationMatrix2D(delta_yaw).transpose() * (delta_position_pred - delta_position);
    residuals[2] = delta_yaw - yaw_pred;  // omitting fuse_core::wrapAngle2D because it is done later on
```

In a way, this is similar to https://github.com/locusrobotics/fuse/blob/2c652c7450c24f9fe462de0c33d5c2c16e21c37d/fuse_constraints/include/fuse_constraints/normal_delta_pose_2d_cost_functor.h#L118-L125

fuse_models/include/fuse_models/unicycle_2d_state_cost_functor.h
```diff
@@ -147,27 +145,36 @@ bool Unicycle2DStateCostFunctor::operator()(
   const T* const acc_linear2,
   T* residual) const
 {
-  T position_pred[2];
```
efernandez, line 150: I believe the comment I made above applies here as well. Indeed, I suspect with your change the compiler must be complaining of unused arguments, isn't it?

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
```diff
@@ -128,7 +128,7 @@ TEST(CostFunction, evaluateCostFunction)
 
   for (size_t i = 0; i < num_parameter_blocks; ++i)
   {
-    EXPECT_MATRIX_NEAR(J_autodiff[i], J[i], std::numeric_limits<double>::epsilon())
```
efernandez, line 131: This was probably too strict, but if we change the cost function or functor, so the residuals are computed differently, I'd expect the analytic jacobians are also computed differently. So I'm afraid that even a small error here is actually a consequence of an incorrect analytic jacobian. Maybe the test values are hiding that because they don't have a strong contribution to the jacobians that should've probably been updated to account for the residuals computation changes. :thinking:

efernandez (changes requested):

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
jakemclaughlin6: In this case, I feel it would just be better to keep what I originally had, and just rotate the error back into the local frame

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
```diff
@@ -99,7 +99,17 @@ TEST(CostFunction, evaluateCostFunction)
 
   // Check jacobians are correct using a gradient checker
   ceres::NumericDiffOptions numeric_diff_options;
-  ceres::GradientChecker gradient_checker(&cost_function, nullptr, numeric_diff_options);
```
efernandez, line 102: This looks like a change from another MR

If possible, get rid of this change here

fuse_models/include/fuse_models/unicycle_2d_state_cost_functor.h
```diff
@@ -163,19 +163,21 @@ bool Unicycle2DStateCostFunctor::operator()(
...
     acc_linear_pred);
 
   // rotate the position residual into the local frame
-  T sin_pred_inv = ceres::sin(-yaw_pred[0]);
-  T cos_pred_inv = ceres::cos(-yaw_pred[0]);
+  const Eigen::Matrix<T, 2, 1> position1_map(position1[0], position1[1]);
+  const Eigen::Matrix<T, 2, 1> position2_map(position2[0], position2[1]);
+  const Eigen::Matrix<T, 2, 1> delta_position = fuse_core::rotationMatrix2D(-yaw1[0]) * (position2_map - position1_map);
+  const T delta_yaw = yaw2[0] - yaw1[0];  // omitting fuse_core::wrapAngle2D because it is done later on
+  const Eigen::Matrix<T, 2, 1> delta_position_pred_map(delta_position_pred[0], delta_position_pred[1]);
```
efernandez: These aren't really Eigen maps, but I just realized you can use `position1` because it's already taken.

Just highlighting that. I don't really have a better option, so it's ok for me. Maybe others have a better suggestion :thinking:

fuse_models/include/fuse_models/unicycle_2d_state_cost_functor.h
```diff
@@ -147,28 +145,39 @@ bool Unicycle2DStateCostFunctor::operator()(
...
+  T delta_position_pred[2];
+  T delta_yaw_pred[1];
   T vel_linear_pred[2];
   T vel_yaw_pred[1];
   T acc_linear_pred[2];
+  T init_position[2];
+  T init_yaw[1];
+  init_position[0] = (T)0.0;
+  init_position[1] = (T)0.0;
+  init_yaw[0] = (T)0.0;
```
efernandez: I believe it's more common to see the following:
```suggestion
  init_position[0] = T(0.0);
  init_position[1] = T(0.0);
  init_yaw[0] = T(0.0);
```

efernandez (changes requested):

efernandez:
This now LGTM. I just made some minor comments.

Does this look good to you now @svwilliams @ayrton04 ? Does it align with what you mentioned before @svwilliams ?

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -119,26 +119,26 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
                 double* residuals,
                 double** jacobians) const override
   {
-    double position_pred_x;
-    double position_pred_y;
+    double delta_position_pred_x;
+    double delta_position_pred_y;
     double yaw_pred;
```
svwilliams: This should be renamed to `delta_yaw_pred` to be consistent with the other variables

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -147,20 +147,25 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
...
-    residuals[6] = parameters[9][0] - acc_linear_pred_x;
-    residuals[7] = parameters[9][1] - acc_linear_pred_y;
+    const Eigen::Vector2d position1(parameters[0][0], parameters[0][1]);
+    const Eigen::Vector2d position2(parameters[5][0], parameters[5][1]);
+    const Eigen::Vector2d delta_position = fuse_core::rotationMatrix2D(-parameters[1][0]) * (position2 - position1);
+    const double delta_yaw = parameters[6][0] - parameters[1][0]; 
+    const Eigen::Vector2d delta_position_pred(delta_position_pred_x, delta_position_pred_y);
+
+    Eigen::Map<Eigen::Matrix<double, 8, 1>> residuals_map(residuals);
+    residuals_map.head<2>() = fuse_core::rotationMatrix2D(-delta_yaw) * (delta_position - delta_position_pred);
```
svwilliams: I'm not sure I understand why the difference of position changes are being transformed into....is this the Pose2 frame?
I feel like this should simply be:
```
residuals_map(0) = delta_position(0) - delta_position_pred(0);
residuals_map(1) = delta_position(1) - delta_position_pred(1);
```

Both `delta_position` and `delta_position_pred` are represented in the Pose1 frame. If we just subtract the positions, that error will also be in the Pose1 frame/orientaiton. And that seems like the natural frame to compute the measurement covariance in as well.

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -147,20 +145,26 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
...
 
-    residuals[0] = parameters[5][0] - position_pred_x;
-    residuals[1] = parameters[5][1] - position_pred_y;
-    residuals[2] = parameters[6][0] - yaw_pred;
-    residuals[3] = parameters[7][0] - vel_linear_pred_x;
-    residuals[4] = parameters[7][1] - vel_linear_pred_y;
-    residuals[5] = parameters[8][0] - vel_yaw_pred;
-    residuals[6] = parameters[9][0] - acc_linear_pred_x;
-    residuals[7] = parameters[9][1] - acc_linear_pred_y;
+    const double delta_yaw = parameters[6][0] - parameters[1][0];
```
svwilliams: To be consistent with naming
```suggestion
    const double delta_yaw_est = parameters[6][0] - parameters[1][0];
```

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -147,20 +145,26 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
...
+    const Eigen::Vector2d position1(parameters[0][0], parameters[0][1]);
+    const Eigen::Vector2d position2(parameters[5][0], parameters[5][1]);
+    const Eigen::Vector2d delta_position_est = fuse_core::rotationMatrix2D(-parameters[1][0]) * (position2 - position1);
+    const Eigen::Vector2d delta_position_pred(delta_position_pred_x, delta_position_pred_y);
+    const fuse_core::Matrix2d R_delta_yaw_inv = fuse_core::rotationMatrix2D(-delta_yaw);
+
+    Eigen::Map<Eigen::Matrix<double, 8, 1>> residuals_map(residuals);
+    residuals_map.head<2>() = R_delta_yaw_inv * (delta_position_pred - delta_position_est);
```
svwilliams: The `delta_position_est` is computed as Pose1^1 * Pose2. That effectively transforms Pose2 into the frame of Pose1.

Similarly, `delta_position_pred` is computed with Pose1 as the origin. So that quantity is also in the frame of Pose1.

If we did a simple subtraction, `residual[0] = delta_position_est.x - delta_position_prod.x`, then the residual would also be represented in the frame of Pose1. Which seems like the natural frame to compute the measurement covariance as well.

Here you are rotating the error into the ... Pose2 frame?
I feel like this should just be:
```
residuals_map(0) = delta_position_est.x - delta_position_pred.x;
residuals_map(1) = delta_position_est.y - delta_position_pred.y;
residuals_map(2) = delta_yaw_est - delta_yaw_pred;
...
```
But that is not a firm belief. I could be convinced otherwise.

ayrton04:
@svwilliams @efernandez what did we conclude on this one?

efernandez:
> @svwilliams @efernandez what did we conclude on this one?

@ayrton04 - @jakemclaughlin6 still has to address @svwilliams comments, but this is a low priority task for us atm. Other than that, I believe we're all happy with the changes proposed in this PR.

jakemclaughlin6:
> > @svwilliams @efernandez what did we conclude on this one?
> 
> @ayrton04 - @jakemclaughlin6 still has to address @svwilliams comments, but this is a low priority task for us atm. Other than that, I believe we're all happy with the changes proposed in this PR.

I got a little stuck on computing the manual jacobians for the new cost functor so I put this on the backburner for a bit until I have time to get back to it

jakemclaughlin6:
@ayrton04 @svwilliams @efernandez I fixed the manual jacobians and the tests are passing. Let me know if we would want more tests and I can make some more, but the existing one should be fairly comprehensive as its running 100 randomized cases

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -115,41 +113,46 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
...
-    residuals[2] = parameters[6][0] - yaw_pred;
+    const double delta_yaw_est = parameters[6][0] - parameters[1][0];
+    const fuse_core::Vector2d position1(parameters[0][0], parameters[0][1]);
+    const fuse_core::Vector2d position2(parameters[5][0], parameters[5][1]);
+    const fuse_core::Vector2d position_diff = (position2 - position1);
+    const double sin_yaw_inv = std::sin(-parameters[1][0]);
+    const double cos_yaw_inv = std::cos(-parameters[1][0]);
+
+    residuals[0] = (cos_yaw_inv * position_diff.x() - sin_yaw_inv * position_diff.y()) - delta_position_pred_x;
+    residuals[1] = (sin_yaw_inv * position_diff.x() + cos_yaw_inv * position_diff.y()) - delta_position_pred_y;
```
efernandez: I think we should declare `R_yaw_inv` here and use it to compute the first two residuals:
```suggestion
    const fuse_core::Vector2d delta_position_delta(delta_position_delta_x, delta_position_delta_y);
    const fuse_core::Matrix2d R_yaw_inv = fuse_core::rotationMatrix2D(-parameters[1][0]);

    residuals.head<2>() = R_yaw_inv * position_diff - delta_position_delta;
```

I think you could even declare `delta_position_delta` before calling `predict` and pass `delta_position_delta.x()` and `delta_position_delta.y()` as arguments, so you avoid declaring `delta_position_delta_x` and `delta_position_delta_y`. I'm not too sure about that though.

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -160,7 +163,7 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
 
     // Scale the residuals by the square root information matrix to account for
     // the measurement uncertainty.
-    Eigen::Map<fuse_core::Vector8d> residuals_map(residuals);
+    Eigen::Map<Eigen::Matrix<double, 8, 1>> residuals_map(residuals);
```
efernandez: Mostly curious - why `fuse_core::Vector8d` doesn't work here?

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -182,6 +185,8 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
       if (jacobians[0])
       {
         Eigen::Map<fuse_core::Matrix<double, 8, 2>> jacobian(jacobians[0]);
+        const fuse_core::Matrix2d R_yaw_inv = fuse_core::rotationMatrix2D(-parameters[1][0]);
+        jacobian.block<2, 2>(0, 0) = R_yaw_inv * jacobian.block<2, 2>(0, 0);
```
efernandez: Mostly curious, but I wonder if you can do this
```suggestion
        jacobian.block<2, 2>(0, 0).applyOnTheLeft(R_yaw_inv);
```

I also wonder if the sign is correct, since we have `- position1`

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -235,6 +240,8 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
       {
         Eigen::Map<fuse_core::Matrix<double, 8, 2>> jacobian(jacobians[5]);
         jacobian = A_.block<8, 2>(0, 0);
+        const fuse_core::Matrix2d R_yaw_inv = fuse_core::rotationMatrix2D(-parameters[1][0]);
+        jacobian.block<2, 2>(0, 0) = jacobian.block<2, 2>(0, 0) * R_yaw_inv;
```
efernandez: I believe this should also work
```suggestion
        jacobian.block<2, 2>(0, 0) *= R_yaw_inv;
```

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -235,6 +240,8 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
       {
         Eigen::Map<fuse_core::Matrix<double, 8, 2>> jacobian(jacobians[5]);
         jacobian = A_.block<8, 2>(0, 0);
+        const fuse_core::Matrix2d R_yaw_inv = fuse_core::rotationMatrix2D(-parameters[1][0]);
```
efernandez: As per my previous suggestion, remove from here and declare when you compute the residuals
```suggestion
```

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -182,6 +185,8 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
       if (jacobians[0])
       {
         Eigen::Map<fuse_core::Matrix<double, 8, 2>> jacobian(jacobians[0]);
+        const fuse_core::Matrix2d R_yaw_inv = fuse_core::rotationMatrix2D(-parameters[1][0]);
```
efernandez: As per my previous suggestion, remove from here and declare when you compute the residuals
```suggestion
```

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
efernandez: Mostly curious - what happened with formatting here? :thinking:

fuse_models/include/fuse_models/unicycle_2d_state_cost_functor.h
```diff
@@ -147,38 +143,50 @@ bool Unicycle2DStateCostFunctor::operator()(
...
+  residual[1] = (sin_yaw_inv * position_diff.x() + cos_yaw_inv * position_diff.y()) - delta_position_pred[1];
+  residual[2] = delta_yaw_est - delta_yaw_pred[0];
+  residual[3] = vel_linear2[0] - vel_linear_pred[0];
+  residual[4] = vel_linear2[1] - vel_linear_pred[1];
+  residual[5] = vel_yaw2[0] - vel_yaw_pred[0];
+  residual[6] = acc_linear2[0] - acc_linear_pred[0];
+  residual[7] = acc_linear2[1] - acc_linear_pred[1];
 
-  fuse_core::wrapAngle2D(residuals_map(2));
+  // fuse_core::wrapAngle2D(residual[2]);
```
efernandez: Why is this commented out?

fuse_models/include/fuse_models/unicycle_2d_state_cost_functor.h
```diff
@@ -147,38 +143,50 @@ bool Unicycle2DStateCostFunctor::operator()(
...
-  residuals_map(7) = acc_linear2[1] - acc_linear_pred[1];
+  const T delta_yaw_est = yaw2[0] - yaw1[0];
+  const Eigen::Matrix<T, 2, 1> position1_map(position1[0], position1[1]);
+  const Eigen::Matrix<T, 2, 1> position2_map(position2[0], position2[1]);
+  const Eigen::Matrix<T, 2, 1> position_diff = (position2_map - position1_map);
+  const T sin_yaw_inv = ceres::sin(-yaw1[0]);
+  const T cos_yaw_inv = ceres::cos(-yaw1[0]);
+
+  residual[0] = (cos_yaw_inv * position_diff.x() - sin_yaw_inv * position_diff.y()) - delta_position_pred[0];
+  residual[1] = (sin_yaw_inv * position_diff.x() + cos_yaw_inv * position_diff.y()) - delta_position_pred[1];
```
efernandez: Same comment as before, although this suggestion might not work as is when `T` is a `ceres::Jet`, so it probably needs some changes:
```suggestion
  const Eigen::Matrix<T, 2, 1> delta_position_delta(delta_position_delta_x, delta_position_delta_y);
  const Eigen::Matrix<T, 2, 2> R_yaw_inv = fuse_core::rotationMatrix2D(-parameters[1][0]);

  residuals.head<2>() = R_yaw_inv * position_diff - delta_position_delta;
```

If it's too hard to make it work, just leave it as is. For the 2D rotation matrix I believe we'd need something like this: https://github.com/ceres-solver/ceres-solver/blob/0a7845fef4a87bc803a3e25e198251bf55858bfd/examples/slam/pose_graph_2d/pose_graph_2d_error_term.h#L42-L50 Maybe `fuse_core` already have that or otherwise should have it.

fuse_models/include/fuse_models/unicycle_2d_state_cost_functor.h
```diff
@@ -147,38 +143,50 @@ bool Unicycle2DStateCostFunctor::operator()(
...
+  residual[5] = vel_yaw2[0] - vel_yaw_pred[0];
+  residual[6] = acc_linear2[0] - acc_linear_pred[0];
+  residual[7] = acc_linear2[1] - acc_linear_pred[1];
 
-  fuse_core::wrapAngle2D(residuals_map(2));
+  // fuse_core::wrapAngle2D(residual[2]);
 
   // Scale the residuals by the square root information matrix to account for
   // the measurement uncertainty.
+  Eigen::Map<Eigen::Matrix<T, 8, 1>> residuals_map(residual);
```
efernandez: Again, mostly curious - why the `Eigen::Map` has to be at the end instead of where it was before?

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
```diff
@@ -56,85 +56,105 @@ TEST(CostFunction, evaluateCostFunction)
...
+
+    EXPECT_TRUE(cost_function.Evaluate(parameters, residuals.data(), jacobians.data()));
+
+    // We cannot use std::numeric_limits<double>::epsilon() tolerance because with the expected state2 above the
+    // residuals are not zero for position2.x = -4.389e-16 and yaw2 = -2.776e-16
+    EXPECT_MATRIX_NEAR(fuse_core::Vector8d::Zero(), residuals, 1e-13);
+
+    // Check jacobians are correct using a gradient checker
+    ceres::NumericDiffOptions numeric_diff_options;
+    #if !CERES_SUPPORTS_MANIFOLDS
```
efernandez: I think we should have the pre-processor macros unindented
```suggestion
#if !CERES_SUPPORTS_MANIFOLDS
```

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
```diff
@@ -56,85 +56,105 @@ TEST(CostFunction, evaluateCostFunction)
...
+    ceres::GradientChecker gradient_checker(
+      &cost_function,
+      static_cast<std::vector<const ceres::LocalParameterization*>*>(nullptr),
+      numeric_diff_options);
+    #else
```
efernandez:
```suggestion
#else
```

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
```diff
@@ -56,85 +56,105 @@ TEST(CostFunction, evaluateCostFunction)
...
+    ceres::GradientChecker gradient_checker(&cost_function, static_cast<std::vector<const ceres::Manifold*>*>(nullptr),
+                                            numeric_diff_options);
+    #endif
```
efernandez:
```suggestion
#endif
```

fuse_models/test/test_unicycle_2d_state_cost_function.cpp
```diff
@@ -56,85 +56,105 @@ TEST(CostFunction, evaluateCostFunction)
...
-    J_autodiff[i].resize(num_residuals, block_sizes[i]);
-    jacobians_autodiff[i] = J_autodiff[i].data();
-  }
-
-  EXPECT_TRUE(cost_function_autodiff.Evaluate(parameters, residuals.data(), jacobians_autodiff.data()));
-
-  const Eigen::IOFormat HeavyFmt(
-      Eigen::FullPrecision, 0, ", ", ";\n", "[", "]", "[", "]");
+  std::random_device dev;
+  std::mt19937 gen(dev());
```
efernandez: If the test fails, would we be able to reproduce it?

I think it'd be better if we recode the seed. I've never than that, but I believe we could print the seed:
```suggestion
  const auto seed = dev();
  std::mt19937 gen(seed);
```

and then pass `<< seed` to all the checks below that depend on the seed, so we can see what seed was used and then we can easily reproduce the test.

Another option is to just use a fixed seed.

efernandez (approved):
That Jacobians look way simpler. I haven't really look at the math, but if the test is passing now, that should be because they're good now.

I just made a few suggestion to simplify the code a bit more, since some operations in the Jacobians are also done for the residuals computation.

fuse_models/include/fuse_models/unicycle_2d_state_cost_functor.h
jakemclaughlin6: It shouldn't matter but the cost function and cost functor have it in different spots. I was just making it consistent across the two

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
```diff
@@ -115,52 +113,41 @@ class Unicycle2DStateCostFunction : public ceres::SizedCostFunction<8, 2, 1, 2,
...
-    residuals[2] = parameters[6][0] - yaw_pred;
-    residuals[3] = parameters[7][0] - vel_linear_pred_x;
-    residuals[4] = parameters[7][1] - vel_linear_pred_y;
-    residuals[5] = parameters[8][0] - vel_yaw_pred;
-    residuals[6] = parameters[9][0] - acc_linear_pred_x;
-    residuals[7] = parameters[9][1] - acc_linear_pred_y;
+    const double delta_yaw_est = parameters[6][0] - parameters[1][0];
+    const fuse_core::Vector2d position1(parameters[0][0], parameters[0][1]);
+    const fuse_core::Vector2d position2(parameters[5][0], parameters[5][1]);
+    const fuse_core::Vector2d position_diff = (position2 - position1);
```
efernandez:
```suggestion
    const fuse_core::Vector2d position_diff = position2 - position1;
```

fuse_models/include/fuse_models/unicycle_2d_state_cost_functor.h
```diff
@@ -147,28 +143,39 @@ bool Unicycle2DStateCostFunctor::operator()(
...
+  const Eigen::Matrix<T, 2, 1> position2_map(position2[0], position2[1]);
+  const Eigen::Matrix<T, 2, 1> position_diff = (position2_map - position1_map);
+  const fuse_core::Matrix<T, 2, 2> R_yaw_inv = fuse_core::rotationMatrix2D(yaw1[0]);
+  Eigen::Map<Eigen::Matrix<T, 2, 1>> delta_position_pred_map(delta_position_pred);
+
   Eigen::Map<Eigen::Matrix<T, 8, 1>> residuals_map(residual);
-  residuals_map(0) = position2[0] - position_pred[0];
-  residuals_map(1) = position2[1] - position_pred[1];
-  residuals_map(2) = yaw2[0] - yaw_pred[0];
+  residuals_map.block(0, 0, 2, 1) = R_yaw_inv * position_diff - delta_position_pred_map;
```
efernandez: I believe a fixed-size block expression is more efficient https://eigen.tuxfamily.org/dox/group__TutorialBlockOperations.html
```suggestion
  residuals_map.block<2, 1>(0, 0) = R_yaw_inv * position_diff - delta_position_pred_map;
```

efernandez (approved):
LGTM - just two minor comments

fuse_models/include/fuse_models/unicycle_2d_state_cost_functor.h
jakemclaughlin6: this cant be used for templated arguments

efernandez: I guess it'd have to be
```suggestion
  residuals_map.template block<2, 1>(0, 0) = R_yaw_inv * position_diff - delta_position_pred_map;
```

:point_right: But now that I look at the actual `residuals_map` type, I think you can even do
```suggestion
  residuals_map.template head<2>() = R_yaw_inv * position_diff - delta_position_pred_map;
```
or, if that doesn't work, the dynamic-size expression
```suggestion
  residuals_map.head(2) = R_yaw_inv * position_diff - delta_position_pred_map;
```

fuse_models/include/fuse_models/unicycle_2d_state_cost_function.h
jakemclaughlin6: it should be, we multiply by -A for position1 (and all "1" variables), and we use positive A for "2" variables

ayrton04 (approved):

carlos-m159 (approved):

jakemclaughlin6:
@svwilliams is this good to merge, or are you guys going to do some validations with this before choosing to merge it?

svwilliams:
I started reading through it, but I got side-tracked. Can you give me until Monday for feedback? Sorry I'm slow.

svwilliams:
Sorry, I still want to do some validations on this PR. Or at least re-review it.

efernandez:
> Sorry, I still want to do some validations on this PR. Or at least re-review it.

@svwilliams Did you have a chance to look into this MR?

I'd be happy to help, if there's anything I can help with :smiley:

PR #368, opened 2024-04-28
Required formatting changes for the latest version of ROS 2 Rolling

svwilliams:
Required formatting changes for the latest version of ROS 2 Rolling. The formatting rules seem to be something of a moving target. This reverts most of but not all of the formatting changes I added to the last PR. Now that there is a Jazzy build happening, I'm hoping this settles down.


carlos-m159 (approved):

ayrton04 (approved):

svwilliams:
The Locus build server does not like the new ROS 2 Rolling formatting rules. One of these days I'll look into a solution for that. And into why the upsteam CI tests are not running. But today is not that day.

svwilliams:
The upstream CI build tests failed, but we are down to only two failures. This is fine because the next PR #369 should fix those remaining test failures.

PR #370, opened 2024-05-15
Cast nullptr to type for manifold support

jakemclaughlin6:
Without this when ceres version == 2.1.0 we get:
`/home/jmclaughlin/colcon_ws/src/perception/fuse/fuse_models/test/test_unicycle_2d_state_cost_function.cpp:102:88: error: call of overloaded ‘GradientChecker(const fuse_models::Unicycle2DStateCostFunction*, std::nullptr_t, ceres::NumericDiffOptions&)’ is ambiguous
  102 |   ceres::GradientChecker gradient_checker(&cost_function, nullptr, numeric_diff_options);
`


efernandez (approved):
I confirm this was also failing to me. It's an ugly fix, but we need it. :+1:

svwilliams (approved):

PR #375, opened 2024-08-02
[RST-10727] Publish the start/stop status of the fixed-lag smoother

svwilliams:
Now that it is possible to construct a fixed-lag smoother without immediately starting it, knowing the running state of the optimizer may be important for coordination with other nodes. Publish the start/stop state of the optimizer as a latched topic.

The status is currently being published as a `std_msgs/Bool`. I do wonder if I should add a dedicated message type to `fuse_msgs` and publish that instead? If we ever want to expand the status information in the future, having our own message type could make that easier. We would simply add new fields to the message, but we would continue to publish the same message type. Any thoughts?

### DIFF
--- a/fuse_optimizers/src/fixed_lag_smoother.cpp
+++ b/fuse_optimizers/src/fixed_lag_smoother.cpp
@@ -39,6 +39,7 @@
 #include <fuse_core/uuid.h>
 #include <fuse_optimizers/optimizer.h>
 #include <ros/ros.h>
+#include <std_msgs/Bool.h>
 
 #include <algorithm>
 #include <iterator>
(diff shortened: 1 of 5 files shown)


fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -115,6 +116,14 @@ FixedLagSmoother::FixedLagSmoother(
     &FixedLagSmoother::startServiceCallback,
     this);
 
+  status_publisher_ = node_handle_.advertise<std_msgs::Bool>(
+    ros::names::resolve(params_.status_topic),
+    1,
+    true);
+  auto status = std_msgs::Bool();
+  status.data = false;
+  status_publisher_.publish(status);
```
carlos-m159: Maybe move this to a function?
If we at any point store the state in a member, might will be easier to maintain.

```
inline void notifyStatus(const bool running)
{
  // Update status topic
  auto status = std_msgs::Bool();
  status.data = running;
  status_publisher_.publish(status);
}
```

carlos-m159 (approved):
> The status is currently being published as a std_msgs/Bool. I do wonder if I should add a dedicated message type to fuse_msgs and publish that instead? 

My initial thoughts would be to keep it with `std_msgs/Bool` and change it if we ever require additional status info. For example, nodes that are toggled using `std_msgs/Bool` or are waiting on some signal to start up or stop: injecting a `fuse_msgs` dependency might be strange. But I am also fine with creating a custom message, and, for example, publish that continuously and not only at start and stop.

DavidLocus (approved):

PR #378, opened 2024-08-23
Fixing issue with Eigen in Ubuntu Jammy on ARM

ayrton04:
On ARM, the use of Eigen 3.4 with g++11 (both the versions in Jammy) causes a warning, which in turn causes the build to fail. Details are here: https://gitlab.com/libeigen/eigen/-/issues/2326. We can suppress the warning for ARM builds by checking for the specific versions of the affected packages and architecture.

If anyone has a more reliable (yet equally succinct) way of testing for ARM, please fill me in.

### DIFF
--- a/fuse_publishers/CMakeLists.txt
+++ b/fuse_publishers/CMakeLists.txt
@@ -31,6 +31,15 @@ catkin_package(
 ###########
 add_compile_options(-Wall -Werror)
 
+# On ARM, the use of Eigen 3.4 with g++11 (both the versions in Jammy) causes a warning, which in turn causes
+# the build to fail. Details are here: https://gitlab.com/libeigen/eigen/-/issues/2326. We can suppress the warning
+# for ARM builds here.
+if(Eigen3_VERSION VERSION_EQUAL 3.4 AND
+CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 11.0 AND
+CMAKE_LIBRARY_ARCHITECTURE STREQUAL "aarch64-linux-gnu")
+    add_compile_options(-Wno-class-memaccess)
+endif()
+
 # fuse_publishers library
 add_library(${PROJECT_NAME}
   src/path_2d_publisher.cpp
--- a/fuse_publishers/CMakeLists.txt
+++ b/fuse_publishers/CMakeLists.txt
@@ -33,11 +33,16 @@ add_compile_options(-Wall -Werror)
 
 # On ARM, the use of Eigen 3.4 with g++11 (both the versions in Jammy) causes a warning, which in turn causes
 # the build to fail. Details are here: https://gitlab.com/libeigen/eigen/-/issues/2326. We can suppress the warning
-# for ARM builds here.
-if(Eigen3_VERSION VERSION_EQUAL 3.4 AND
-CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 11.0 AND
-CMAKE_LIBRARY_ARCHITECTURE STREQUAL "aarch64-linux-gnu")
-    add_compile_options(-Wno-class-memaccess)
+# for ARM builds here. However, this package does not depend directly upon Eigen, but includes headers from other
+# packages that do. So to prevent the error, we need to find Eigen and then check the version.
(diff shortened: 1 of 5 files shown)


fuse_publishers/CMakeLists.txt
```diff
@@ -31,6 +31,15 @@ catkin_package(
 ###########
 add_compile_options(-Wall -Werror)
 
+# On ARM, the use of Eigen 3.4 with g++11 (both the versions in Jammy) causes a warning, which in turn causes
+# the build to fail. Details are here: https://gitlab.com/libeigen/eigen/-/issues/2326. We can suppress the warning
+# for ARM builds here.
+if(Eigen3_VERSION VERSION_EQUAL 3.4 AND
```
ayrton04: This is actually a bit trickier than I thought.

This package itself has no direct dependency on Eigen, but it includes files that eventually include Eigen, and so we end up with the same build error.

However, because we don't directly depend on Eigen here, the variable `Eigen3_VERSION` isn't even declared, so we'll never enter this block.

Does anyone know if there's a way for packages that we include to export their build settings so that downstream packages pick them up? Ideally, any package that depends on, e.g., `fuse_core` would automatically get the `-Wno-class-memaccess` flag.

svwilliams: https://docs.ros.org/en/api/catkin/html/dev_guide/generated_cmake_api.html#catkin_package

In the docs for the `catkin_package()` command:
> DEPENDS (list of strings) – a list of CMake projects which this project depends on. Since they might not be find_packagable or lack a pkg-config file their INCLUDE_DIRS and LIBRARIES are passed directly. This requires that it has been find_package-ed before and all variables (<name>_FOUND, <name>_INCLUDE_DIRS, etc.) have the same case as this argument.

Based on that description, I would have expected dependent packages would have had access to the `Eigen3_FOUND` and similar variables after calling `find_package(catkin REQUIRED COMPONENTS fuse_core)`.

But it does say the string case used in `catkin_package(DEPENDS)` must match the case of the CMake variables. In `fuse_core` I have a depend on `EIGEN3` all caps:
https://github.com/locusrobotics/fuse/blob/devel/fuse_core/CMakeLists.txt#L29

Is that what the problem is? Just a case mismatch between `fuse_core` and this new if-block?

Barring that, using the `CFG_EXTRAS` parameter might be another way to make this work. If `fuse_core` exports some build flags using `CFG_EXTRAS`, dependent packages should pick that up, right? If their CMake files are written correctly. I think.

ayrton04: No, the all-caps version is empty even in packages that use Eigen directly. I had to print out all the available variables to figure that one out.

I'll try your other suggestion, thanks!

svwilliams: So maybe the `fuse_core` package is wrong, and it should be `Eigen3` instead of `EIGEN3`?

ayrton04: I think `fuse_core` is exporting the correct dependency:

https://github.com/ros/geometry2/blob/6df9e94363061a24ba890a6ff29771a96b0d756b/tf2_eigen/CMakeLists.txt#L34
https://github.com/cra-ros-pkg/robot_localization/blob/af30f3bc0592d84fcc06bdbdf6063c126f834493/CMakeLists.txt#L43

However, the `rolling` branch of this package uses this when exporting `ament` dependencies:

https://github.com/locusrobotics/fuse/blob/ff41ec07de91bca8f0b756d90e26be80753a6bab/fuse_core/CMakeLists.txt#L118

svwilliams (approved):
Are you still working out the details?

ayrton04:
Yes, sorry. Haven't had the cycles to look at this. Will pick it back up.

ayrton04:
@svwilliams are you still OK with this?

svwilliams:
Yep. Do you need the same thing in one of the ROS2 branches as well?

ayrton04:
Probably, but I haven't tried building it in ROS 2 on ARM/Jammy yet. Still, there's nothing in this change that's tied to any version of ROS, so I can add another branch for that.

PR #381, opened 2024-09-16
Fix stringop-overread compile issues

svwilliams:
When compiling ROS2 Rolling on Ubuntu Noble 24.04 using GCC 13.2.0, there are many instances of the following compiler error:
```
/usr/lib/gcc/x86_64-linux-gnu/13/include/emmintrin.h:169:19: error: ‘void* __builtin_memcpy(void*, const void*, long unsigned int)’ reading 16 or more bytes from a region of size 8 [-Werror=stringop-overread]
  169 |   *(__m128d *)__P = __A;
```

This seems _similar to_ a previously encountered compiler bug in GCC 12
```
# include/avxintrin.h:893:24: error: array subscript ‘__m256d_u[0]’ is partly outside array bounds of ‘Eigen::internal::plain_matrix_type<Eigen::Product<Eigen::Matrix<double, 3, 3, 1>, Eigen::Map<Eigen::Matrix<double, -1, 1> >, 0>, Eigen::Dense>::type [1]’ {aka ‘Eigen::Matrix<double, 3, 1> [1]’} [-Werror=array-bounds]
#   893 |   return *(__m256d_u *)__P;
```

But I do not see any open bug tickets about this issue in either Eigen or GCC. This issue may be entirely harmless; I'm really not sure.

However, every instance seems to occur when an Eigen Fixed-sized object is being assigned to an Eigen Dynamic-sized object. E.g. some code along the lines of:
```
fuse_core::Vector3d a;
a << 1, 2, 3;
fuse_core::VectorXd b = a;
```
A vast majority of these assignments occur in unit tests, so fixing those is trivial. I have no concerns about runtime at all:
```
fuse_core::VectorXd a(3);
a << 1, 2, 3;
fuse_core::VectorXd b = a;
```

There are a couple of places in actual library code where I update what type of Eigen object is being created, in `sensor_proc.hpp` and `unicycle_2d_ignition.cpp`. Both of those files are in the `fuse_models` package. I do not expect a meaningful performance impact from either of these changes.

And one instance in `normal_delta_orientation_3d_cost_functor.hpp` in the `fuse_constraints` package. That one should probably be classified as a bug. The size is known at compile-time; the `Eigen::Map` should never have be configured for a dynamic size.

Also, the CI build tests are going to fail. There is an API change to the `rclcpp::Waitable` class that is being fixed here: #379

### DIFF
--- a/fuse_constraints/include/fuse_constraints/normal_delta_orientation_3d_cost_functor.hpp
+++ b/fuse_constraints/include/fuse_constraints/normal_delta_orientation_3d_cost_functor.hpp
@@ -120,7 +120,7 @@ public:
 
     // Scale the residuals by the square root information matrix to account for the measurement
     // uncertainty.
-    Eigen::Map<Eigen::Matrix<T, Eigen::Dynamic, 1>> residuals_map(residuals, A_.rows());
+    Eigen::Map<Eigen::Matrix<T, 3, 1>> residuals_map(residuals);
     residuals_map.applyOnTheLeft(A_.template cast<T>());
 
     return true;
--- a/fuse_constraints/test/test_absolute_constraint.cpp
+++ b/fuse_constraints/test/test_absolute_constraint.cpp
@@ -59,9 +59,9 @@ TEST(AbsoluteConstraint, Constructor)
   {
     fuse_variables::AccelerationAngular2DStamped variable(
       rclcpp::Time(1234, 5678), fuse_core::uuid::generate("robby"));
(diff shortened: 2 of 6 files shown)


DavidLocus (approved):
LGTM

carlos-m159 (approved):

henrygerardmoore (approved):
Applying this on top of #379 allowed me to build in release and debug, and fixed the stringop-overread compile problems. :+1:

PR #384, opened 2024-11-01
Switch docs to use LaTeX instead of images

henrygerardmoore:
Currently the docs don't render well if you're in dark mode on github. Github supports inline LaTeX with MathJax, so this PR switches the docs to use that. It also removes a dead link. Here's a screenshot showing the difference (previewing in VSCode) where the bottom equation is using MathJax and the top one is the current image. This change also makes modifying the equations or creating new ones easier.

image

In reviewing, please do make sure I didn't make any typos in the LaTeX translations, and also feel free to add or change line breaks to make the `.md` files more readable :smile:


doc/Variables.md
```diff
@@ -47,8 +47,6 @@ places where most of the dimensions are unused. However, including too few physi
 also leads to inefficient and cumbersome usage when even the simplest of observation models involve many variables.
 This is one of those "Goldilocks principle" situations.
 
-![Goldilocks principle](http://home.netcom.com/~swansont_2/goldilocks.jpg)
-
```
henrygerardmoore, line 51: This link is dead, but if there is another image that works well as an example here, feel free to add or request me to add it.

henrygerardmoore:
~~This all looks nice in vscode, but it seems that github has some [MathJax rendering bugs](https://github.com/orgs/community/discussions/36915#discussioncomment-11124869) that mean it doesn't look as nice online. I'm not going to close this PR preemptively but I understand if you prefer the original images.~~

Looks like this is somehow just an issue with my browser :grin:

PR #386, opened 2024-11-20
Fix mutliple definition error

jakemclaughlin6:


jakemclaughlin6:
@svwilliams Minor change here to allow for multiple outside files to include these two headers.

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
```diff
@@ -88,7 +93,41 @@ class NormalDeltaPose3DCostFunctor
     const T* const orientation1,
     const T* const position2,
     const T* const orientation2,
-    T* residual) const;
+    T* residual) const {
```
svwilliams: /tmp/ws/src/fuse/fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h:96:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]

fuse_constraints/include/fuse_constraints/normal_prior_pose_3d_cost_functor.h
```diff
@@ -77,47 +77,40 @@ class NormalPriorPose3DCostFunctor
...
+  orientation_functor_(fuse_core::Matrix3d::Identity(), b_.tail<4>())  // Delta will not be scaled
+  {
+  }
 
   /**
    * @brief Evaluate the cost function. Used by the Ceres optimization engine.
    */
   template <typename T>
-  bool operator()(const T* const position, const T* const orientation, T* residual) const;
+  bool operator()(const T* const position, const T* const orientation, T* residual) const {
```
svwilliams: /tmp/ws/src/fuse/fuse_constraints/include/fuse_constraints/normal_prior_pose_3d_cost_functor.h:91:  when starting a new scope, { should be on a line by itself  [whitespace/braces] [4]

svwilliams (changes requested):
Minor linter fixes are needed

fuse_constraints/include/fuse_constraints/normal_delta_pose_3d_cost_functor.h
efernandez: Also above it looks like the style has changed.

Unless you @svwilliams have a `.clang-format` file we could use, I think you @jakemclaughlin6 will have to fix the style manually here and in the rest of the file. Just preserved the style from the original code when you moved it here.

fuse_constraints/include/fuse_constraints/normal_prior_pose_3d_cost_functor.h
```diff
@@ -77,47 +77,40 @@ class NormalPriorPose3DCostFunctor
...
 };
-
-NormalPriorPose3DCostFunctor::NormalPriorPose3DCostFunctor(const fuse_core::Matrix6d& A, const fuse_core::Vector7d& b) :
-  A_(A),
-  b_(b),
-  orientation_functor_(fuse_core::Matrix3d::Identity(), b_.tail<4>())  // Delta will not be scaled
-{
-}
-
-template <typename T>
```
efernandez, line 102: TBH I think this templated function/operator could be here - same for the other class.

The constructor isn't templated, so that one should probably be inside the class definition or in a cpp file. So for that one I actually understand it has to be moved to avoid compilation errors.

Either way, if this works, that's ok for me. :smiley:

efernandez (approved):
LGTM other than the style fixes required. Just keep the original style of the code you moved @jakemclaughlin6

svwilliams (approved):

DavidLocus (approved):

jakemclaughlin6:
@svwilliams I don't have permission to merge

Issue #387, opened 2025-01-21
Wrong time output in error message


roncapat:
Bug. Affected lines:
https://github.com/locusrobotics/fuse/blob/98cf1f3bebd0973a84f71647d8d9d4b39bb2216d/fuse_optimizers/src/fixed_lag_smoother.cpp#L411-L414

In my setup, I got:
```
The current lag expiration time is 1737452656123703516. The queued transaction with timestamp 1737452655904978711 from sensor visual_odometry has a minimum involved timestamp of 1737452655904978711, which is 218724805 seconds too old. Ignoring this transaction.
```

Assuming timestamps in nanoseconds, we have:
```
1737452656123703516
Assuming that this timestamp is in nanoseconds (1 billionth of a second):
GMT: Tuesday 21 January 2025 09:44:16.123
```

```
1737452655904978711
Assuming that this timestamp is in nanoseconds (1 billionth of a second):
GMT: Tuesday 21 January 2025 09:44:15.904
```

so the difference of `218724805 seconds` shall be instead `0.218724805` seconds.

I can submit a PR that fixes this time scale problem when printing the time difference.

PR #389, opened 2025-02-03
Fix warning message

roncapat:
Related to #387 but on `rolling` (ROS 2) branch.


svwilliams (approved):
I'm inclined to switch the timestamp printing to `.seconds()` as well as the durations.

needphpsql (approved):

PR #392, opened 2025-02-27
Prevent optimization thread from doing its work when a stop was requested

carlos-m159:
Publisher-specific implementations may lead to a crash if the optimizer attempts to call `notify` after it has been stopped.  A potential issue arises if an optimization request is already in progress when the `stop` call is issued.

### DIFF
--- a/fuse_optimizers/src/batch_optimizer.cpp
+++ b/fuse_optimizers/src/batch_optimizer.cpp
@@ -157,6 +157,11 @@ void BatchOptimizer::optimizationLoop()
     }
     {
       std::lock_guard<std::mutex> lock(optimization_mutex_);
+      if (!started_)
+      {
+        ROS_DEBUG_STREAM("Optimizer stopped while trying to acquire optimization_mutex_. Skipping optimization.");
+        continue;
+      }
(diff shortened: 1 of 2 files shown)


svwilliams (approved):

svwilliams (approved):

fuse_optimizers/src/batch_optimizer.cpp
```diff
@@ -289,19 +294,19 @@ void BatchOptimizer::start()
 void BatchOptimizer::stop()
 {
   ROS_INFO_STREAM("Stopping optimizer.");
-  // Tell all the plugins to stop
-  stopPlugins();
+  started_ = false;
```
needphpsql, line 297: Actually, this 'started_' variable itself is a shared resource as well between main thread and the optimization thread. But it only changes in main thread. So no protection should be okay.

needphpsql (approved):

fuse_optimizers/src/batch_optimizer.cpp
carlos-m159: In this case, started is an atomic bool, so we are safe.

PR #393, opened 2025-03-06
Prevent optimizer thread from calling notify on stopped publishers

carlos-m159:
Port fix to ros2 branch: https://github.com/locusrobotics/fuse/pull/392


svwilliams (approved):
Did the batch optimizer not need to be updated as well?

carlos-m159:
> Did the batch optimizer not need to be updated as well?

Looks like stop/start was not ported to ros2.

DavidLocus (approved):

PR #394, opened 2025-04-10
Imu2D: Use transformed twist when computing covariance

d3murphy:
Fix an error where we do not use the `transformed_twist` after computing it.

### DIFF
--- a/fuse_models/src/imu_2d.cpp
+++ b/fuse_models/src/imu_2d.cpp
@@ -47,18 +47,23 @@
 #include <memory>
 #include <utility>
 
+
 // Register this sensor model with ROS as a plugin.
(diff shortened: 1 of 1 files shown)


d3murphy:
cc: @efernandez: I don't think I can add reviewers.

efernandez:
> cc: @efernandez: I don't think I can add reviewers.

:thinking: not sure why is that - anyway, I've added some reviewers now

efernandez (approved):
LGTM - good catch

If you can't, please squash the commits into one, although I believe when they merge it that'd be done anyway - I'm just not 100% sure

svwilliams (approved):

DavidLocus (approved):

PR #395, opened 2025-04-14
Only instantiate tf_listeners if needed

d3murphy:
- We find that the transform listeners in sensor models are negatively impacting performance
- Do not instantiate transform listeners if target frames are empty
- Target frames are empty by default
- Existing logic already does not do the transforms when target frames are empty

### DIFF
--- a/fuse_models/src/acceleration_2d.cpp
+++ b/fuse_models/src/acceleration_2d.cpp
@@ -41,6 +41,8 @@
 #include <pluginlib/class_list_macros.hpp>
 #include <ros/ros.h>
 
+#include <memory>
+
 
 // Register this sensor model with ROS as a plugin.
 PLUGINLIB_EXPORT_CLASS(fuse_models::Acceleration2D, fuse_core::SensorModel)
@@ -51,7 +53,6 @@ namespace fuse_models
 Acceleration2D::Acceleration2D() :
   fuse_core::AsyncSensorModel(1),
   device_id_(fuse_core::uuid::NIL),
-  tf_listener_(tf_buffer_),
   throttled_callback_(std::bind(&Acceleration2D::process, this, std::placeholders::_1))
 {
 }
@@ -70,6 +71,11 @@ void Acceleration2D::onInit()
   {
     ROS_WARN_STREAM("No dimensions were specified. Data from topic " << ros::names::resolve(params_.topic) <<
                     " will be ignored.");
+  }
+
+  if (!params_.target_frame.empty())
+  {
+    tf_listener_ = std::make_unique<tf2_ros::TransformListener>(tf_buffer_);
   }
 }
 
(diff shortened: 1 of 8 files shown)


d3murphy:
CC: @efernandez

efernandez (approved):

fuse_models/src/acceleration_2d.cpp
```diff
@@ -71,6 +70,11 @@ void Acceleration2D::onInit()
     ROS_WARN_STREAM("No dimensions were specified. Data from topic " << ros::names::resolve(params_.topic) <<
                     " will be ignored.");
   }
+
```
efernandez, line 75: You need to `#include <memory>` above in order to use `std::make_unique`. The linter is failing on that: https://build.ros.org/job/Npr__fuse__ubuntu_focal_amd64/207/testReport/junit/roslint/LintCheck/roslint_fuse_models/
```bash
/tmp/ws/src/fuse/fuse_models/src/acceleration_2d.cpp:76:  Add #include <memory> for make_unique<>  [build/include_what_you_use] [4]
```

efernandez (changes requested):

svwilliams (approved):

efernandez (approved):

fuse_models/src/acceleration_2d.cpp
```diff
@@ -71,6 +72,11 @@ void Acceleration2D::onInit()
...
+  if (!params_.target_frame.empty())
+  {
+    tf_listener_ = std::make_unique<tf2_ros::TransformListener>(tf_buffer_);
```
ayrton04, line 78: I'm assuming one of the reasons we're seeing performance issues is that we only care about static transforms in some cases, and the listener is having to deserialize all transforms. Would it be worth adding a param to the sensor models that indicate we only want static transforms? Rolling [has that option now].(https://github.com/ros2/geometry2/blob/7767a7072e7fb9acb7d18127984fccee048de060/tf2_ros/src/transform_listener.cpp#L43) It forces the listener to only subscribe to `/tf_static`.

anhillman (approved):

fuse_models/src/acceleration_2d.cpp
d3murphy: Hi @ayrton04 . At least on our system, the performance issue isn't coming from deserialization but rather from the condvar timed waits (and associated context switching) in the listener thread. This seems to scale poorly with both the number of threads in a process and the overall system load.

You may be able to get a similar improvement by using the static tf listener and not starting the thread but I'm not sure whether you'd now see latency on the call to the buffer since you don't have the dedicated thread. I guess this depends on the internal details of how it's implemented. 

In our case, the high usage from the listener threads jumps out in profiler results so you could try it and see if 1) you are seeing a lot of usage in the listener threads and 2) how the static TF performs with no dedicated thread.

d3murphy:
Hi. I don't have permission to merge this. Could someone please merge for me if everything looks good? Thanks!

PR #400, opened 2025-04-16
Port patch #395 into ROS 2 Rolling

svwilliams:
Only instantiate tf_listeners if needed (#395)

### DIFF
--- a/fuse_models/src/acceleration_2d.cpp
+++ b/fuse_models/src/acceleration_2d.cpp
@@ -31,6 +31,8 @@
  *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  *  POSSIBILITY OF SUCH DAMAGE.
  */
+#include <memory>
+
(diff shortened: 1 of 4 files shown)


anhillman (approved):

fuse_models/src/acceleration_2d.cpp
```diff
@@ -86,13 +88,15 @@ void Acceleration2D::onInit()
...
-  );
+  if (!params_.target_frame.empty()) {
+    tf_listener_ = std::make_unique<tf2_ros::TransformListener>(
+      *tf_buffer_,
+      interfaces_.get_node_base_interface(),
+      interfaces_.get_node_logging_interface(),
+      interfaces_.get_node_parameters_interface(),
+      interfaces_.get_node_topics_interface()
+    );
+  }
```
needphpsql, line 99: this should be fine since these are decided in compiling time.

needphpsql (approved):

PR #409, opened 2025-05-26
[RST-12952] Modified the loadDeviceId() function to accept optional parameters

svwilliams:
Modified the loadDeviceId() function to accept optional parameter names for the device UUID and device name

### DIFF
--- a/fuse_variables/src/stamped.cpp
+++ b/fuse_variables/src/stamped.cpp
@@ -43,7 +43,10 @@ namespace fuse_variables
 {
 
 fuse_core::UUID loadDeviceId(
-  fuse_core::node_interfaces::NodeInterfaces<fuse_core::node_interfaces::Parameters> interfaces,
+  fuse_core::node_interfaces::NodeInterfaces<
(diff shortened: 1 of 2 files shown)


fuse_variables/src/stamped.cpp
```diff
@@ -37,25 +37,35 @@
...
 
-  device_str = fuse_core::getParam(interfaces, "device_name", std::string());
+  device_str = fuse_core::getParam(interfaces, name_parameter, std::string());
   if (!device_str.empty()) {
     return fuse_core::uuid::generate(device_str);
   }
 
+  if (!silent) {
+    RCLCPP_WARN_STREAM(
+      rclcpp::get_logger("fuse"),
```
carlos-m159: Maybe not for this PR as we could break some "_client_" code, but would it make sense to pass the logging interface in the `NodeInterfaces` _mixin_?

carlos-m159 (approved):

DavidLocus (approved):

PR #410, opened 2025-06-30
add user configurable diag mapping support of termination types.

needphpsql:
Previously, the NO_CONVERENCE termination type is always reported as warnings in diagnostics. However, it is not suitable for all the use cases. Therefore, this PR introduces two user configurable mapping parameters diagnostic_warning_status and diagnostic_error_status so that the user can define what should be reported as error or warning in diagnostics.

For example:
diagnostic_warning_status { "NO_CONVERGENCE" };

diagnostic_error_status {"FAILURE", "USER_FAILURE"};

If the user doesn't do anything, the behavior is the same as before this PR.

### DIFF
--- a/fuse_optimizers/src/fixed_lag_smoother.cpp
+++ b/fuse_optimizers/src/fixed_lag_smoother.cpp
@@ -692,7 +692,7 @@ diagnostic_msgs::DiagnosticStatus terminationTypeToDiagnosticStatus(const ceres:
       }
       else if (contains(diag_errors, diag_level))
       {
-        makeDiagnosticStatus(diagnostic_msgs::DiagnosticStatus::ERROR, "Optimization failed");
+        return makeDiagnosticStatus(diagnostic_msgs::DiagnosticStatus::ERROR, "Optimization failed");
       }
       else
       {
--- a/fuse_optimizers/src/fixed_lag_smoother.cpp
+++ b/fuse_optimizers/src/fixed_lag_smoother.cpp
@@ -653,7 +653,6 @@ std::string mapCeresLogToDiagLog(const ceres::TerminationType & termination_type
  *
  * @param[in] vec The list of strings
  * @param[in] str The str to search
- * @param[in] diag_errors The diagnostic errors
  * @return true if contains, false otherwise
  */
 inline bool contains(const std::vector<std::string>& vec, const std::string& str)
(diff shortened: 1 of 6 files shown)


fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -633,17 +671,33 @@ diagnostic_msgs::DiagnosticStatus makeDiagnosticStatus(const int8_t level, const
...
 {
+  std::string diag_level = mapCeresLogToDiagLog(termination_type);
   switch (termination_type)
   {
     case ceres::TerminationType::CONVERGENCE:
     case ceres::TerminationType::USER_SUCCESS:
       return makeDiagnosticStatus(diagnostic_msgs::DiagnosticStatus::OK, "Optimization converged");
     case ceres::TerminationType::NO_CONVERGENCE:
-      return makeDiagnosticStatus(diagnostic_msgs::DiagnosticStatus::WARN, "Optimization didn't converge");
+      if (contains(diag_warnings, diag_level))
```
svwilliams: Shouldn't this `if(contains(...)` block replace the entire the `switch()` statement?
The log level of any TerminationType can be controlled by the two new parameters?

needphpsql: But diag_warning or diag_error doesn't include the success types. If replacing the entire switch, putting them into 'default'?

svwilliams: I think so.
* If the TerminationType is in the diag_errors vector, then report it as a ERROR.
* If the TerminationType is in the diag_warnings vector, then report it as WARN.
* Otherwise assume it is fine and report it as OK

And that can apply to all TerminationType values.

fuse_optimizers/src/fixed_lag_smoother.cpp
```diff
@@ -624,28 +624,59 @@ diagnostic_msgs::DiagnosticStatus makeDiagnosticStatus(const int8_t level, const
...
+      return "FAILURE";
+  }
+}
+
+/**
+ * @brief Helper function to check if a string is contained in the list
+ *
+ * @param[in] vec The list of strings
+ * @param[in] str The str to search
+ * @param[in] diag_errors The diagnostic errors
```
svwilliams: diag_errors is not an input to this function
```suggestion
```

svwilliams (approved):

carlos-m159 (approved):

PR #413, opened 2025-07-03
Moving StampedVariableSynchronizer from fuse_publishers to fuse_variables

DavidLocus:
Moves the StampedVariableSynchronizer from fuse _publishers to the fuse_variables package.


svwilliams (approved):
You even got the deprecated header fallback. :+1: 

Don't forget to port this change to the ROS2 version in the `rolling` branch as well.

anhillman (approved):

PR #414, opened 2025-07-08
Porting StampedVariableSynchronizer changes to ROS 2

DavidLocus:


fuse_variables/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ endif()
 
 find_package(ament_cmake_ros REQUIRED)
 find_package(fuse_core REQUIRED)
+find_package(fuse_graphs REQUIRED)
```
svwilliams: Since `fuse_graphs` is a test-only dependency, it seems like it should not be included in the main build? The CMake file in the `test` folder already finds `fuse_graphs` internally.

fuse_core/include/fuse_core/callback_wrapper.hpp
```diff
@@ -158,7 +158,7 @@ class CallbackAdapter : public rclcpp::Waitable
 public:
   explicit CallbackAdapter(std::shared_ptr<rclcpp::Context> context_ptr);
 
-  std::vector<std::shared_ptr<rclcpp::TimerBase>> get_timers() const override
+  std::vector<std::shared_ptr<rclcpp::TimerBase>> get_timers() const
```
svwilliams: The `override` keyword was removed. I'm wondering if this is a difference between `rolling` and `jazzy`? It may need to stay in place for `rolling`?

The `jazzy` version of `CallbackAdapter` doesn't have a `get_timers()` function at all.
https://github.com/locusrobotics/fuse/blob/jazzy/fuse_core/include/fuse_core/callback_wrapper.hpp

So I think this change should be dropped.

svwilliams (changes requested):

fuse_core/include/fuse_core/callback_wrapper.hpp
DavidLocus: That's fine, I will add it back, ROS2 Hotdog just doesn't build with the `override` keyword there which got extremely annoying very quickly

svwilliams (approved):

anhillman (approved):

PR #420, opened 2025-08-01
Add possibility to make variable constant during optimization in the base variable class

carlos-m159:


carlos-m159:
I will need to revisit this, closing it for now.

PR #422, opened 2026-03-06
Fix Ceres inclusion and path export

jprestwo:
Ceres is referenced in fuse header files and the existing way it was being included in fuse_core did not export the include paths for Ceres which breaks downstream packages when they include fuse_core headers.

This was only exposed through the per-package builds since the existing merged builds get lucky that all the include dirs are merged together.


garyservin (approved):

fuse_core/CMakeLists.txt
```diff
@@ -73,15 +83,16 @@ target_include_directories(${PROJECT_NAME}
...
     ${GLOG_INCLUDE_DIRS}
 )
 target_link_libraries(${PROJECT_NAME}
-  ${Boost_LIBRARIES}
-  ${catkin_LIBRARIES}
-  ${CERES_LIBRARIES}
-  ${GLOG_LIBRARIES}
+  PUBLIC
+    Ceres::ceres
+    ${catkin_INCLUDE_DIRS}
```
biancabnd: This should be `catkin_LIBRARIES` here

fuse_core/CMakeLists.txt
```diff
@@ -73,15 +83,16 @@ target_include_directories(${PROJECT_NAME}
   SYSTEM PUBLIC
     ${Boost_INCLUDE_DIRS}
     ${catkin_INCLUDE_DIRS}
-    ${CERES_INCLUDE_DIRS}
```
biancabnd, line 76: we don't need the include dirs here anymore?

jprestwo: I'm trying to remember... I may need to test again to be sure, but IIRC Ceres didn't export that variable, which is why I had to manually set `CERES_INCLUDEDIRS` from `INTERFACE_INCLUDE_DIRECTORIES`.

In my debugging I was printing out `CERES_INCLUDE_DIRS` and it was empty...

fuse_core/CMakeLists.txt
jprestwo: Good catch! I don't think it actually caused a problem funny enough, but I'll add it back

svwilliams (commented):
I want to be really careful here. `fuse` is a public, open-source repo. I don't want to break anything for official builds of ROS1, end of life though it may be. Do these changes work with older versions of Ceres (2.0, 2.1, etc) and Ubuntu? Should I make a devel branch specific to Ubuntu Jammy and/or Noble?

jprestwo:
> I want to be really careful here. fuse is a public, open-source repo. I don't want to break anything for official builds of ROS1, end of life though it may be. Do these changes work with older versions of Ceres (2.0, 2.1, etc) and Ubuntu? Should I make a devel branch specific to Ubuntu Jammy and/or Noble?

@svwilliams I'm unsure about older versions. But at the high level I think fuse was flat out broken anyways regarding this. I guess everyone out there does `--merge-install`?

The core issue is that ceres itself doesn't export the proper variables, or at least not `INCLUDE_DIRS`. 

Also, is this fork public? I don't see that it is...
Edit: I'm not well informed ^^^. Its indeed public :rofl:

jprestwo:
Also this bit specifically really is the correct way to do things:
```
 PUBLIC
    Ceres::ceres
    ${catkin_LIBRARIES}
```
There are headers in fuse_core which reference ceres headers, so these need to be defined as public otherwise anything including the fuse_core header won't be able to find the ceres headers. I think the big question (even for me) is about pulling in the include directories in somewhat of a non-standard fashion (i.e. not simply using `ceres_INCLUDE_DIRS`)

fuse_core/CMakeLists.txt
jprestwo: Ok actually I think you're right. I don't think this caused a build failure, but I'll keep this (except the new variable)

PR #434, opened 2026-08-15
Fix FixedLagSmoother crash on first optimization cycle in auto-start mode

ikhann:
## Problem

When the fixed-lag smoother is configured **without any ignition sensors** it auto-starts
("No ignition sensors were specified. Optimization will begin immediately."), and then aborts on
the very first sensor transaction:

```
[INFO] [...] [fixed_lag_node]: No ignition sensors were specified. Optimization will begin immediately.
terminate called after throwing an instance of 'std::runtime_error'
  what():  can't compare times with different time sources
[ros2run]: Aborted
```

## Root cause

`lag_expiration_` is default-constructed, which produces an `rclcpp::Time` with the
`RCL_SYSTEM_TIME` clock type. On the first optimization cycle it is passed to `processQueue()`
before ever being assigned, and compared against `RCL_ROS_TIME` sensor stamps
(`if (min_stamp < lag_expiration)`), which throws. The exception escapes the optimization thread,
so the whole node terminates.

Configurations with an ignition sensor return from `processQueue()` before reaching that
comparison, which is why this only manifests in auto-start mode. The existing
`// TODO(CH3): We might have to make sure lag_expiration_ has been initialised` at the
`processQueue()` call site, and the reset path (which already re-initializes `lag_expiration_`
with `RCL_ROS_TIME`), both point at this.

## Reproduction (before this patch)

Any config with a sensor model but no ignition sensor, e.g. Unicycle2D + Odometry2D, then:

```bash
ros2 run fuse_optimizers fixed_lag_smoother_node --ros-args -r __node:=fixed_lag_node --params-file autostart.yaml
ros2 topic pub -r 2 /odom nav_msgs/msg/Odometry "{header: auto, pose: {pose: {position: {x: 1.0}}, covariance: [...]}}"
```

The node aborts on the first optimization cycle after the first message.

## Fix

Initialize the member with the `RCL_ROS_TIME` clock type, mirroring the existing `start_time_`
member two declarations below it. The now-resolved TODO comment is removed.


veluneos:
Hi — I independently reproduced the before/after behavior around this PR and ran the resulting executions through a structural comparison.

I used the current base commit as the Reference and this PR head as the Target, with alternating run order (R1 → T1 → R2 → T2).

The result was consistent across all four trials: the Reference runs terminated on the first optimization cycle before producing any `/odom` output, while the Target runs remained active and produced 55 `/odom` messages over the same bounded run.

Across all four Reference→Target comparisons, VeluneOS surfaced `/odom` as a Target-only stream with roughly 5.7 seconds of unmatched coverage. The same-condition controls were substantially quieter, below 20 ms.

I’m not treating this as an independent judgment that the proposed fix should be merged or as a separate root-cause conclusion. I’m testing whether before/after runtime comparison can consistently localize where execution behavior diverges.

If useful, I can share the exact commits, reproduction setup, run matrix, raw captures, and comparison outputs.

If there’s a current internal before/after robotics run where this kind of structural review would be useful, I’d also be interested in running the same comparison on one Reference/Target pair.

